PR #3, opened 2018-03-24
Feature/rescue board

theunkn0wn1:
**copied description from the original PR.**
~~This is a work in progress, and is not complete. Therefore, it should NOT be merged yet!~~
~~I am opening this pull request as I am at a point where i would like feedback, as well as to say i am nearly done with the `Rescue` object. I think everything that mecha needs to keep track of has a property now, and most of that has tests. i still have a couple more tests to write before this can be merged.~~

~~Further, the Rescue board itself is just a stub, that will need to be implemented before this PR can be merged. If anyone has any feedback of what they want changed/added/removed: now is the time.~~

Unless i have missed something obvious, this feature has been sufficiently developed. the only thing it is missing is API integration and a few methods should be converted to async for easier API integration.

Unfortunately markdown doesn't support checkboxes in tables and collapsing the lists breaks part of the integration... so this is going to get lengthy. To save on vertical space (lol), checklist items can only be checked off _**After**_ their associated tests **and** implementation have been completed.

Checklist time!
----------
# Rescue
- [x] `case_id`
- [x] `client`
- [x] `system`
- [x] `created_at`
- [x] `updated_at`
- [x] `unidentified_rats`
- [x] `active`
- [x] `quotes`
- [x] `is_open`
- [x] `epic`
- [x] `code_red`
- [x] `successful`
- [x] `title`
- [x] `first_limpet`
- [x] `board_index`
- [x] `mark_for_deletion`
- [x] `lang_id`
- [x] `irc_nickname`
- [x] `rats`
- [x] `add_rat`
- [x] `platform`
- [x] `__eq__`

# Rats

## properties

- [x] `name`
- [x] `uuid`
- [x] `platform`

## class members

- [x] cache_by_id
- [x] cache_by_ratname

## classmethods

- [x] `flush`
- [x] `get_rat`

## constructor

- [ ] `__init__`

# Quotes

## constructor
- [x] `__init__`

## properties

- [x] `message`
- [x] `author`
- [x] `created_at`
- [x] `updated_at`
- [x] last_author

## methods

- [x] `modify`

# RatBoard
## constructor
- [x] `__init__`

## properties
 - [x] rescues

## attributes
 - [ ] handler

## methods
- [x] search
- [x] remove
- [x] modify
- [x] clear_board
- [x] `__contains__`
- [x] `find_by_index`
- [x] `find_by_name`
- [x] `find_by_uuid`
- [ ] `retrieve_open_cases_from_api`


theunkn0wn1:
Would be nice if github told be what was in conflict >.>

Modules/Rat_Board.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def find_by_name(self, client: str) -> Rescue or None:
+        """
+        Searches for and returns a Rescue for a given client, should it exist.
+
+        Returns:
+            Rescue: found rescue
+            None:   no such rescue
+        """
+        for index, rescue in self._rescues.items():
```
MHajoha: When you don't need the keys, dict.values() will provide you with a set (ish) representation of the values.

Modules/Rat_Board.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    pass
+
+
+class IndexNotFreeError(RescueBoardException):
+    """ Exception raised when someone attempts to attach a Rescue to a board index in-use"""
+    pass
+
+
+class RatBoard(object):
```
MHajoha: Can you provide me with a __contains__ method? Might also need a __eq__ on the rescue object for that.

Modules/Rat_Board.py
```diff
@@ -0,0 +1,197 @@
...
+            client (str): client's commander name / irc nickname, may be None
+            api_id (UUID): api id of rescue, may be None
+
+        Returns:
+            Rescue: Found rescue on the board, otherwise None
+
+        Raises:
+            ValueError: command invoked without params, or all params are None
+        """
+        raise NotImplementedError
```
MHajoha: Best to kill the method entirely. Afaik nothing's using it yet and raising NotImplementedError instead doesn't really help.

Modules/Rat_Board.py
theunkn0wn1: Sure thing. I presume `__contains__ is for `Rescue` objects?

Modules/Rat_Board.py
theunkn0wn1: Its in the process of being reimplemented to accept a string. the NotImplementedError is temporary >.>

Modules/Rat_Board.py
theunkn0wn1: Implemented, see latest commits

Modules/Rat_Board.py
theunkn0wn1: *nods* til

Modules/rat_rescue.py
```diff
@@ -93,6 +93,49 @@ def __init__(self, case_id: UUID, client: str, system: str, irc_nickname: str,
...
+                assert self.title == other.title
+                assert self.first_limpet == other.first_limpet
+                assert self.mark_for_deletion == other.mark_for_deletion
+                assert self._lang_id == other._lang_id  # TODO: implement property
+                assert self.rats == other.rats
+                assert self._irc_nick == other._irc_nick  # TODO: implement property
+            except AssertionError as ex:
+                raise ex
+                # if at least one of the above statements is false.
+                return False
```
MHajoha: 1. the `return False` is unreachable
2. this isn't a great way of doing things. have a look at the `all` built-in function
3. perhaps checking whether two rescue objects refer to the same rescue would be more useful than checking if all their attributes match. What I'm thinking of is to, when an update comes and the rescue is (or has become) open, check if that rescue is on our board and if so, modify it.

theunkn0wn1: 1. i spotted it after i pushed and already fixed that.
2. i agree this probably isn't the best way of doing it, i was after something that worked. 
3. you asked for `__eq__`, which is to check if two object instances are equal. That is different from asking if two objects are the same. If you want to check if a newly created Rescue is a duplicate of an existing, perhaps a check against `search_by_name` may work here instead.

theunkn0wn1: Il look into a more clean solution for 2.
i can probably loop over attributes or something

theunkn0wn1: ok, so in this instance i think using a non-standard behavior for __contains__ may be best.
When an update comes in / rescue object is to be created it will check to see if there is *an* rescue on the board that already matches whats about to be created.

The check would have to verify guid, client name, and/or platform?

Modules/rats.py
```diff
@@ -0,0 +1,183 @@
...
+            LOG.debug(f"value was a string with data '{value}'")
+            uuid = UUID(value)
+            LOG.debug("parsed value into a valid UUID.")
+            self._uuid = uuid
+        elif isinstance(value, UUID):
+            self._uuid = value
+
+        else:
+            raise TypeError(f"expected type None or type str, got "
+                            f"{type(value)}")
```
MHajoha: The string representation for types is `<class 'int'>` etc.. You'll want to use `type(value).__name__` here.

Modules/rats.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Sets the platform for a given rat
+
+        Args:
+            value (Platforms): new platform
+        """
+        if isinstance(value, Platforms):
+            self._platform = value
+        else:
+            raise TypeError(f"Expected a {Platforms} object, got type {type(value)}")
```
MHajoha: Similar here as above: assuming that `Platforms` is a enum, you'll get `<enum 'Platforms'>` here. Platforms.__name__ is what you want.

Modules/rats.py
theunkn0wn1: calling a magic method for a type error does not feel like an elegant solution. The fact an exception is being raised here and the information provided by the string representation should be sufficient for debugging purposes

Modules/rats.py
theunkn0wn1: Platforms is an enum, though again it should be enough information to point us at the fault for thorough debugging and bugfixing. Calling a magic method for something like this seems needless

Modules/rats.py
MHajoha: All good, this is, of course, your decision and only a cosmetic change.
Incidentally, __name__ is not a method but an attribute (or property, not actually sure) and since no function exists for getting the name of a class, this is generally how it's done.

PR #8, opened 2018-04-22
Implement Rescue.status

theunkn0wn1:
This PR implements the `Rescue.status` property and its Setter.
This property returns an instance of the `Status` enum, which i placed in `ratlib\names.py`

The Status enum includes three states:
`OPEN` case is open and active
`CLOSED` case is marked as closed
`INACTIVE` case is marked as open and inactive

The `Rescue.is_open` property has been renamed `rescue.open` and has been reimplemented to make use of the `Rescue.status` property. The `rescue._open` attribute has been removed.

the `Rescue.inactive` property has been reimplemented to use `Rescue.status` property.
the attribute `Rescue._inactive` has been removed.

### DIFF
--- a/Modules/rat_rescue.py
+++ b/Modules/rat_rescue.py
@@ -537,7 +537,7 @@ class Rescue(object):
             bool: is case open?
 
         """
-        return False if self.status is Status.CLOSED else True
+        return self.status is not Status.CLOSED
 
     @open.setter
     def open(self, value: bool) -> None:
--- a/Modules/rat_rescue.py
+++ b/Modules/rat_rescue.py
@@ -18,7 +18,7 @@ from uuid import UUID
 import config
 from Modules.rat_quotation import Quotation
 from Modules.rats import Rats
-from ratlib.names import Platforms
+from ratlib.names import Platforms, Status
 
 LOG = logging.getLogger(f"{config.Logging.base_logger}.{__name__}")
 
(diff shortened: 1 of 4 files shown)


Modules/rat_rescue.py
```diff
@@ -502,23 +529,20 @@ def unidentified_rats(self, value) -> None:
...
-        - this cannot be named `Rescue.open` as it would shadow the name from
-        the outer scope (bad)
+        Helper method for determining if a case is considered open or not
 
         Returns:
             bool: is case open?
 
         """
-        return self._open
+        return False if self.status is Status.CLOSED else True
```
MHajoha: `return self.status is not Status.CLOSED`?

Modules/rat_rescue.py
```diff
@@ -28,17 +28,18 @@ class Rescue(object):
...
-                 board: 'RatBoard' = None, created_at: datetime = None, updated_at: datetime = None,
-                 unidentified_rats=None, active=True, quotes: list = None, is_open=True, epic=False,
-                 code_red=False, successful=False, title: str = '',
-                 first_limpet: UUID or None = None, board_index: int = None,
-                 mark_for_deletion: dict or None = None, lang_id: str = "EN", rats: list = None):
+    def __init__(self, case_id: UUID, client: str, system: str, irc_nickname: str, board: 'RatBoard' = None,
+                 created_at: datetime = None, updated_at: datetime = None, unidentified_rats=None, active=True,
+                 quotes: list = None, epic=False, title: str = '', first_limpet: UUID or None = None,
+                 board_index: int = None, mark_for_deletion: dict or None = None, lang_id: str = "EN",
+                 rats: list = None, status: Status = Status.OPEN, code_red=False):
```
MHajoha: The outcome should still be part of the constructor. Otherwise, commands handling already finished rescues can't be handled properly. Plus, if you hardcode it there's no point in us tracking it in the first place.

I personally also believe that there should be a setter for it, so that, in the future, a command changing the outcome of an already finished case is possible. (And I don't believe in limiting developers for any other reason than hiding implementation or providing compatability.) This point may require further discussion in #popcorn, though.

Modules/rat_rescue.py
theunkn0wn1: `Status` are not booleans.

MHajoha: No shit Sherlock. 
In glorious Python fashion, `is not` is a single operator and returns the opposite of `is` in all cases. Including testing two references against each other.

MHajoha: Here, it would be equivalent to using `!=`

theunkn0wn1: oh wait, i misread you completely... :/
yes `is not` should be the implementation. will fix

Modules/rat_rescue.py
theunkn0wn1: if we need to expose this functionality, we can refactor it in later. As we discussed in IRC we are at a difference in methodology. Mecha has no reason to be modifying that property, and should always set it to None. If we need to change this functionality, we can refactor it in at a later point in time.

PR #9, opened 2018-04-22
Clean up and the rule decorator we were chatting about earlier

MHajoha:
This implements the rule decorator and makes changes to the trigger object.
While I was at it, I also did some relatively extensive clean up of rat_command.py and its tests.
Proper description will follow tomorrow, but now it's late.


tests/test_trigger.py
```diff
@@ -1,80 +1,71 @@
...
+@pytest.fixture
+def bot():
+    return MockBot()
+
+def test_create_manual_channel(bot):
+    trigger = Trigger(bot, ["don't", "care"], ["don't care", "care"], "test_nick", "#somechannel",
+                      "test_ident", "test.vhost")
+    assert trigger.channel == "#somechannel"
 
-class TriggerTests(unittest.TestCase):
```
theunkn0wn1: I  would appreciate this still being in a class, for the sake of test organization, but change it to inherit from `Object` rather than `TestCase`

(consistency / nitpick)

Modules/rat_command.py
```diff
@@ -205,12 +170,20 @@ def real_decorator(func):
...
-    def get_command(cls, name: str):
-        # remove the prefix.
-        name = name.strip(cls.prefix)
-        # see if its a command
-        if name in cls._registered_commands:
-            cls.log.debug("command found!")
-            return cls._registered_commands[name]
-        else:
-            return None
+    def rule(cls, regex: str):
```
theunkn0wn1: I do not think all commands hooked via a rule should receive the event if no conventional commands were found, though perhaps a wildcard regex would suffice?

tests/test_trigger.py
```diff
@@ -1,80 +1,71 @@
...
-    @async_test
-    async def test_reply_channel(self):
-        trigger = Trigger(
-            self.bot, "test_nick", "#somechannel", "test_ident", "test.vhost")
-        await trigger.reply("Exceedingly smart test message.")
-        self.assertIn({
-            "target": "#somechannel",
-            "message": "Exceedingly smart test message."
-        }, self.bot.sent_messages)
+@async_test
```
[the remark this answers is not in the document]
MHajoha: Good point, didn't know that existed

tests/test_trigger.py
```diff
@@ -1,80 +1,71 @@
-import unittest
-
+import pytest
 from aiounittest import async_test
 
 from Modules.trigger import Trigger
 from tests.mock_bot import MockBot
 
+@pytest.fixture
+def bot():
```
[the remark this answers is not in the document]
MHajoha: Aye, I wasn't sure what your intention behind conftest.py was. The name seems odd to me. Wouldn't fixtures.py be more appropriate.

Modules/rat_command.py
MHajoha: I'm afraid I don't quite follow you?

tests/test_trigger.py
MHajoha: I don't know... This doesn't need to be object-oriented so I made it not be...

As for consistency, we are already using two separate test frameworks ;)

Modules/rat_command.py
MHajoha: Only one matching rule is executed if no conventional command was found. Since python dicts from 3.6 onward are more or less ordered, this will be the first registered rule.

tests/test_trigger.py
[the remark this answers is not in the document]
theunkn0wn1: `Conftest` is the required name, according to the Documentation for PyTest.

PyTest looks for a `conftest.py` within the working directory, higher level `conftest`s take precedence over lower-level ones.

Optimally i would name it `fixtures.py`

tests/test_trigger.py
theunkn0wn1: We should probably convert all the old to the new at some point 😉 

I was using classes more for Struture than anything else.. its a minor nit pick

Modules/rat_command.py
theunkn0wn1: ahh, so its First *matching* rule. ok.

That means i misunderstood. carry on

tests/test_trigger.py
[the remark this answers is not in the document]
MHajoha: Okay that's alright then. Strange name choice on their part though.

tests/test_trigger.py
MHajoha: Yeah, a lot of people have a lot of opinions on where and to what degree code should be object-oriented.

I personally believe it's great and powerful, but is often overused.

tests/test_trigger.py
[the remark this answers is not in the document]
theunkn0wn1: https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions

Not sure why, but this is a design decision on their part 😕

tests/test_trigger.py
theunkn0wn1: Well, i guess its vestigial, a left over from using `unittests` for testing as it required tests be instance methods of a class based on `unittest.TestCase`.

tests/test_trigger.py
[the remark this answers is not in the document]
theunkn0wn1: Also one small request, please rename this to `bot_fx`,
I am leaning toward suffixing fixture factories with fx to signify they are not just a normal old function.

That is the naming convention I am using for the other Test Fixtures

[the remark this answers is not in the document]
MHajoha: aight, that's sensible

theunkn0wn1:
LGTM

PR #10, opened 2018-04-27
Add method of tracking Rescue properties through changes

theunkn0wn1:
This PR provides an override implementation for `Rescue.__setattr__()`.

The override is necessary to be able to (cleanly) track attribute changes of a Rescue object as they are updated, storing them in the instance attribute `changedAttr` for later use.

Requested by MHajoha to aid with API integration.

### DIFF
--- a/Modules/rats.py
+++ b/Modules/rats.py
@@ -38,7 +38,7 @@ class Rats(object):
     """Cache of rat objects by their UUID"""
     cache_by_name = {}
     """Cache of rat objects by rat name (str)"""
-    apiHandler = None
+    api_handler = None
     """API handler"""
 
     def __init__(self, uuid: UUID, name: str = None, platform: Platforms = Platforms.DEFAULT):
@@ -190,8 +190,8 @@ class Rats(object):
             found = Rats.cache_by_name[name]
         except KeyError:
             # no such rat in cache
-            if cls.apiHandler is not None:
-                found = await cls.apiHandler.someApiCall(name=name)  # pragma: no cover
+            if cls.api_handler is not None:
+                found = await cls.api_handler.someApiCall(name=name)  # pragma: no cover
                 # FIXME: replace SomeApiCall with the actual call once we have the interface
             return found
         else:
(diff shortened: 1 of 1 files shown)


Modules/rat_rescue.py
```diff
@@ -95,6 +96,17 @@ def __init__(self, case_id: UUID, client: str, system: str, irc_nickname: str,
...
+        # post-init cleanup
+        self.changedAttr = []
+
+    def __setattr__(self, key, value):
+        """set an attribute"""
+        # call the super to render effect
+        super().__setattr__(key, value)
+        # and do a sanity check to prevent recursion (Bad)
+        if key not in self.changedAttr and key != 'changedAttr':
+            self.changedAttr.append(key)
```
MHajoha: You'll find that this is the perfect use case for a set, as you would not have to check whether it's already in the set.

theunkn0wn1: You know what you are absolutely right. I completely overlooked that property of Sets.

Modules/rat_rescue.py
```diff
@@ -95,6 +96,15 @@ def __init__(self, case_id: UUID, client: str, system: str, irc_nickname: str,
         self._lang_id = lang_id
         self._status = status
 
+        # post-init cleanup
+        self.changedAttr.clear()
```
MHajoha: lower case and underscores m8.
Yeet.
Is that how the kids talk nowadays?

UncleClapton: camelCase > underscore_delimited :^)

MHajoha: I'm a bit on the fence.
I find lowercase with underscores more easily readable, but camelCase often shorter and quicker to type.

MHajoha: But underscores are Python convention

UncleClapton: Yeah I'm mainly just joking around. :P

MHajoha: I figured ;D

theunkn0wn1: honestly, i far prefer camelCase, but we are a bit too far down the snake_case rabit hole ;(

PR #12, opened 2018-04-28
Implement ability to fetch a rat by uuid alone

theunkn0wn1:
This PR adds a method `Rats.get_rat_from_uuid` and refactors `Rats.get_rat` into `Rats.get_rat_by_name`
Both methods are now async due to the API calls, associated tests have been updated respectively.

Both new functions will first check their respective client-side Rats class attribute caches and, in the event of a cache miss, make an API query (TODO: requires API interface) to fetch the rat from the API.

This PR also Adds `Rats.__eq__`, and fixes a couple elusive bugs in Rats and their associated tests. `RatGood_fx` now uses hard-coded UUIDs to remove randomness from that part of the test suite, this helped simplify debugging of a bigger problem (which is why I implemented a `Rats.__eq__` method.)
Further, the scope of `RatGood_fx` has been lowered back to test level, as the underlying bug that required them to be at a higher scope has been fixed. Woops.

### DIFF
--- a/Modules/rats.py
+++ b/Modules/rats.py
@@ -165,7 +165,7 @@ class Rats(object):
     @classmethod
     async def get_rat_by_name(cls, name: str,
                               platform: Platforms = Platforms.DEFAULT,
-                              bot=None) -> 'Rats' or None:
+                              ) -> 'Rats' or None:
         """
         Finds a rat by name and optionally by platform
 
--- a/Modules/rats.py
+++ b/Modules/rats.py
@@ -177,6 +177,7 @@ class Rats(object):
             # no such rat in cache
             if cls.apiHandler is not None:
                 found = await cls.apiHandler.someApiCall(name=name)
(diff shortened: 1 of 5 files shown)


Modules/rats.py
```diff
@@ -162,15 +183,45 @@ def get_rat(cls, name: str, platform: Platforms = Platforms.DEFAULT) -> 'Rats' o
...
+            Rats: found Rescue
+        """
+        if not isinstance(uuid, UUID):
+            raise TypeError
+        else:
+            found = None
+            if uuid in cls.cache_by_id:
+                found = cls.cache_by_id[uuid]
+            elif cls.apiHandler is not None:  # pragma: no cover
+                found = await cls.apiHandler.someApiCall(uuid=uuid)
```
MHajoha: I can tell you right now that the signature is `async def get_rat_by_id(self, id: Union[str, UUID]) -> Rats`

PR #14, opened 2018-04-30
Updated default for Rescue.title in the Rescue constructor

theunkn0wn1:
This PR changes the default value for the `Rescue.title` property to None, rather than empty string.

I have also updated the type hints accordingly, trying something a little different by using the Typing library. feedback would be appreciated

### DIFF
--- a/Modules/rat_rescue.py
+++ b/Modules/rat_rescue.py
@@ -40,7 +40,7 @@ class Rescue(object):
                  active=True,
                  quotes: list = None,
                  epic=False,
-                 title: Union(str, None)= None,
+                 title: Union[str, None]= None,
                  first_limpet: UUID or None = None,
                  board_index: int = None,
                  mark_for_deletion: dict or None = None,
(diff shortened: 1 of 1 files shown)


Modules/rat_rescue.py
```diff
@@ -83,7 +94,7 @@ def __init__(self, case_id: UUID, client: str, system: str, irc_nickname: str,
         self._epic: bool = epic
         self._codeRed: bool = code_red
         self._outcome: None = None
-        self._title: str = title
+        self._title: Union(str, None) = title
```
MHajoha: This might just be a typo, but I believe `Union[str, None]` is what you need here like you have it in the signature. Not actually sure what `Union(str, None)` would do.

theunkn0wn1: That was a typo i thought i fixed. but yes, `()` will cause unintended behavior (raise an exception) which is why i had to fix it in 6aefa192.

PR #15, opened 2018-05-02
Database Manager

beepbeat:
simple Database wrapper to interact with an PSQL DB, hopefully async, using ODBC. Test pass now, none are skipped. Hope I did not forget something, but unkn0wn always finds something.


Modules/database_manager.py
```diff
@@ -0,0 +1,232 @@
...
+                raise ValueError(f"Connector {connector} not supported")
+
+            condition_string: str = ""
+            for k, v in condition.items():
+                condition_string += f" {k}='{v}' {connector}"
+            if len(condition) > 0:
+                condition_string = condition_string[:-len(connector)]
+                condition_string = f" WHERE {condition_string}"
+
+            return self.connection.execute("SELECT * FROM '{tablename}' {cond_str}".
```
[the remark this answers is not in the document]
MHajoha: This might well be a concern, though I believe the arguments should be checked in get_fact

Modules/database_manager.py
```diff
@@ -0,0 +1,232 @@
...
+            if not len(condition) > 0:
+                raise ValueError("No conditions were given!")
+
+            connector = connector.upper()
+            if connector not in ["AND", "OR"]:
+                raise ValueError(f"Connector {connector} not supported")
+
+            condition_string: str = ""
+            for k, v in condition.items():
+                condition_string += f"{k}='{v}' {connector} "
```
MHajoha: `", ".join(Iterable)` might clarify these pieces of code

Modules/database_manager.py
```diff
@@ -0,0 +1,232 @@
...
+            condition_string = condition_string[:-(len(connector) + 1)]
+
+            self.connection.execute(f"DELETE FROM '{table_name}' WHERE {condition_string}")
+
+    def get_fact(self, name: str, preferred_lang: str):
```
MHajoha: Could you hint the return type in the signature as well? `-> SomeType`

Modules/database_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        Return the fact for the given fact name and language.
+        If the preferred_language doesnt exist, it returns the fact for the english language.
+        If neither exist, it returns an empty string
+        Args:
+            name(str): name of the fact
+            preferred_lang (str): preferred language
+
+        Returns(str): fact for the given language, the english language or an empty string, if neither exist.
```
MHajoha: I believe the format here needs to be:
```
Returns:
    str: fact for the given language, the english language or an empty string, if neither exist.
```

theunkn0wn1: Yup, the return type and meaning go on a newline

Modules/database_manager.py
theunkn0wn1: Good catch, please do include a return type hint in the signature

Modules/database_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+This module is built on top of the Pydle system.
+"""
+import sqlite3
+
+
+class DataBaseManager:
+    instance = None
+
+    def __init__(self, file_name: str = "mecha3.db"):
```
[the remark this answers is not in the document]
theunkn0wn1: You can create a DataBase class and put your configs inside that, we shouldn't end up with a merge conflict.

tests/test_database_manager.py
```diff
@@ -0,0 +1,73 @@
...
+
+See LICENSE.md
+
+This module is built on top of the Pydle system.
+"""
+from unittest import TestCase, skip
+from Modules.database_manager import DataBaseManager
+
+
+class TestDB(TestCase):
```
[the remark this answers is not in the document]
theunkn0wn1: No problem, thats why I am informing you now. Thankfully you only have a handful of tests so it will be relatively easy to convert them. The key difference between `unittest` and `pytest` is that PyTests do not need to be in a class (as unittest requires tests be in a subclass of TestCase), and that PyTests make use of bare assertions. 

While `unittest` designed tests will work under pytest, it is undesirable to add additional tests targeting that framework as you cannot make good use of PyTests advanced functionalities.

Modules/database_manager.py
beepbeat: Done

Modules/database_manager.py
beepbeat: Done

Modules/database_manager.py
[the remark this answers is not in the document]
beepbeat: Done

Modules/database_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+            Returns: None
+            """
+            if self._has_table(name):
+                raise ValueError(f"Table {name} already exists")
+            type_string: str = ""
+            for k, v in types.items():
+                type_string += f" {k} {v.upper()},"
+            type_string = type_string[:-1]
+            self.connection.execute("CREATE TABLE {name} ({t_string})".format(name=name,
```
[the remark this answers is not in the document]
beepbeat: Done afap

tests/test_database_manager.py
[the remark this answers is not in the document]
beepbeat: Hopefully done

Modules/database_manager.py
```diff
@@ -0,0 +1,232 @@
...
+            name(str): name of the fact
+            preferred_lang (str): preferred language
+
+        Returns(str): fact for the given language, the english language or an empty string, if neither exist.
+
+        """
+        result = self.instance._select_rows("mecha3-facts", "AND", {"lang": preferred_lang, "fact":name})
+        if len(result) <= 0:
+            result = self.instance._select_rows("mecha3-facts", "AND", {"lang": "en", "fact":name})
+        return result[0][2] if len(result) > 0 else ""
```
MHajoha: Am I right in assuming that this returns an empty string when no fact could be found?
Because I'm not sure that's so great. Raising an exception might be better so that the caller can wrap the whole shebang in a try except block without worrying too much about return value.
Though returning None could also be alright. Just something more concrete than an empty string.

beepbeat: You are not only right in assuming it does, it actually does in a documented fashion 😛

MHajoha: That's good, I hadn't noticed that on my small phone screen.
However it only answers the first part of my comment. I don't believe returning an empty string is very concise, though this is probably a matter of opinion.

theunkn0wn1:
general reminder: please update your branch at some point with the upstream 😉

beepbeat:
general reminder: let me fix stuff before moaning about unimportant things :wink:

.circleci/config.yml.old
```diff
@@ -0,0 +1,73 @@
+# Python CircleCI 2.0 configuration file
```
theunkn0wn1: do we need to keep the old configuration stored in git? its contained in the history of the repo

Modules/database_manager.py
```diff
@@ -0,0 +1,155 @@
...
+# noinspection SqlNoDataSourceInspection
+class DatabaseManager(metaclass=Singleton):
+
+    def __init__(self):
+        """
+        Connects to the DB, sets up the connection, retrieves the cursor.
+        Creates the default tables should they not exist.
+        """
+        # connect to PostgreSQL (PSQL) database
+        # FIXME: Insert actual credentials / read from Config
```
theunkn0wn1: this fixme still valid?

Modules/database_manager.py
```diff
@@ -0,0 +1,155 @@
...
+        self.connection.maxwrite = 1024 * 1024 * 1024  # Again, increase this for faster access
+        self.cursor = self.connection.cursor()
+        # Ensure all the default tables are defined
+        self.cursor.execute("CREATE TABLE IF NOT EXISTS"
+                            " mecha3_facts (name VARCHAR, lang VARCHAR, response VARCHAR);")
+
+    async def _select_rows(self, table_name: str, connector: str, condition: dict = None,
+                           skipdouble_dash_test = False) -> list:
+        """
+        returns all rows matching the cconditions in the given table
```
theunkn0wn1: Docstring should be in the google format.

```py
"""
returns all rows matching the conditions in the given table

Args:
    table_name(str): description
....
```

requirements.txt
```diff
@@ -1,6 +1,9 @@
...
-pytest
-pytest-cov
-pytest-asyncio
-pure-sasl
-coloredlogs
+-e git://github.com/Shizmob/pydle.git@asyncio#egg=pydle
+pytest
+pytest-cov
+pytest-asyncio
+aiounittest
```
theunkn0wn1: `aiounittest` is no longer used, and should not be listed in requirements.txt

tests/test_database_manager.py
```diff
@@ -0,0 +1,93 @@
...
+
+pytestmark = pytest.mark.asyncio
+
+
+# noinspection PyProtectedMember
+class TestStuff(object):
+    manager = DatabaseManager()
+
+    @classmethod
+    def setup_class(cls):
```
theunkn0wn1: We can make use of PyTest's fixtures to do class/function setup routines. using `setup_class` is the legacy method

.circleci/config.yml.old
beepbeat: nope

Modules/database_manager.py
beepbeat: nope

tests/test_database_manager.py
beepbeat: done
hopefully

Modules/database_manager.py
```diff
@@ -0,0 +1,159 @@
...
+            cond_str = cond_str[0:-len(connector) - 1]
+            if ("--" in cond_str) and not skipdouble_dash_test:
+                raise ValueError("Suspicion of SQL-Injection. Aborting")
+            return self.cursor.execute(f"SELECT * FROM {table_name} WHERE {cond_str};").fetchall()
+        else:
+            raise ValueError(f"Table {table_name} does not exists!")
+
+    async def _has_table(self, name: str) -> bool:
+        """
+        checks whether the table exists.
```
theunkn0wn1: please update all docstrings to the Google Format.

on a side note, you *may* want to change pycharm / your IDE to generate Google Docstrings rather than the default RST ones

Modules/database_manager.py
```diff
@@ -0,0 +1,159 @@
...
+            val_str = ""
+            for k, v in values.items():
+                val_str += f"{k} = '{v}',"
+            val_str = val_str[:-1]
+            cond_str = ""
+            for k, v in condition.items():
+                cond_str += f"{k} = '{v}' {connector}"
+            cond_str = cond_str[0:-len(connector) - 1]
+            if ("--" in cond_str or "--" in val_str) and not skipdouble_dash_test:
+                raise ValueError("Suspicion of SQL-Injection. Aborting")
```
theunkn0wn1: can we also write a log.CRITICAL message, to ensure we get a log entry of this?

postgres-files/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# Use alpine image
+FROM postgres:alpine
+# Bind a directory to store PostGres data and databases
+ENV PGDATA /psql-data
+COPY ./psql-startup.sh ./
+RUN chmod +x ./psql-startup.sh
```
theunkn0wn1: missing newline at EOF

postgres-files/psql-setup.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/sh
+pg_ctl -w start
+psql -c 'CREATE ROLE mecha3 LOGIN CREATEDB;'
+psql -c 'CREATE DATABASE mecha3 WITH OWNER = mecha3;'
```
theunkn0wn1: missing newline at EOF

postgres-files/psql-startup.sh
```diff
@@ -0,0 +1,3 @@
+su postgres -c './docker-entrypoint.sh'
+su postgres -c "psql -c 'CREATE ROLE mecha3 LOGIN CREATEDB;'"
+su postgres -c "psql -c 'CREATE DATABASE mecha3 WITH OWNER = mecha3;'"
```
theunkn0wn1: missing newline at EOF

config/configuration.md
```diff
@@ -74,4 +74,15 @@ API configuration elements
 |--------|-------------|
 |online_mode|should mecha start up in online mode?|
 |url|base url for the API|
-|tokenfile|name of the API token file, relative to `certs/`|
\ No newline at end of file
+|tokenfile|name of the API token file, relative to `certs/`|
+
+# DataBase
```
theunkn0wn1: Suggested changes:
```md
# Database
Configures the connection to the PostgreSQL database

| Element| description |
|------- |------------ |
|server| ip address or hostname for the DB server|
|port| port to connect to|
|database| name of the database to use|
|username| name of the database user|
|password| password for the database user|
```

tests/test_database_manager.py
```diff
@@ -0,0 +1,96 @@
...
+    async def test_create_table(self):
+        if await self.manager._has_table("testtablecreate"):
+            await self.manager._drop_table("testtablecreate")
+        if not await self.manager._has_table("testtablecreate"):
+            await self.manager._create_table("testtablecreate", {"string1": "VARCHAR"})
+        with pytest.raises(ValueError):
+            await self.manager._create_table("testtablecreate", {"test": "VARCHAR"})
+
+    async def test_insert_row(self):
+        if not await self.manager._has_table("testtableinsert"):
```
theunkn0wn1: Test does not have an assert?

tests/test_database_manager.py
```diff
@@ -0,0 +1,96 @@
...
+            await self.manager._drop_table("testtableinsert")
+
+    async def test_update_row(self):
+        if not await self.manager._has_table("testtableupdate"):
+            await self.manager._create_table("testtableupdate", {"string1": "VARCHAR"})
+        await self.manager._insert_row("testtableupdate", ("thest",))
+        await self.manager._update_row("testtableupdate", "AND", {"string1": "FizzBuzz"},
+                                       {"string1": "thest"})
+
+        assert \
```
theunkn0wn1: multiple statements on one continued line. please consider assigning the output of the awaited call to a var, then asserting against that.

PR #17, opened 2018-05-03
Make Trigger and Permission objects hashable and more or less immutable (SPARK-12)

MHajoha:
This PR implements `__hash__` methods for the Trigger and Permission objects.
It also implements an `__eq__` method for the Trigger object and makes both objects relatively immutable.

The Trigger object's `__hash__` simply xors the hashes of its key attributes together, but it only does it once when the hash is first needed. The Permission's `__hash__` just uses the permission level to be in line with its `__eq__` method.

I likely won't give the API handlers `__eq__` and `__hash__` methods because, on seconds thought, that's pretty fucking pointless.

### DIFF
--- a/Modules/trigger.py
+++ b/Modules/trigger.py
@@ -112,7 +112,7 @@ class Trigger(object):
             else:
                 return True
         else:
-            return super().__eq__(other)
+            return False
 
     def __hash__(self) -> int:
         if self._hash is None:
--- a/Modules/trigger.py
+++ b/Modules/trigger.py
@@ -100,6 +100,19 @@ class Trigger(object):
         """
         await self.bot.message(self.channel if self.channel else self.nickname, msg)
 
+    def __eq__(self, other):
+        if self is other:
+            return True
+        elif isinstance(other, Trigger):
+            for name, value in Trigger.__dict__.items():
+                if isinstance(value, property):
+                    if getattr(self, name) != getattr(other, name):
+                        return False
+            else:
+                return True
+        else:
+            return super().__eq__(other)
+
     def __hash__(self):
         if self._hash is None:
             attrs = (self._words_eol[0], self._nickname, self._target, self._ident, self._hostname,
--- a/Modules/trigger.py
+++ b/Modules/trigger.py
@@ -1,7 +1,8 @@
+from functools import reduce
+from operator import xor
 from typing import List
 
 import pydle
-import sys
 
 
 class Trigger(object):
@@ -100,7 +101,7 @@ class Trigger(object):
         """
         await self.bot.message(self.channel if self.channel else self.nickname, msg)
 
-    def __eq__(self, other):
+    def __eq__(self, other) -> bool:
         if self is other:
             return True
         elif isinstance(other, Trigger):
@@ -113,16 +114,10 @@ class Trigger(object):
         else:
             return super().__eq__(other)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         if self._hash is None:
             attrs = (self._words_eol[0], self._nickname, self._target, self._ident, self._hostname,
                      self._realname, self._away, self._account)
-
-            offset = sys.hash_info.width
-            interval = offset // len(attrs)
-            self._hash = 0
-            for attr in attrs:
-                offset -= interval
-                self._hash |= hash(attr) << offset
+            self._hash = reduce(xor, map(hash, attrs))
 
         return self._hash
--- a/Modules/trigger.py
+++ b/Modules/trigger.py
@@ -1,6 +1,7 @@
 from typing import List
 
 import pydle
+import sys
 
 
 class Trigger(object):
@@ -44,6 +45,8 @@ class Trigger(object):
         self.account = account
         self.identified = identified
 
+        self._hash = None
+
     @classmethod
     def from_bot_user(cls, bot: pydle.BasicClient, nickname: str, target: str, words: List[str],
                       words_eol: List[str] = None) -> 'Trigger':
@@ -83,3 +86,17 @@ class Trigger(object):
             msg (str): Message to send.
         """
         await self.bot.message(self.channel if self.channel else self.nickname, msg)
+
+    def __hash__(self):
+        if self._hash is None:
+            attrs = (self.words_eol[0], self.nickname, self.target, self.ident, self.hostname,
+                     self.realname, self.away, self.account)
+
+            offset = sys.hash_info.width
+            interval = offset // len(attrs)
+            self._hash = 0
+            for attr in attrs:
+                offset -= interval
+                self._hash |= hash(attr) << offset
+
+        return self._hash
--- a/Modules/trigger.py
+++ b/Modules/trigger.py
@@ -33,19 +33,32 @@ class Trigger(object):
                 nickname.
             identified (bool): Whether or not the user is identified with NickServ.
         """
-        self.bot = bot
-        self.words = words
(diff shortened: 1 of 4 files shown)


Modules/trigger.py
```diff
@@ -83,3 +100,24 @@ def channel(self) -> str or None:
...
+        if self is other:
+            return True
+        elif isinstance(other, Trigger):
+            for name, value in Trigger.__dict__.items():
+                if isinstance(value, property):
+                    if getattr(self, name) != getattr(other, name):
+                        return False
+            else:
+                return True
+        else:
```
theunkn0wn1: If `other` is not of Type Trigger this should raise a TypeError rather than calling the super

Modules/trigger.py
```diff
@@ -32,17 +34,32 @@ def __init__(self, bot: pydle.BasicClient, words: List[str], words_eol: List[str
...
+        self._hash = None
+        self._immutable = True
+
+    bot = property(lambda self: self._bot)
+    words = property(lambda self: self._words)
+    words_eol = property(lambda self: self._words_eol)
+    nickname = property(lambda self: self._nickname)
+    target = property(lambda self: self._target)
+    ident = property(lambda self: self._ident)
+    hostname = property(lambda self: self._hostname)
```
theunkn0wn1: Question: does this create a read-only property? we may want to be able to change these values at runtime

Modules/trigger.py
```diff
@@ -83,3 +100,24 @@ def channel(self) -> str or None:
             msg (str): Message to send.
         """
         await self.bot.message(self.channel if self.channel else self.nickname, msg)
+
+    def __eq__(self, other) -> bool:
+        if self is other:
+            return True
+        elif isinstance(other, Trigger):
+            for name, value in Trigger.__dict__.items():
```
theunkn0wn1: Why are we calling a magic rather than comparing the relevant attributes by name?

Modules/trigger.py
MHajoha: Unless there is convention which I am not aware of, there should not be a problem with this.
By definition, calling the super follows the convention of returning `False` if the other object is of a different type which Python's buitins follow.
An argument could be made that it should just return `False` without asking the super, since in the case of subclassing object, the two are equivalent.

Modules/trigger.py
MHajoha: Omitting the setter makes the property read-only. 
Though technically one could write to the private attribute directly.

theunkn0wn1: Of course we could, but that would be bad practice to write to a private directly. What im saying here is we may want to change some of these properties at runtime therefore should not limit them to read-only

Modules/trigger.py
MHajoha: Short answer: I'm lazy and this object has a bunch of attributes.
Better answer: As far as I'm aware, `type.__dict__.values()` is the only way of getting an iterator over all the properties of the class.

MHajoha: Accessing magic isn't in itself a warcrime, either. Many of the functions which wrap it simply provide extra functionality, so they should be used when available.

Modules/trigger.py
MHajoha: The Trigger object, as it currently is, is designed to provide a command with information on how and by whom the command was invoked. This should be a snapshot of the state at the point of invocation.
e.g. if a user changes their nickname during execution of a command by them, the nickname in the trigger object should not be updated
The same goes for the other user attributes and even more strongly for words and words_eol.

MHajoha: Long story short, the Trigger object is a prime example of an object that really ought to be immutable.

The alternative could lead to bad things when a command function suddenly finds its Trigger object changed in such a way that it no longer matches the state of the command function.

theunkn0wn1: Nope, trigger should be immutable, i misread.

beepbeat: If my understanding of Trigger objects being invalidated after command execution has finished, write-access after initializing them should be forbidden, due to the fact that information should be reliable  and they are also short-lived, as on the next command a new Trigger would be constructed, reflecting any change that might have happened in the mean time.

I fully agree with MH here

theunkn0wn1: YEah. my original comment was based in a misunderstanding on my part, Trigger objects should be immutable after creation, and should not be able to modify the bot or its context.

Modules/trigger.py
theunkn0wn1: The two methods would be equal, but i see no reason to invoke the super in this context. would it not be more efficient to simply return false, since the super call is both unnecessary and yield the same result?

tests/test_trigger.py
```diff
@@ -63,3 +63,36 @@ def test_create_from_user_query(bot_fx):
...
+                       "not_a_channel", "test_ident", "test.vhost")
+
+    assert trigger1 == trigger1
+    assert trigger2 == trigger2
+
+    assert trigger1 == trigger2
+    assert trigger2 == trigger1
+
+def test_trigger_ne(bot_fx):
+    trigger1 = Trigger(bot_fx, ["some", "thing"], ["some thing", "thing"], "test_nick",
```
theunkn0wn1: may i suggest providing `trigger` as a fixture? would reduce duplicity. Failing that you could paramatrize this, so you can test with various inputs rather than a single pair of hardcodes

tests/test_trigger.py
```diff
@@ -63,3 +63,36 @@ def test_create_from_user_query(bot_fx):
...
+                       "not_a_channel", "test_ident", "test.vhost")
+    trigger2 = Trigger(bot_fx, ["some", "thing"], ["another thing", "thing"], "stupid_nick",
```
theunkn0wn1: can we avoid vulgarity in our test suites? keep it PG 😉

tests/test_trigger.py
```diff
@@ -63,3 +63,36 @@ def test_create_from_user_query(bot_fx):
...
+                       "not_a_channel", "test_ident", "go.away")
+
+    assert trigger1 != trigger2
+    assert trigger2 != trigger1
+
+def test_trigger_hash():
+    trigger1 = Trigger(None, ["some", "thing"], ["some thing", "thing"], "test_nick",
+                       "not_a_channel", "test_ident", "test.vhost")
+    trigger2 = Trigger(None, ["some", "thing"], ["some thing", "thing"], "test_nick",
```
theunkn0wn1: Methinks 
```py
trigger2 = deepcopy(trigger1)
```
would be a more effective solution, rather than a copy paste

Modules/trigger.py
theunkn0wn1: _grumbles_

tests/test_trigger.py
MHajoha: I honestly don't see `go.away` and `stupid_nick` not being PG

tests/test_trigger.py
MHajoha: I don't think there's justification for making this a fixture, since I need a few different Trigger objects and they likely wouldn't be used in any other tests.
Parametrized tests wouldn't really increase test coverage much, but definitely wouldn't hurt...

Modules/trigger.py
MHajoha: Hmm, alrighty

tests/test_trigger.py
MHajoha: That would work, though these are just tests so performance isn't really an issue. And I believe this to be more verbose. Potatoes potahtoes though.

PR #19, opened 2018-05-11
SPARK-18

theunkn0wn1:
- [x] context
Implements features of SPARK-18


Modules/diagnostics/Diagnostics.py
```diff
@@ -0,0 +1,71 @@
...
+from Modules.commandcontext import CommandContext
+from Modules.permissions import require_permission
+from Modules.rat_command import Commands
+
+log = logging.getLogger(f'{config.Logging.base_logger}.handlers')
+
+
+@require_permission(permissions.TECHRAT)
+@Commands.command("debug-fetch")
+async def cmd_testNick(bot, trigger: CommandContext):
```
MHajoha: I believe we were chatting once about keeping the actual commands separate from the infrastructure.
In that spirit, perhaps a better place for this file would be in `/commands/debug.py`?

Modules/commandcontext.py
```diff
@@ -0,0 +1,104 @@
+from functools import reduce
+from operator import xor
+from typing import List, Union
+
+import pydle
+
+from Modules.user import User
+
+
+class CommandContext(object):
```
MHajoha: We will be throwing these objects around a lot. Perhaps `Context` would be much shorter and quicker to use while not sacrificing conciseness?

Modules/commandcontext.py
```diff
@@ -0,0 +1,104 @@
...
+    """MechaClient instance"""
+    words: [str] = property(lambda self: self._words)
+    words_eol: list = property(lambda self: self._words_eol)
+    """words deliminated to EOL"""
+    user: User = property(lambda self: self._user)
+    """invoking IRC user"""
+
+
+    @classmethod
+    async def from_bot_user(cls, bot: pydle.BasicClient, nickname: str, target: str,
```
MHajoha: With the (very good) `User.from_bot` method existing, there might not be much use in this one, as it doesn't serve much use over the constructor. The `User` object could be (and would probably make more sense to be) created by the caller.

Modules/user.py
```diff
@@ -0,0 +1,139 @@
...
+        # sets the permission based on the hostmask
+        # which requires stripping the username and the leading period
+        try:
+            # FIXME: access to protected member
+            # loop over every known vhost
+            for key, value in permissions._by_vhost.items():
+                # if a vhost matches, and the user is identified
+                if hostname.endswith(key) and identified is True:
+                    self._permission_level = value
+        except KeyError:
```
MHajoha: Hey, I see you added the permission level as a property of `User`. I have two concerns with this:

1. As you point out yourself, this accesses the private `_by_vhost` directly. Simply creating a public, top-level function `has_level(vhost: str, level:Permission)` in `permissions.py` will easily solve this.

2. Why not just leave it as it was? Letting the `require_permission` decorator: a) worked fine, so why try (and fail) to fix it? b) avoids adding another public member to the permissions module

I suppose there could at some point be some sort of reason for a command function to inquire the users permission after it has been called, however I can't come up with a circumstance for this and, as you were so intent on in the `Rescue.outcome` discussion, where our opinions were reversed, why build what we don't need (and will likely never need)?

This also brings up another, deeper problem I see in development style (and this is not at all the only instance of this): Adding TODOs and FIXMEs to code which you know to be broken does not make it any better. In fact, it gives an impression, to me at least, that you didn't care to fix the issue itself. You moved the JIRA story on this to 'In review'. That leads me to believe that you consider this done and up for merging. So why is it still broken, especially when the solution is not that hard? I ask you to strive towards cleaner and better polish here. If not before pushing, as I believe would be ideal, then before peer review.

I hope you don't see me saying that as rude, because I don't want it to be and I have tried to keep it polite, however I also don't believe in keeping quiet about these things as this does bug me.

theunkn0wn1: It was a "this might be useful" kinds of things, to which i forgot was already implemented elsewhere. You are absolutely right for calling me on this. the `require_permission` decorators job is to handle user permissions, i should not be reinventing the wheel.

It is that i do not care to fix a problem itself, more that i know its broken but i cannot fix it without things that do not yet exist. I have reviewed all my TODOs and FIXMEs and all bot two of them require at least the API interface to exist. (after removing items that are no longer valid.) 
It is NOT that i do not care to fix issues, it is that i know there are issues that i cannot resolve without making fatal assumptions dependent on features that do not exist within the codebase (yet). Sure i can peak over to your branch and write the associated call, but i would rather wait until your branch is PRed in so you can make any changes you need prior to pushing without needing to inform you made a breaking change.

AS for why i marked the feature as "in review" as the first part of the large story (the Context object) is what i consider to be ready for review. Granted i should have created a task rather than moving the whole story, but i was under the impression the changes i made to the title and description of this PR were sufficient to communicate that fact. Evidently i am mistaken and i took the incorrect path to expressing my intentions.

corrections incoming.

main.py
```diff
@@ -51,85 +62,115 @@ class MechaClient(Client):
...
     """
-    log.warning(f"cmd_ping triggered on channel '{trigger.channel}' for user "
+    LOG.warning(f"cmd_ping triggered on channel '{trigger.channel}' for user "
                 f"'{trigger.nickname}'")
     await trigger.reply(f"{trigger.nickname} pong!")
 
 
+@require_permission(permissions.RAT)
+@Commands.command("version", "potato", "ver")
+async def cmd_version(bot, trigger: Context):
```
MHajoha: Maybe these should also be moved to the `commands` package?

Modules/permissions.py
```diff
@@ -115,15 +120,20 @@ def require_permission(permission: Permission,
...
-        async def guarded(bot, trigger):
-            if trigger.identified and trigger.hostname in _by_vhost.keys() \
-                    and _by_vhost[trigger.hostname] >= permission:
-                return await func(bot, trigger)
+        async def guarded(bot, context: 'Context'):
+            log.debug(f"context = {context}")
+            granted_permissions = NONE
+            for key, value in _by_vhost.items():
+                if context.user.hostname.endswith(key) and context.user.identified is True:
+                    granted_permissions = value
```
MHajoha: This method of iterating over `_by_vhost` is imperfect. Simply grabbing `_by_vhost[context.user.hostname]` if it exists worked fine, and should be more clear and be more performant.

Asking whether the user's hostname ends with the vhost is fragile in any event, as, for example, this would cause overlap between `techrat.fuelrats.com` and `rat.fuelrats.com`, which would lead to a sever bug of the techrat not being able to execute a command with a required permission level higher than `RAT`, since the rat level comes before the techrat level in `_by_vhost` and python dicts are basically ordered these days.

beepbeat: ``` granted_permissions = value if (not granted_permissions or granted_permissions < value) ``` should always give you the highest eligible permission level, assuming I didn't messed up that python

main.py
```diff
@@ -51,85 +62,115 @@ class MechaClient(Client):
...
 # entry point
 if __name__ == "__main__":
-    log.info("hello world!")
+    LOG.info("hello world!")
 
-    pool = ClientPool()
-    log.debug("starting bot for server...")
+    POOL = ClientPool()
+    LOG.debug("starting bot for server...")
     try:
```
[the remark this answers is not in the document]
theunkn0wn1: According to Pylint, these are constants. The value is set during load but is not changed at runtime. methinks thats the reasoning behind treating this as a constant

main.py
theunkn0wn1: I want at least the version / ping command to remain in main, as an indication that Mechasqueak started up OK

Modules/permissions.py
theunkn0wn1: it does not work by grabbing the hostname, as the hostname is more than just the host mask. 
Normal users' hostname looks something like `{name}.{ident}@{role}.fuelrats.com` and admins are `{name}@{role}.fuelrats.com`. 

Therefore, this is the simplest method was to itterate over the known hosts and checking if the hostname ends with a mask.

main.py
beepbeat: I agree with having some kind of ping capability that does not rely on other modules being loaded

Modules/context.py
```diff
@@ -0,0 +1,116 @@
...
+        self._words = words
+        self._words_eol = words_eol
+        self._user = user
+        self._hash = None
+        self._target = target
+
+    bot: 'MechaClient' = property(lambda self: self._bot)
+    """MechaClient instance"""
+    words: [str] = property(lambda self: self._words)
+    words_eol: list = property(lambda self: self._words_eol)
```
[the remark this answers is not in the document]
theunkn0wn1: lol, sure thing.

main.py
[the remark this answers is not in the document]
MHajoha: Hrmpf. These are local variables not used anywhere else. I'd like to have a word with these pylint people...

main.py
MHajoha: This still relies on other modules being loaded. The commands and permissions modules to be exact.

Modules/permissions.py
MHajoha: Err, we're both half right. You're right, I forgot that what I assume to be a person's fuelrats.com username being part of their vhost. `_by_vhost[hostname]` won't work as I had hoped. However what you call `name`, generally called ident or username is not part of the hostname and (unless pydle's whois does some really strange things) is stored seperately in `user.username`.

Either way, there is a severe bug here which needs to be fixed. I see the simplest solution as not breaking when a match has been found or, slightly more performantly, reversing the `_by_vhost` dict (and maybe also making it a list as we don't need it to be a dict any longer and they do come with some overhead). This would ensure that the highest permission level which passed the `endswith(vhost)` test is always chosen. This is still somewhat fragile though, as if we ever come into the situation of a lower permission level ending with the vhost of a higher one, this would allow the lower-permissioned rat to attain the higher permission level.

An entirely different approach, which would likely be much more concise and clean, would be to include information of whether the vhost is preceded by stuff in each `Permission` object and, if so, cut out everything up until the first dot.
```python
found_permission = None
for permission in _permission_levels: # wouldn't need _by_vhost any longer
  if permission.vhost == (user.hostname.split(".", 1)[1] if permission.prefixed_by_username else user.hostname):
    found_permission = permission
    break
```
Or, functionally, which I would prefer:
```python
try:
  permission = next(filter(lambda permission: permission.vhost == (user.hostname.split(".", 1)[1] if permission.prefixed_by_username else user.hostname), _permission_levels))
except StopIteration:
  permission = NONE
```

main.py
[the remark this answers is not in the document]
MHajoha: Either way, these values are not only local, they are also mutable and they change, they really aren't constants, no matter what pylint says.

Modules/permissions.py
UncleClapton: @MHajoha What will always remain consistent with our vhosts is that they *always* end with `{role}.fuelrats.com`.

would there be a concise way of checking for the amount of periods in the string, and based off of that either strip at the first period, or at the @ ?

If you could write it like that, you would be able to avoid writing a boolean to determine if the vhost has an ident in it.

MHajoha: Yeah that'd work, great idea. Could use `hostname.rsplit(".", 3)[1]` to get at that part. I'm all for it. @theunkn0wn1, what do you think? Might be worth a new issue.

theunkn0wn1: `str.endswith()` allows us to compare the end of the string against another string. we do not need to do any fancy splits

theunkn0wn1:
Alright the git history on this branch is giant mess. im burning this down and redoing it

PR #20, opened 2018-05-19
Make mecha obey set prefix command, make case-insensitive

nathanctech:
Hopefully I'm doing this correctly. :-) ~masterxc

### DIFF
--- a/Modules/rat_command.py
+++ b/Modules/rat_command.py
@@ -55,6 +55,13 @@ class NameCollisionException(CommandException):
     pass
 
 
+class PrefixInvalidException(CommandException):
+    """
+    Someone used the wrong prefix for a command.
+    """
+    pass
+
+
 class Commands:
     """
     Handles command registration and execution
@@ -88,7 +95,7 @@ class Commands:
             # command didn't start with prefix, ignore it
         if not message.startswith(cls.prefix):
             cls.log.debug(f"caught a command without prefix: {message}")
-            return None
+            raise PrefixInvalidException(f"I got an unknown prefix!'")
         cls.log.debug(f"triggered! message is {message}")
 
         # remove command prefix and make lowercase
@@ -132,6 +139,9 @@ class Commands:
         """
         if isinstance(names, str):
             names = [names]  # idiot proofing
+
+        # transform commands to lowercase
+        names = [x.lower() for x in names]
 
         if func is None or not callable(func):
             # command not callable
--- a/Modules/rat_command.py
+++ b/Modules/rat_command.py
@@ -93,9 +93,6 @@ class Commands:
             raise CommandException(f"Bot client has not been created or not handed to Commands.")
 
             # command didn't start with prefix, ignore it
-        if not message.startswith(cls.prefix):
-            cls.log.debug(f"caught a command without prefix: {message}")
-            raise PrefixInvalidException(f"I got an unknown prefix!'")
         cls.log.debug(f"triggered! message is {message}")
 
         # remove command prefix and make lowercase
--- a/Modules/rat_command.py
+++ b/Modules/rat_command.py
@@ -85,10 +85,14 @@ class Commands:
             # someone didn't set me.
             raise CommandException(f"Bot client has not been created or not handed to Commands.")
 
+            # command didn't start with prefix, ignore it
+        if not message.startswith(cls.prefix):
+            cls.log.debug(f"caught a command without prefix: {message}")
+            return None
         cls.log.debug(f"triggered! message is {message}")
 
-        # remove command prefix
-        raw_command: str = message.lstrip(cls.prefix)
+        # remove command prefix and make lowercase
+        raw_command: str = message.lstrip(cls.prefix).lower()
 
         words = []
         words_eol = []
--- a/Modules/rat_command.py
+++ b/Modules/rat_command.py
@@ -92,7 +92,6 @@ class Commands:
             # someone didn't set me.
             raise CommandException(f"Bot client has not been created or not handed to Commands.")
 
-            # command didn't start with prefix, ignore it
         cls.log.debug(f"triggered! message is {message}")
 
         # remove command prefix and make lowercase
--- a/Modules/rat_command.py
+++ b/Modules/rat_command.py
@@ -133,7 +133,7 @@ class Commands:
             names = [names]  # idiot proofing
 
         # transform commands to lowercase
-        names = [lowered_name.lower() for lowered_name in names]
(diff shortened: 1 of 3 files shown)


Modules/rat_command.py
```diff
@@ -85,10 +85,14 @@ class Commands:
...
+            # command didn't start with prefix, ignore it
+        if not message.startswith(cls.prefix):
+            cls.log.debug(f"caught a command without prefix: {message}")
+            return None
         cls.log.debug(f"triggered! message is {message}")
 
-        # remove command prefix
-        raw_command: str = message.lstrip(cls.prefix)
+        # remove command prefix and make lowercase
+        raw_command: str = message.lstrip(cls.prefix).lower()
```
theunkn0wn1: it amy be advisable to also cast command registrations to lower case, otherwise we can get commands not found that should be found

Modules/rat_command.py
```diff
@@ -85,10 +85,14 @@ class Commands:
             # someone didn't set me.
             raise CommandException(f"Bot client has not been created or not handed to Commands.")
 
+            # command didn't start with prefix, ignore it
+        if not message.startswith(cls.prefix):
+            cls.log.debug(f"caught a command without prefix: {message}")
+            return None
```
theunkn0wn1: This should probably raise an exception, rather than a null return.

theunkn0wn1:
Alright, well the tests are working now. You now need to write additional tests to increase code coverage to in excess of 85%. right now you are at 71.

Modules/rat_command.py
```diff
@@ -85,10 +92,14 @@ class Commands:
             # someone didn't set me.
             raise CommandException(f"Bot client has not been created or not handed to Commands.")
 
+            # command didn't start with prefix, ignore it
+        if not message.startswith(cls.prefix):
+            cls.log.debug(f"caught a command without prefix: {message}")
+            raise PrefixInvalidException(f"I got an unknown prefix!'")
```
MHajoha: Hey, good thing you added this check. (I completely forgot to reimplement that after I redid the command invocation)

However, raising an exception on expected behaviour is something that should generally be avoided. 

More importantly, this case will happen tons of times (on each message which isn't a command, likely more often than proper command invocations) and exceptions aren't the most performant thing in programming, especially when they are propagated all the way to the event loop (i.e. not caught by us anywhere) and printed, as would be the case here.

Long story short, this would likely grind Mecha to a snails pace, simply not doing anything / returning immediately would be preferable. (Doing this check in `MechaClient.on_message` would be the best solution in my opinion as it gets rid of an unnecessary function call, but that part is debatable)

nathanctech: Yeah the initial code was to just return None, but unknown wanted an exception thrown. I preferred the first method although in my testing it only fires on non-alphanumeric characters as the first character like a command trigger (so *, ., + trigger it but normal text does not) so it won't fire as often. I'm fine with either method that makes sense.

I'll correct my git settings, thanks. This is the first ever project I've contributed to so I'm learning still.

theunkn0wn1: The way i read it is this condition should not be reached, as it shouldn't get past the on_message handler. Therefore this would not be a normal runtime condition and instead an error condition.
That said, the check for the command prefix should probably occur in on_message as that would be the first step in the command execution chain.

nathanctech:
I moved the check to `on_message` since it makes the most sense there. Overhead for a single function call isn't high in the grand scheme of things but no sense processing a command that will get discard anyway.

Modules/rat_command.py
```diff
@@ -55,6 +55,13 @@ class NameCollisionException(CommandException):
     pass
 
 
+class PrefixInvalidException(CommandException):
```
theunkn0wn1: Is this still necessary? If someone uses the wrong prefix we should ignore it akin any other message coming from the pipe

Modules/rat_command.py
```diff
@@ -129,6 +136,9 @@ def _register(cls, func, names: list or str) -> bool:
         if isinstance(names, str):
             names = [names]  # idiot proofing
 
+        # transform commands to lowercase
+        names = [x.lower() for x in names]
```
theunkn0wn1: Usage of a single char variable. Given the context, im going to have to ask you to change this.

main.py
```diff
@@ -86,6 +86,11 @@ class MechaClient(Client):
             log.debug("received message from myself ignoring!.")
             return None
 
+        if not str(message).startswith(Commands.prefix):
+            # prevent bot from processing commands without the set prefix
+            log.debug(f"caught a command without prefix. Ignoring.")
```
theunkn0wn1: How do we know it is a command, specifically if it does not have the prefix?
May i suggest rewording it to something more along the lines of 
```py 
log.debug(f"ignoring message {message} as it does not start with my prefix")
```

main.py
```diff
@@ -86,6 +86,11 @@ class MechaClient(Client):
...
+            return None
```
MHajoha: Hey, thanks for checking this here now. Raising a special exception in the trigger method isn't necessary any longer, as unknown pointed out. However a small `assert message.startswith(thingy)` in the trigger method wouldn't hurt and be quite cheap. Assertions could even be disabled for production.

One last thing: while it does help with completion in ides, there's no need to explicitly call `str(message)` here, it would only copy the message string, which we don't need.
Python3(.?) does have a syntax for type hinting, much like statically typed languages. In the arguments, `message: str` will enable the ide of your choice (pycharm?) to provide code completion without a runtime hit to performance.

nathanctech: Made the changes, thanks for the tips. An assert will definitely point out any funny business with that bit of code.

main.py
nathanctech: Updated.

Modules/rat_command.py
nathanctech: Also updated, new variable is lowered_name.

Modules/rat_command.py
```diff
@@ -129,6 +132,9 @@ def _register(cls, func, names: list or str) -> bool:
         if isinstance(names, str):
             names = [names]  # idiot proofing
 
+        # transform commands to lowercase
+        names = [lowered_name.lower() for lowered_name in names]
```
MHajoha: Hey, this is a really tiny nitpick and also the last one I could find, but since you're lowering it only now, `lowered_name` seems unfitting.

theunkn0wn1: methinks ```py
names = [name.lower() for name in names
```
would be more appropriate

MHajoha: Yeah

PR #21, opened 2018-05-19
Enable Docker Support

theunkn0wn1:
This PR enables the ability for Mecha to run within a Docker Container.
Specifically, mecha3's container is build off Alpine Linux, a minimalistic Linux distribution.
The only things added to it are Git, Mecha's codebase, and her dependencies.

Docker will allow us to have a uniform testing environment, outside what Travis provides. Further, It is to my understanding mecha3 may end up running within a container when she goes live. This facilitates that intention. 

Trezy, asking you to review things as you have Docker experience more than I, please let me know if you spot something obviously flawed with my approach 😄

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,14 +18,12 @@ FROM python:3.6.5-alpine
 COPY ./requirements.txt ./
 # fetch git, as we will need it.
 RUN apk add --no-cache git
-# Copy the current directory contents into the container at /app
-ADD . /mechasqueak
 
 # Install any needed packages specified in requirements.txt
 RUN pip install --trusted-host pypi.python.org -r /requirements.txt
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ADD . /mechasqueak
+
+# Install any needed packages specified in requirements.txt
+RUN pip install --trusted-host pypi.python.org -r /mechasqueak/requirements.txt
+
+# Set the working directory to /mechasqueak
+WORKDIR /mechasqueak
+
+# Run our tests when the container launches by default
+CMD ["pytest"]
```
trezy: I don't think you need this command in your `Dockerfile` since it's also run in your `docker-compose.yml`.

PR #23, opened 2018-05-23
SPARK-19: JSON-based configuration files

theunkn0wn1:
This PR implements Spark-19, JSON based configuration files
These files will live in `config/` and are synced between the host and the containers. Changes to the configuration files on the host will cause the files on the container to be updated, there is no need to rebuild containers for configuration changes.

**Bonus features**:
This Pr also enables the EXTERNAL authentication mechanism for SASL, allowing pydle to authenticate by presenting a client-side certificate to the irc server.

To enable external authentication, drop a `.pem` file (certificate and key) into the `certs/` directory and make the associated configuration changes.
```json
  "authentication": {
    "method":"EXTERNAL",
    "plain": {
      "username": "",
      "password": "",
      "SASL": true,
      "identity": ""
    },
    "external":{
      "tls_client_key":"mecha3.pem",
      "tls_client_cert":"mecha3.pem"
    }
```
This will cause mecha to use a client cert located at `certs/mecha3.pem`

This directory is also mounted as a volume, so certs can be replaced without rebuilding the container.

Checklist:

- [x] IRC presence
- [x] logfile configuration
- [x] authentication
- [x] commands
- [x] api
- [x] documentation


config.py
```diff
@@ -1,72 +1,56 @@
...
-    """
-    Configuration options for the RatBoard
+    Args:
+        filename (str): path and filename to load.
     """
-    CASE_LIMIT = 30
-    """Default limit to board indexies, may be dynamically re-sized during runtime."""
+    global CONFIGURATION
+
+    path = f"config/{filename}"
```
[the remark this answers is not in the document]
beepbeat: It's not such a great idea to hardcode extensions, as they are not required and some may want to use the file named "config" and not "config.json".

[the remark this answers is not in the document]
nathanctech: Well, the file names would be develop.json etc and always be json format but I see your point...linux doesn't require extensions. Just encountered it during my debugging but I'll get over it. ;)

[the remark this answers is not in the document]
theunkn0wn1: Linix does not require file extensions, and it is possible to create files without extensions on windows. I do not believe it is in our best interests to require configuration files to be suffixed with `.json` due to this.

config/config.template.json
```diff
@@ -0,0 +1,34 @@
+{
+  "irc": {
+    "presence": "Mechasqueak3[BOT]",
```
MHajoha: While we're at it, can we change this to `nickname`? Presence always sounded very ominous and confused the bananas out of me at first.

config.py
```diff
@@ -1,72 +1,56 @@
...
+    # set the minimum severity the logger will report.
+    # uncomment for production:
+    # log.setLevel(logging.INFO)
+    # uncomment for develop:
+    log.setLevel(logging.DEBUG)
+
+    logging.info("[Mecha] configuration file loading...")
+    logging.basicConfig(level=logging.DEBUG)  # write all the things
+
+    """provides facilities for managing a configuration from disk"""
```
MHajoha: Stray docstring?

config.py
```diff
@@ -1,72 +1,56 @@
...
-class Logging:
-    """
-    Log configurations
+import os
+from typing import Union
+
+CONFIGURATION: Union[None, dict] = None
+
+
+def setup_logging(root_logger: str, logfile: str):
```
MHajoha: Does the logging setup belong here? I don't believe it has anything to do with the config, as such I feel it should not be in this module.

main.py
```diff
@@ -11,40 +11,28 @@
...
+
 from pydle import ClientPool, Client
+
+import config
+
+# this feels really ugly, im probably going to replace this with argparse or similar.
+if len(argv) >= 2:  # assume first argument is the config file to use
+    config_filepath = argv[1]
+else:  # we didn't get one
+    config_filepath = "./config/config.template.json"
```
MHajoha: I may be wrong, but this probably could be done in `config.py` itself, which would get rid of doing this before the imports (though the logging stuff will have to remain here) and also make it certain that the config has been read when it is imported and used.

I also agree that this should be using argparse, which is a great bit of standard library. This would make it much more expandable in case we want to add more command-line arguments in the future. Argparse also automatically provides a lovely `-h` for printing info on available arguments.

main.py
```diff
@@ -117,16 +104,44 @@ class MechaClient(Client):
...
+                     tls=CONFIGURATION['irc']['tls'])
     except Exception as ex:
-        log.error(f"unable to connect to {IRC.server}:{IRC.port} "
+        log.error(f"unable to connect to {CONFIGURATION['irc']['server']}:"
+                  f"{CONFIGURATION['irc']['port']}"
                   f"due to an error.")
         log.error(ex)
-        from sys import exit
-        exit(42)
+        raise ex
```
MHajoha: I feel somewhat that this except block is a bit useless. You're printing the error and then printing the exception only to re-raise the exception, which will see it printed out anyhoo.

nathanctech: I think this can be useful for logging errors to either a database or file. Mecha will be run unattended so having it dump to a file will be valuable, but we still want to re-throw the error so mecha stops.

config/config.template.json
theunkn0wn1: lol ok. Methinks i used "presence" as that is what pydle called it. Will adjust

config.py
theunkn0wn1: Uhh... that was supposed to be at the top of file. So yes, stray. Will correct.

main.py
theunkn0wn1: Yeah, this method ensures the exception gets logged to disk. Since it is still a fatal error we want to re-raise it afterwards to ensure mecha crashes as fatal errors are not recoverable.

config.py
theunkn0wn1: Logging needs to be setup after the log file exists, I felt it was appropriate to call it after loading the configuration from disk

main.py
MHajoha: Wouldn't have to pass argv explicitly, can just access sys.argv from config.py

theunkn0wn1:
Many of the things mecha currently has as a configuration option would require a reboot as it were, the irc credentials for example. The API configuration (once implemented) may require a board reindex, and will disrupt caseload. 

As it stands, i fail to see how we can gracefully reload Mecha's configuration at runtime without significant service disruption that would warrant rebooting mecha anyways.

-----------
Ive migrated the CLI parsing to argparse, having it parse upon that modules **first import**.
Due to the way the Python Cache works, to my understanding, the module can be safely imported in multiple modules without Python re-executing the argument parsing. I made a similar change to the configuration manager

beepbeat:
My understanding agrees with your understanding, as every imported module is "cached" in system.modules, and every further import will be given the reference from there (probably not if you call magic directly, but that is frowned upon).

config.py
```diff
@@ -1,72 +1,77 @@
...
+    Sets up the module by loading the specified configuration file from disk
+
+    Args:
+        filename (str): path and filename to load.
     """
-    CASE_LIMIT = 30
-    """Default limit to board indexies, may be dynamically re-sized during runtime."""
+    global config
+
+    path = f"config/{filename}"
```
MHajoha: I believe the config path cli argument should not be relative to the config directory as it's not only somewhat against UNIX convention, will confuse users and breaks autocompletion in the console, it also defeats half the reason for there being a cli argument to set the config file: Being able to set any file in my file system as the config file.

Modules/cli_manager.py
```diff
@@ -0,0 +1,52 @@
...
+
+# register optional flag for verbose logging
+_parser.add_argument("-verbose", "-v", help="Enable verbose logging. "
+                                            "Earmuffs sold separately", action="store_true")
+
+
+# parse the *known* arguments into an object
+# parse_known_args returns a tuple (Namespace, unknown-args*)
+# we only care about the known arguments, thus [0]
+args = _parser.parse_known_args()[0]
```
MHajoha: This should be using `ArgumentParser.parse_args()`. `parse_known_args()` won't complain when rubbish cli arguments have been provided, so you'll sit there wondering why `./main.py --config-fiIe my_config.json` doesn't load `my_config.json` but runs fine, looking for a bug in `config.py` until you realise you've misspelled the cli argument but haven't been told.

PR #30, opened 2018-06-08
Assorted fixes and cleanup

MHajoha:
This PR does some minor bits of cleanup and fixes the logging setup as follows:
* Hardcode `sys.argv` in `conftest.py`.
* Don't repeat the logging setup in `logging_fx` as that breaks things. Clear the log capture instead.

### DIFF
--- a/Modules/cli_manager.py
+++ b/Modules/cli_manager.py
@@ -21,22 +21,15 @@ import logging
 # create a new parser
 _parser = argparse.ArgumentParser()
 
-# add arguments
-# - I would wrap this in a list of tuples to be looped over, but each argument may is too specific
-# - to enable looping. So we will register them manually.
-
 # register optional flag for --clean-log, truncating existing log file and starting a new blank one.
 _parser.add_argument("--clean-log", help="Start with a clean log file.  "
                                          "Deletes any existing log data.",
                      action="store_true")
+
 # register optional argument for the config file
 _parser.add_argument("--config-file", "--config", "-c", help="Specify the configuration file to load, "
                                                        "relative to config/",
                      default="configuration.json")
-# i had to set the above default to testing as pytest refuses to accept CLI args it doesn't
-#   expect and there is (apparently) no better solution other than to not give pytest
-#   a argument it doesn't expect >.>
-
 
 # register optional flag for verbose logging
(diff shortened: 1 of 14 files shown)


theunkn0wn1:
The logging system was not entirely finished, and @shatteredbeam was in the process of correcting many of the points you address in this PR.

I was under the impression you would just be fixing the unix compliance issue and the sys,arvg thing for testing. 

If possible, please remove the logging-specific fixes so Shatt's work on the issue is not ill-spent

MHajoha:
Shatt and I have been able to come to agree here and I've removed my changes to `setup_logging`, as I did admittedly get up in his business there, while keeping those to the logging test fixtures and logging tests themselves, so that our tests still run successfully.

shatteredbeam:
MHajoa and I ave indeed come to an agreement. the test improvements are OK with me to remain.

theunkn0wn1:
it appears all the merging I did today caused merge conflicts, sorry 'bout that :(

Modules/cli_manager.py
```diff
@@ -16,40 +16,29 @@
...
-_parser.add_argument("--clean-log", help="Start with a clean log file.  "
-                                         "Deletes any existing log data.",
+_parser.add_argument("--clean-log",
+                     help="Start with a clean log file. Deletes any existing log data.",
                      action="store_true")
+
 # register optional argument for the config file
-_parser.add_argument("--config-file", "--config", help="Specify the configuration file to load, "
-                                                       "relative to config/",
+_parser.add_argument("--config-file", "--config", "-c",
```
theunkn0wn1: i don't agree with `-c` being a valid shorthand for this cli argument. imho single char arguments should be flags.

MHajoha: In terms of convention, there shouldn't really be a difference between regular options and 'long' options `--` in what they are used for. [The POSIX specification added options (`-o`), while the GNU specification added long options (`--option`) as a more easily understandable version of the same thing.](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html)
Also, only one out of three of the flags we have has a short option, so we aren't really on that route anyways.
Interestingly, GNU wants long options with arguments to be referenced by `--option=arg`, but I don't think argparse supports this.

theunkn0wn1: grumbles

PR #31, opened 2018-06-08
Improvement/spark 20 | made rat_command module level

theunkn0wn1:
This PR makes one itsy, bitsy, but ever so important change: made rat_command module level.
This gets rid of the `rat_command.Commands` class and moves all its class methods and attributes to the module scope. 

this change simplifies command registration from
 ```py
@Commands.command(*alias)
async def foo(Context):
   pass  # etc
```

to 
```py
@command(*alias)
async def bar(Context):
   pass # etc
```


Modules/rat_command.py
```diff
@@ -55,141 +53,166 @@ class NameCollisionException(CommandException):
...
+        raise CommandException(f"Bot client has not been created"
+                               f" or not handed to Commands.")
+
+    # check for trigger
+    assert message.startswith(prefix), f"message passed that did not contain prefix."
+
+    log.debug(f"Trigger! {message}")
+
+    # remove command prefix and make lowercase
+    raw_command: str = message.lstrip(prefix).lower()
```
nathanctech: My broken code is here still...was this fix actually pushed to develop?

Modules/rat_command.py
```diff
@@ -55,143 +53,168 @@ class NameCollisionException(CommandException):
...
+
+    else:
+        for alias in names:
+            if alias in _registered_commands:
+                # command already registered
+                raise NameCollisionException(f"attempted to re-register command(s) {alias}")
             else:
-                raise CommandNotFoundException(f"Unable to find command {words[0]}")
+                formed_dict = {alias: func}
+                _registered_commands.update(formed_dict)
```
MHajoha: You know while we're here, this could be simplified to `_registered_commands[alias] = func`.

PR #33, opened 2018-06-10
SPARK-26: Epic object

theunkn0wn1:
implements a Epic data structure for storing Epic data from the API.

Because `Rescue` and `Rats` are now potentially Nullable, ive made the appropriate changes to their `__eq__` methods and added corresponding tests. 🤢 
- [x] class Epic
- [x] Epic tests
- [x] Rescue adaption
- [x] Rescue tests
- [x] Rats tests
- [x] revised Rescue tests
- [x] Documentation


theunkn0wn1:
Turns out this was blocked by SPARK-11, so I implemented those while i was here.

Modules/rats.py
```diff
@@ -71,14 +74,30 @@ def __eq__(self, other: 'Rats')->bool:
...
             bool: equal if uuid, platform, and name match
+
+        Raises:
+            TypeError : invalid type for equality check.
         """
+        if other is None:
+            return False
+
+        if not isinstance(other, Rats):
+            raise TypeError
```
MHajoha: Here's a thing I've recently come across: As part of the `__eq__` contract in Python, [it should return `NotImplemented`](https://docs.python.org/3/reference/datamodel.html#object.__eq__) if the class doesn't know how to compare itself to the other object. Python can then try using the `__eq__` method of the other object or return `False` if neither object knows how to compare itself to the other one.

Modules/rat_rescue.py
```diff
@@ -146,6 +152,29 @@ def __eq__(self, other) -> bool:
...
+            self.active,
+            self.code_red,
+            self.outcome,
+            self.title,
+            self.first_limpet,
+            self.lang_id,
+            self.irc_nickname,
+        )
+        if self._hash is None:
+            self._hash = reduce(xor, map(hash, attributes))
```
MHajoha: By moving the tuple creation into the `if` block, you could save it from being executed when the hash is already available.

Modules/rat_rescue.py
```diff
@@ -577,12 +606,12 @@ def open(self, value: bool) -> None:
             raise TypeError(f"expected type bool, got {type(value)}")
 
     @property
-    def epic(self) -> bool:
+    def epic(self) -> Epic:
```
MHajoha: The `epic` property should be a list (or set or tuple) of `Epic` objects, I believe a rescue can have multiple epics.

Modules/epic.py
```diff
@@ -0,0 +1,111 @@
...
+            rats
+        """
+        return self._rat
+
+    def __eq__(self, other: 'Epic') -> bool:
+        # null check
+        # while it is bad style to check a value against None, its possible this object
+        # will be None inside the rescue object, for rescues that are not Epics.
+        if other is None:
+            return False
```
MHajoha: It is not at all bad style to check a value against `None`, only using `==` is frowned upon. Also, if you choose to go with returning `NotImplemented` if the instance check fails, you won't need this extra check anyways as the comparison will evaluate to `False` regardless.

Modules/rat_rescue.py
theunkn0wn1: ah right, i overlooked that >.>

Modules/epic.py
```diff
@@ -91,7 +91,7 @@ def __eq__(self, other: 'Epic') -> bool:
         # while it is bad style to check a value against None, its possible this object
         # will be None inside the rescue object, for rescues that are not Epics.
         if other is None:
-            return False
+            return NotImplemented
 
         # type check
         if not isinstance(other, Epic):
```
MHajoha: If I read the contract correctly, here is where you'd return `NotImplemented` rather than if other is `None`.

theunkn0wn1: ... here i was thinking i fixed all that. yes, the None check is redundant as it would be caught by the `isinstance` check, which would also need to return `NotImplemented` rather than a raised type error.

Guess i missed  a spot

MHajoha:
Oh, quickly, before merging this, could you add SPARK-11 to the PR title in the hopes that JIRA figures it out and links them? Thanks.

PR #34, opened 2018-06-11
SPARK-27/add utf-8 support to file handler

shatteredbeam:
Fixes handling of extended characters in unicode/UTF-8, such as cyrillic.  Eliminates logging errors.


shatteredbeam:
This is a solution to bug/SPARK-27.

PR #35, opened 2018-06-11
SPARK-17 Enforce channel requirements for Dispatch Commands (part I)

theunkn0wn1:
This PR implements the `@permissions.require_channel()` decorator, which strictly enforces any command decorated with it to be invoked from a channel context. Any decorated command invoked from a PM will be returned with `This command must be invoked in a channel.` as an error message and the invoked command is canceled.

`@require_channel`
-----------------------------

This PR implements the technical side of SPARK-17, with the implementation portion being as simple as decorating commands with `@require_permission(message=str)`

`message` is optional, and defaults to `This command must be invoked in a channel.`

`@require_dm`
-----------------------------

This PR has a bonus feature! `

Basically does the opposite of `@require_channel` 
- [x] decorator
- [x] tests
- [x] documentation


Modules/permissions.py
```diff
@@ -128,4 +127,67 @@ def real_decorator(func):
                                     else permission.denied_message)
 
         return guarded
+
+    return real_decorator
+
+
+def require_channel(message: str = "This command must be invoked in a channel."):
```
MHajoha: It might prove a friendly piece of user interface for the failure message to include the invoked command's name in case multiple commands are invoked in quick succession?
I would also drop the parameter completely and make the decorator direct, but that is debatable.

commands/__init__.py
```diff
@@ -8,5 +8,6 @@
 
 See LICENSE.md
 """
-from . import *
 __all__ = ["debug"]
+
+from commands import *
```
MHajoha: `import .debug` might make the point of this clearer. Also, by going with that pattern, we can in the future choose which commands to import with ifs.

theunkn0wn1: I think the cleanest approach is to do a wildcard import of all modules in commands, we can disable the loading of specific commands via logic. eg i can check for the presence of the debug mode flag before loading the actual debug commands. This would only slightly reduce startup times, but not affect runtime speeds.

As for Clarity, i think this is clear enough. it imports every module from commands into the current namespace, as defined by the `__all__` entry

Modules/permissions.py
theunkn0wn1: as with the permission check decorator, i am leaving the option open for custom error messages. as for the command name? that should be pullable from the Context object. il look into this

theunkn0wn1: Done.

PR #37, opened 2018-06-12
SPARK-30: MarkForDeletion object

theunkn0wn1:
This PR reimplements `Rescue.mark_for_deletion`, giving it its own data structure `MarkForDelete`.

Most of our complex data structures now exist as Class objects, such as `User`, `Rats`, and `Quotation`.
Therefore, i reason, it is about time we do the same for MarkForDeletion.

Doing it this way better separates the API from Mecha's runtime, as the API would be able to change the background structure of the object without requiring significant change to Mechasqueak itself, but rather its API driver. 

This change also contributes to documentation and ease-of development as many IDEs find it easier to provide completions for Class objects than for dictionary entries.

Introduced class:
----------------
`MarkForDelete`

## instance attributes:
`marked`: bool, indicates if the rescue is marked for deletion
`reporter`: str, irc nickname of MD'ing user
`reason`: str reason for marking rescue as MD

Introduced helper methods in `Rescue`
----------------------
`Rescue.mark_delete` marks a case for deletion, requires author and a reason.
`Rescue.unmark_delete` unmarks a case as MD'ed.

Feedback on method naming would be appreciated

Checklist:
- [x] MarkForDeletion
- [x] Rescue refactoring
- [x] Rescue helper methods
- [x] tests
- [x] documentation

### DIFF
--- a/Modules/rat_rescue.py
+++ b/Modules/rat_rescue.py
@@ -19,6 +19,7 @@ from typing import Union, Optional, List
 from uuid import UUID
 
 from Modules.epic import Epic
+from Modules.mark_for_deletion import MarkForDeletion
 from Modules.rat_quotation import Quotation
 from Modules.rats import Rats
 from utils.ratlib import Platforms, Status
@@ -43,11 +44,11 @@ class Rescue(object):
                  quotes: list = None,
                  epic: List[Epic] = None,
                  title: Optional[str] = None,
-                 first_limpet: Optional[UUID]= None,
+                 first_limpet: Optional[UUID] = None,
                  board_index: Optional[int] = None,
-                 mark_for_deletion: Optional[dict]= None,
+                 mark_for_deletion: MarkForDeletion = None,
                  lang_id: str = "EN",
-                 rats: list = None,
+                 rats: List[Rats] = None,
                  status: Status = Status.OPEN,
                  code_red=False):
         """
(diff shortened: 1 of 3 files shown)


Modules/rat_rescue.py
```diff
@@ -675,7 +672,7 @@ def title(self, value: str or None) -> None:
             raise TypeError(f"expected type None or str, got {type(value)}")
 
     @property
-    def mark_for_deletion(self) -> dict:
+    def mark_for_deletion(self) -> MarkForDeletion:
```
MHajoha: Could you rename this to `marked_for_deletion` in order to emphasize that it is not an action, to contrast it from `mark_deletion`?

Modules/rat_rescue.py
```diff
@@ -699,30 +696,13 @@ def mark_for_deletion(self, value) -> None:
...
-                    raise ValueError("Data validation failed. at least one key contains invalid"
-                                     "data.")
-            else:
-                log.debug(f"Data of value is: {value}")
-                raise ValueError("required fields missing and/or keys!")
+        if isinstance(value, MarkForDeletion):
+            self._mark_for_deletion = value
         else:
-            raise TypeError(f"expected type dict, got type {type(value)}")
+            raise TypeError
```
MHajoha: Could you give these some kind of message? A common convention is to simply use the name of the wrong type as a message, but anything will do.

Modules/rat_rescue.py
```diff
@@ -805,6 +785,38 @@ def rats(self, value):
...
+        Args:
+            reporter (str): person marking rescue as deleted
+            reason (str): reason for the rescue being marked as deleted.
+
+        Raises:
+            TypeError: invalid params
+        """
+        # type enforcement
+        if not isinstance(reporter, str) or not isinstance(reason, str):
+            raise TypeError
```
MHajoha: Same here.

Modules/rat_rescue.py
theunkn0wn1: sure thing.

PR #38, opened 2018-06-15
Feature/spark 38

nathanctech:
This PR implements a search() function on the Rat Board.

The board has three find_by_x functions, but this function will allow commands to work properly as it will accept a valid type (str, int, UUID) and run the proper search function.

This PR also implements tests for the function and refactors case_id to uuid for clarity. It also makes the uuid writable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .idea/
 .pytest_cache/
-config/configuration.json
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
@@ -108,6 +108,7 @@ config/config.json
 config/live.json
 config/drill.json
 config/develop.json
+config/configuration.json
 
 #certs
 *.pem
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .idea/
 .pytest_cache/
-
+config/configuration.json
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/Modules/rat_board.py
+++ b/Modules/rat_board.py
@@ -151,7 +151,7 @@ class RatBoard(object):
 
         return index
 
-    def find_by_index(self, index: int) -> Rescue or None:
+    def find_by_index(self, index: int) -> Optional[Rescue]:
         """
         Searches for and returns a Rescue at a given `index` position, should it exist
 
@@ -168,12 +168,12 @@ class RatBoard(object):
             found = self.rescues[index]
         except KeyError:
             # the key doesn't exist, therefore no rescue at that index.
-            pass
+            return None
         else:
             # we found something
             return found
 
-    def find_by_name(self, client: str) -> Union[Rescue, None]:
+    def find_by_name(self, client: str) -> Optional[Rescue]:
         """
         Searches for and returns a Rescue for a given client, should it exist.
 
@@ -202,6 +202,45 @@ class RatBoard(object):
             if rescue.case_id == guid:
                 return rescue
         return None
+
+    def search(self, case: Union[str, int, UUID]) -> Optional[Rescue]:
+        """
+        Helper for searching for a case using implemented search functions.
+
+        Args:
+            case (Union[str, int, UUID]): the client's name, a case ID, or a rescue UUID
+
+        Returns:
+            Rescue: found rescue
+            None:   No rescue found
+
+        Raises:
+            ValueError: argument was not an expected type
+        """
+        # ensure a proper type was passed
+        if not isinstance(case, str) and not isinstance(case, int) and not isinstance(case, UUID):
+            raise ValueError
+
+        if isinstance(case, str):
+            # strings could be any of the valid types (such as using commands), check for the best one.
+            if case.isdigit():
+                    return self.find_by_index(int(case))
+            else:
+                # UUIDs always start with @ and are 37 characters, so try to parse it
+                if case[0] == "@" and len(case) == 37:
+                    try:
+                        guid = UUID(case[1:])
+                    except (AttributeError, ValueError):
+                        # not a valid UUID
+                        return None
+                    else:
+                        return self.find_by_uuid(guid)
+                else:
+                    return self.find_by_name(case)
+        elif isinstance(case, int) and not isinstance(case, bool):
+            return self.find_by_index(case)
+        else:
+            return self.find_by_uuid(case)
 
     def append(self, rescue: Rescue, overwrite: bool = False) -> None:
         """
--- a/Modules/rat_board.py
+++ b/Modules/rat_board.py
@@ -222,7 +222,7 @@ class RatBoard(object):
             raise ValueError
 
         if isinstance(case, str):
-            # strings could be any of the valid types (such as using commands), check for the best one.
+            # strings could be any of the valid types, check for the best one.
             if case.isdigit():
                     return self.find_by_index(int(case))
             else:
--- a/Modules/rat_board.py
+++ b/Modules/rat_board.py
@@ -219,10 +219,11 @@ class RatBoard(object):
         """
         # ensure a proper type was passed
         if not isinstance(case, str):
-            raise ValueError(f"Expected str, got {type(case)} instead")
+            raise ValueError(f"Expected str, got {type(case)}")
         # strings could be any of the valid types, check for the best one.
         if case.isdigit():
-            return self.find_by_index(int(case))
+            case_number = int(case)
+            return self.find_by_index(case_number)
         # UUIDs always start with @ and are 37 characters, so try to parse it
         elif case[0] == "@" and len(case) == 37:
             guid_str = case[1:]
--- a/Modules/rat_board.py
+++ b/Modules/rat_board.py
@@ -219,7 +219,7 @@ class RatBoard(object):
         """
         # ensure a proper type was passed
         if not isinstance(case, str):
-            raise ValueError(f"Expected str, got {type(case)}")
+            raise TypeError(f"Expected str, got {type(case)}")
         # strings could be any of the valid types, check for the best one.
         if case.isdigit():
             case_number = int(case)
--- a/Modules/rat_board.py
+++ b/Modules/rat_board.py
@@ -203,12 +203,12 @@ class RatBoard(object):
                 return rescue
         return None
 
-    def search(self, case: Union[str, int, UUID]) -> Optional[Rescue]:
+    def search(self, case: str) -> Optional[Rescue]:
         """
         Helper for searching for a case using implemented search functions.
 
         Args:
-            case (Union[str, int, UUID]): the client's name, a case ID, or a rescue UUID
+            case str: The case string to search on
 
         Returns:
             Rescue: found rescue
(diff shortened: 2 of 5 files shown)


tests/test_rat_board.py
```diff
@@ -280,8 +280,9 @@ def test_search_valid(self, test_input, rat_board_fx: RatBoard, rescue_plain_fx:
                             True,
                             3.14,
                             '@12345678-9876-53d1-ea5e-0000dead',
-                            '@12345678-9876-53d1-ea5e-000deadsheep'
-                            "42",
+                            '@12345678-9876-53d1-ea5e-0000deadbee-',
+                            '@12345678-9876-53d1-ea5e-000deadsheep',
```
theunkn0wn1: this is not a valid uuid as `s` is not within the range of hexidecimal.

Modules/rat_board.py
```diff
@@ -199,10 +199,50 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
...
+                    return self.find_by_index(int(case))
+            else:
+                # UUIDs always start with @ and are 37 characters, so try to parse it
+                if case[0] == "@" and len(case) == 37:
+                    try:
+                        guid_str = case[1:]
+                        guid = UUID(guid_str)
+                    except (AttributeError, ValueError):
+                        # not a valid UUID
+                        return None
```
theunkn0wn1: methinks you are writing too defensively. the only time this function should return None is if a rescue was not found. an invalidly formatted uuid should be raised.

Modules/rat_board.py
```diff
@@ -199,10 +199,50 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
...
+                    else:
+                        return self.find_by_uuid(guid)
+                else:
+                    return self.find_by_name(case)
+        elif isinstance(case, int) and not isinstance(case, bool):
```
theunkn0wn1: remind me, why are we considering booleans here?

Modules/rat_board.py
```diff
@@ -199,10 +199,50 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
...
+            return self.find_by_index(case)
+        else:
+            return self.find_by_uuid(case)
```
theunkn0wn1: multiple calls to `find_by_uuid`, consider refactoring

Modules/rat_rescue.py
```diff
@@ -336,18 +336,31 @@ def board_index(self, value: int or None) -> None:
...
-            This field is **READ ONLY**
         """
 
         return self._id
 
+    @uuid.setter
+    def uuid(self, uuid) -> None:
+        """
+        Sets the API uuid associated with the Rescue
+        Args:
```
theunkn0wn1: Newline before `Args:`, otherwise the docstring doesn't render correctly.

tests/test_rat_board.py
nathanctech: This ensures the exception is hit as it's the proper length (37 characters) but not a valid UUID. Mostly for user error since this function will take user input. Otherwise I didn't get 100% coverage ;)

Modules/rat_board.py
nathanctech: This function is considered "untrusted" because it will be taking user input, so it will need to deal with all sorts of issues with input. If you want it to raise an exception and make the caller deal with it, I'm on board with that.

Modules/rat_board.py
nathanctech: Python assumes True/False are also ints, so the isinstance(case, int) will return True. Perhaps this check is better served in the first assert statement.

Modules/rat_board.py
```diff
@@ -199,10 +199,50 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
 
         """
         for rescue in self.rescues.values():
-            if rescue.case_id == guid:
+            if rescue.uuid == guid:
                 return rescue
         return None
 
+    def search(self, case: Union[str, int, UUID]) -> Optional[Rescue]:
```
MHajoha: I believe this signature to be too broadly defined. This makes the very defensive programming, which unknown pointed out, necessary. Having to type check your arguments a lot is often an indicator of a badly designed function / method signature, even in Python.

For getting rescues directly by an int (board index) or by a UUID we already have the appropriate methods. This method's purpose, therefore, should not overlap with that of the others.

More specifically, this method should, in my mind, only take a string to then figure out what it represents, convert it and pass it to the appropriate method.

nathanctech: That makes sense to me. The purpose of the function is to allow the commands to "search" for the appropriate rescue and it'll always pass a string anyway. I'll refactor.

Modules/rat_board.py
theunkn0wn1: indeed this function is accepting untrusted user input, but its not responsible to what happens when the data it is given is invalid. I feel the correct course of action is to leave the raised exception to be handled by the caller - the command context.

 Since this method exists outside the command context there is no easy way beyond exceptions to express the fact something went wrong back to the user. Therefore silently handling the error here (defensive programming) makes it far more difficult to implement proper error handling

Modules/rat_board.py
theunkn0wn1: True/false are singletones in Python, it may be better to assert if the value /is/ a boolean rather than an instance thereof.

.gitignore
```diff
@@ -1,6 +1,6 @@
 .idea/
 .pytest_cache/
-
+config/configuration.json
```
theunkn0wn1: this belongs down near line 106, in the section provided for configuration files. Minor nitpick

Modules/rat_board.py
```diff
@@ -199,10 +199,43 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
...
+        if case.isdigit():
+            return self.find_by_index(int(case))
+        # UUIDs always start with @ and are 37 characters, so try to parse it
+        elif case[0] == "@" and len(case) == 37:
+            try:
+                guid_str = case[1:]
+                guid = UUID(guid_str)
+            except (AttributeError, ValueError):
+                # not a valid UUID
+                raise ValueError("Specified uuid is not valid.")
```
theunkn0wn1: catch/re-raise only muddles the stacktrace.  which makes debugging more difficult. i suggest not catching it at all

Modules/rat_board.py
```diff
@@ -199,10 +199,38 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
...
+        Returns:
+            Rescue: found rescue
+            None:   No rescue found
+
+        Raises:
+            ValueError: argument was not an expected type
+        """
+        # ensure a proper type was passed
+        if not isinstance(case, str):
+            raise ValueError(f"Expected str, got {type(case)} instead")
```
theunkn0wn1: Minor nitpick, remove `instead` if you could please; its not needed 😉

Modules/rat_board.py
```diff
@@ -199,10 +199,38 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
...
+        # strings could be any of the valid types, check for the best one.
+        if case.isdigit():
+            return self.find_by_index(int(case))
```
theunkn0wn1: two actions on one line.

Modules/rat_rescue.py
```diff
@@ -336,22 +336,37 @@ def board_index(self, value: int or None) -> None:
...
         Returns: API id
 
-        Notes:
-            This field is **READ ONLY**
         """
 
         return self._id
 
+    @uuid.setter
+    def uuid(self, uuid) -> None:
```
theunkn0wn1: Setter convention dictates the signature is
 ```py
def uuid(self, value)
```
Further, it may prove useful to put a type hint in here, eg
```py
def uuid(self, value:UUID)
```

tests/test_rat_board.py
```diff
@@ -262,3 +262,29 @@ def test_next_free_mixed_board(self, keys: list, expected: int, rat_board_fx: Ra
...
+        assert rat_board_fx.search(test_input) == test_rescue
+
+    @pytest.mark.parametrize("test_input", [
+                            "unit_tes",
+                            '@12345678-9876-53d1-ea5e-0000dead',
+                            '@12345678-9876-53d1-ea5e-0000deadbee-',
+                            '@12345678-9876-53d1-ea5e-000deadsheep',
+                            42,
+                            "42.5"])
+    def test_search_garbage(self, test_input, rat_board_fx: RatBoard, rescue_plain_fx: Rescue):
```
theunkn0wn1: why are you making a shalow copy of the rescue fixture? it is unnecessary and doesn't really do anything.
(leftover from `deepcopy`?)

tests/test_rat_board.py
```diff
@@ -262,3 +262,29 @@ def test_next_free_mixed_board(self, keys: list, expected: int, rat_board_fx: Ra
...
+        test_rescue = rescue_plain_fx
+        test_rescue.uuid = UUID('12345678-9876-53d1-ea5e-0000deadbeef')
+        rat_board_fx.append(test_rescue)
+        try:
+            assert rat_board_fx.search(test_input) is None
+        except ValueError:
```
theunkn0wn1:
```py
with pytest.raises(ValueError):
    assert rat_board_fx.search(test_input) is None

File: tests/test_rat_board.py (hunk 12 above)
Comment (nathanctech) [reply tests/test_rat_board.py]: It gives me an "unresolved reference" error if I don't include it.

File: tests/test_rat_board.py (hunk 13 above)
Comment (nathanctech) [reply tests/test_rat_board.py]: This will require search() to raise an exception any time it can't find a case. Is this what we want?

Comment (theunkn0wn1) [reply tests/test_rat_board.py]: no? it would appear this test is asserting an exception was raised. if that is not the case remove the second unit.

Tests should only assert one thing.

File: Modules/rat_board.py [hunk 14]
```diff
@@ -199,10 +199,39 @@ def find_by_uuid(self, guid: UUID) -> Optional[Rescue]:
...
+        Returns:
+            Rescue: found rescue
+            None:   No rescue found
+
+        Raises:
+            ValueError: argument was not an expected type
+        """
+        # ensure a proper type was passed
+        if not isinstance(case, str):
+            raise ValueError(f"Expected str, got {type(case)}")
```
theunkn0wn1: This should be a `TypeError`, as its the type that is the problem.

`ValueError` is for when the value is the correct type, but isn;t within the acceptable value range

tests/test_rat_board.py
```diff
@@ -262,3 +262,39 @@ def test_next_free_mixed_board(self, keys: list, expected: int, rat_board_fx: Ra
...
+                            False,
+                            42,
+                            uuid4(),
+                            None])
+    def test_search_garbage_raises_exception(self, test_input, rat_board_fx: RatBoard,
+                                             rescue_plain_fx: Rescue):
+        rescue_plain_fx.uuid = UUID('12345678-9876-53d1-ea5e-0000deadbeef')
+        rat_board_fx.append(rescue_plain_fx)
+        with pytest.raises((TypeError, ValueError,)):
+            assert rat_board_fx.search(test_input) is None
```
theunkn0wn1: Couldn't you just call .search directly, and not assert its output? as it doesn't output because its expected to raise?

PR #39, opened 2018-06-16
Spark-29/Remove unit testing from test_rescue.py

shatteredbeam:
Unit testing removed from test_rescue.py + misc small changes.

### DIFF
--- a/tests/test_rescue.py
+++ b/tests/test_rescue.py
@@ -14,6 +14,7 @@ import pytest
 from copy import deepcopy
 from datetime import datetime
 from uuid import uuid4, UUID
+from Modules.epic import Epic
 from Modules.mark_for_deletion import MarkForDeletion
 from Modules.rats import Rats
 from Modules.rat_rescue import Rescue
@@ -195,14 +196,17 @@ def test_rescue_quotes(rescue_sop_fx):
 
 def test_epic_rescue_attached(epic_fx):
     """
-    Verifies Epic obj data attached to rescue is returned properly.
+    Verifies attached epic object referenced by rescue is the same object, and
+    that only one rescue has been added.
     """
     # Create local rescue object
     test_rescue = Rescue(uuid4(), 'TestClient', 'Alioth', 'Test_Client', epic=[epic_fx])
 
-    for epic in test_rescue.epic:
-        assert epic.notes == 'my notes package'
-        assert str(epic.uuid) != ''
+    # One rescue added, the List should return only ONE Epic.
+    assert len(test_rescue.epic) == 1
+
+    # Check that the epic object in list is epic_fx
+    assert test_rescue.epic[0] is epic_fx
 
 
 @pytest.mark.parametrize('expected_title', ['Operation Unit Hazing', 'Dumbo Drop', 'Delight'])
--- a/tests/test_rescue.py
+++ b/tests/test_rescue.py
@@ -102,21 +102,14 @@ def test_uuid_is_set(rescue_sop_fx):
 
 def test_client_is_set(rescue_sop_fx):
     """
-    Verfies that rescue_sop_fx._client is set.
+    Verifies that rescue_sop_fx._client is set.
     """
     assert rescue_sop_fx.client != ''
 
 
-def test_client_attribute_exists(rescue_sop_fx):
-    """
-    Verifies client attribute exists
-    """
-    assert hasattr(rescue_sop_fx, 'client')
-
-
 def test_created_at_date_exists(rescue_sop_fx):
     """
-    Verfies rescue.created_at datetime is set, and in the past.
+    Verifies rescue.created_at datetime is set, and in the past.
     """
     expected_time_differential = (datetime.utcnow() - rescue_sop_fx.created_at)
     assert expected_time_differential != 0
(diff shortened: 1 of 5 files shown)


shatteredbeam:
There's one section that didnt get tested as of yet - having to do with Rats being added/removed from the rat cache.  unknown will likely need to provide coverage for that area.

tests/test_rescue.py
```diff
@@ -10,569 +10,419 @@
...
-        garbage = ["pc", None, 42, 12.2, []]
-        for piece in garbage:
-            with self.subTest(piece=piece):
-                with self.assertRaises(TypeError):
-                    self.rescue.platform = piece
+from Modules.rat_rescue import Rescue
+from utils.ratlib import Status
 
 
 class TestRescuePyTests(object):
```
theunkn0wn1: if you are going to move the tests out of the class, you may as well remove the class entirely

shatteredbeam: I'll get .open, and .title covered, as well as fix the indents.  Expect another commit to this PR on Saturday.

tests/test_rescue.py
```diff
@@ -10,569 +10,447 @@
...
             reason (str): md reason
             reporter(str) md reporter
-        """
-        with pytest.raises(TypeError):
-            rescue_sop_fx.marked_for_deletion.reason = reason
+    """
+    with pytest.raises(TypeError):
+        rescue_sop_fx.marked_for_deletion.reason = reason
 
         with pytest.raises(TypeError):
```
theunkn0wn1: inner `pytest.raises` unreachable.

tests/test_rescue.py
```diff
@@ -10,569 +10,447 @@
...
+@pytest.mark.asyncio
+async def test_add_rat_from_cache(rat_good_fx: Rats, rescue_sop_fx: Rescue):
+    """
+    Verifies rat added from cache matches rat object in Rescue
+    """
+    await rescue_sop_fx.add_rat(rat_good_fx.name)
+    assert rat_good_fx == rescue_sop_fx.rats[0]
+
+
+@pytest.mark.parametrize("garbage", [(None,), (42,), (-2.2,), (uuid4(),)])
```
theunkn0wn1: should this not just be a plain list of objects rather than a list of single-element tupples? this test otherwise only tests tupples of len 1

tests/test_rescue.py
```diff
@@ -584,85 +462,204 @@ def test_mark_for_deletion_setter_bad_data(self, reason: str or None, reporter:
...
+        rescue_sop_fx.mark_delete(None, "sna")
 
         with pytest.raises(TypeError):
             rescue_sop_fx.mark_delete("sna", None)
 
         with pytest.raises(ValueError):
             rescue_sop_fx.mark_delete("unit_test", "")
+            
+
+def test(rescue_sop_fx: Rescue):
```
theunkn0wn1: While this is technically a valid name and will be detected during test discovery, can we please rename this to describe what the test actually is? ((based on the git diff methinks i might be to blame for this test's name... 😭 ))

tests/test_rescue.py
shatteredbeam:
```py
def test_mark_for_deletion_setter_bad_data(reason: str or None, reporter: str or None,
                                           marked: bool, rescue_sop_fx: Rescue):
    """
    Verifies setting the mark for deletion property succeeds when the data is valid

        Args:
            rescue_sop_fx (): plain rescue fixture
            reason (str): md reason
            reporter(str) md reporter
    """
    with pytest.raises(TypeError):
        rescue_sop_fx.marked_for_deletion.reason = reason

    with pytest.raises(TypeError):
        rescue_sop_fx.marked_for_deletion.reporter = reporter

    with pytest.raises(TypeError):
        rescue_sop_fx.marked_for_deletion.marked = marked

    assert rescue_sop_fx.marked_for_deletion.marked is False
    assert rescue_sop_fx.marked_for_deletion.reason != reason
    assert rescue_sop_fx.marked_for_deletion.reporter != reporter
```

Verified these inner with statements are reachable by changing TypeError to ValueError, and verifying with failed test by wrong Exception with pytest.raised(), with the above format.  The unreachable inners seem to have been due to indent only.

tests/test_rescue.py
shatteredbeam:
```py
@pytest.mark.parametrize("garbage", [(None), (42), (-2.2), (uuid4())])
```
Changed to proper format for multiple tests.  Pycharm is whining about extraneous parentheses, but the syntax is valid.

tests/test_rescue.py
shatteredbeam: renamed :)

shatteredbeam:
Also, there is now officially no negative coverage from this PR.

tests/test_rescue.py
```diff
@@ -10,659 +10,656 @@
...
+def test_epic_rescue_attached(epic_fx):
+    """
+    Verifies Epic obj data attached to rescue is returned properly.
+    """
+    # Create local rescue object
+    test_rescue = Rescue(uuid4(), 'TestClient', 'Alioth', 'Test_Client', epic=[epic_fx])
+
+    for epic in test_rescue.epic:
+        assert epic.notes == 'my notes package'
+        assert str(epic.uuid) != ''
```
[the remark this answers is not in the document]
shatteredbeam: The thought process was that Rescue.Epic returns List[Epic], so to make the test friendly with multiple epic objects, I set up a loop.  I've made the test more specific, and checked for the same instance of the fixture instead.

PR #41, opened 2018-06-25
SPARK-39: MD `__eq__` and `__hash__` methods

theunkn0wn1:
this PR implements the missing `MarkForDeletion.__eq__` method.

As a supplement, i have also implemented a `__hash__` method for this object.

### DIFF
--- a/tests/test_mark_for_deletion.py
+++ b/tests/test_mark_for_deletion.py
@@ -1,0 +1,72 @@
+"""
+test_mark_for_deletion.py - tests for MarkForDeletion
+
+Copyright (c) 2018 The Fuel Rats Mischief,
+All rights reserved.
+
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+
+This module is built on top of the Pydle system.
+"""
+from typing import Optional
+from uuid import uuid4
+
+import pytest
+
+from Modules.mark_for_deletion import MarkForDeletion
+
+
+@pytest.mark.parametrize("reason,reporter,marked", [
+    ([], 42.2, -1),
+    (-2.1, {"Potato"}, None),
+    ([], 42, "md reason"),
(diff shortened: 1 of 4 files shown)


tests/test_mark_for_deletion.py
```diff
@@ -0,0 +1,81 @@
...
+@pytest.mark.parametrize("reason,reporter,marked", [
+    ([], 42.2, -1),
+    (-2.1, {"Potato"}, None),
+    ([], 42, "md reason"),
+    (True, -42.2, uuid4())
+])
+def test_mark_for_deletion_setter_bad_data(reason: str or None, reporter: str or None,
+                                           marked: bool, mark_for_deletion_fx: MarkForDeletion):
+    """
+    Verifies setting the mark for deletion property succeeds when the data is valid
```
MHajoha: Is there an `only` missing here?

tests/test_mark_for_deletion.py
```diff
@@ -0,0 +1,81 @@
...
+    marked_a, reason_a, reporter_a = data_a
+    md_a = MarkForDeletion(marked_a, reporter_a, reason_a)
+
+    marked_b, reason_b, reporter_b = data_b
+    md_b = MarkForDeletion(marked_b, reporter_b, reason_b)
+    assert md_a != md_b
+
+
+def test_hash(mark_for_deletion_fx):
+    assert hash(mark_for_deletion_fx) != 0
```
MHajoha: I feel like testing the hash might warrant one or two more, simple tests. Like creating another, different object and making sure the hashes are unequal and creating an equal object and making sure that the hashes are equal.

theunkn0wn1: I really drew a blank about testing the hash. but that sounds like a good plan. Il see to those tests in the morning

PR #42, opened 2018-06-25
Addendums to #15

beepbeat:


beepbeat:
This is now part of #51

PR #43, opened 2018-06-26
SPARK-13 The RatCache

theunkn0wn1:
Migrates the rat cache out of Rats and into its own module.

Further, `Rats` has been refactor/renamed to `Rat` as it is more befitting


Modules/rat.py
```diff
@@ -18,29 +18,23 @@
...
+    cache.
     """
-
-    cache_by_id = {}
-    """Cache of rat objects by their UUID"""
-    cache_by_name = {}
-    """Cache of rat objects by rat name (str)"""
-    api_handler = None
-    """API handler"""
+    cache = RatCache()
```
MHajoha: I believe it to be problematic to save the instance of a singleton somewhere else. For example, we might want to replace `RatCache._instance` for testing (or otherwise) purposes at some point. This change wouldn't affect the Rat object as one would expect, however, as it is using its own reference to the instance.
Calling `RatCache()` whereever the instance is needed would, in my opinion, be a safer approach, even if it does add a few calls.

beepbeat: I think this is one of the instances where the Singleton class I use for my DBM would be applicable as it does exactly what has been proposed.

MHajoha: I hadn't realised you had changed that from the `get_instance` approach. While I do like the general idea of your new approach (overriding `__call__` in a metaclass), your implementation seems to, well, be horribly broken from what I can see, and wouldn't provide a benefit over the `__new__` approach here either way.

To clarify, this review comment is about storing the instance of the singleton in a second place in addition to the singleton itself. No approach in the implementation of the singleton can effectively prevent that.

theunkn0wn1: I have removed the static reference to `RatCache` in `Rat`, the additional function calls should not introduce significant overhead as @MHajoha pointed out.

theunkn0wn1:
(branch rebased against develop, to fix rebase merge conflict)

PR #46, opened 2018-06-29
SPARK-21: Configurable Permissions

theunkn0wn1:
This PR implements configurable permission vhosts as described in SPARK-21.

### DIFF
--- a/Modules/permissions.py
+++ b/Modules/permissions.py
@@ -13,7 +13,7 @@ This module is built on top of the Pydle system.
 """
 import logging
 from functools import wraps
-from typing import List
+from typing import List, Dict, Set
 
 from Modules.context import Context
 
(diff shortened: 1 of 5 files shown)


Modules/permissions.py
```diff
@@ -24,23 +26,140 @@ class Permission:
...
+        # determine which items are new
+        added = value - self.vhosts
+
+        # recompute _by_vhost to exclude removed vhosts we have removed
+        # ( as there is no bulk-remove method, this avoids a for loop and some function calls )
+        all_vhosts = {vhost: permission for vhost, permission in _by_vhost.items() if
+                      vhost not in removed}
+
+        # create a dict of new vhosts to register
+        new_vhosts = {vhost: self for vhost in added}
```
MHajoha: It will probably be safer, simpler and faster to just build up `_by_vhost` from scratch.

theunkn0wn1: `_by_vhost` is being rebuilt from scratch. 
The line you cite there is where im building a dict of all new permissions being registered, which gets added to `all_vhosts` which is `_by_vhost` stripped of all vhosts that are being deregistered.

`all_vhosts` gets combined with `new_vhosts` which is then pushed to `_by_vhost`

MHajoha: I'm only noticing your response now, sorry. Looks like I didn't get an email.
You're right, I misread part of the code, it is necessary to rely on the old `_by_vhost`, although it could be simplified with something along the lines of:
```python
global _by_vhost
_by_vhost = {vhost: permission for vhost, permission in _by_vhost.items() if permission is not self or vhost}  # all vhosts that aren't ours
_by_vhost.update((vhost, self) for vhost in value)  # ours
```
But that's just optional isn't it.

theunkn0wn1: _by_vhost is a dict, and you appear to be suggesting a tupple comprehension? 

at any rate i think it is desirable to keep the dict comprehension as its own step.

MHajoha: `dict.update` has an overload `(Iterable[tuple])`, but as I said your version is fine

PR #47, opened 2018-06-29
Spark 45 / Documentation Update

shatteredbeam:
Update Documentation:

* README.MD
* CONTRIBUTING.MD
* TESTING.MD (New)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This script can be run as a Docker image or run locally.
 
 To run as a docker image, you may use the provided docker-compose.
 
-If you would prefer to run the service locally, bare in mind you may need to run a local PostgreSQL Database for the fact module.
+If you would prefer to run the service locally, bear in mind you may need to run a local PostgreSQL Database for the fact module.
 
 ### Building via Docker
 We have provided a `docker-compose.yml` suitably configured for building and running Mecha
@@ -65,5 +65,5 @@ By default, Mecha will attempt to load `configuration.json` (file not provided) 
 To configure Mecha locally, please copy the provided `config.template.json` to a new file.
 `configuration.json` and fill in the appropriate fields, see `config/configuration.md` for details.
 
-Please note that, when run via docker, it is not necessary to rebuild the image after changing
+When run via docker, it is not necessary to rebuild the image after changing
 configuration options as that directory is shared with the container (see `config/configuration.md`)
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ docker-compose run mechasqueak
 > If you are not using pip, please see requirements.txt for the list of required libraries.  These will need to be installed manually if pip is not available.
 
 1. Clone the repository from the ``master`` branch, or for bleeding edge, use ``develop``.  Please keep in mind, Develop changes frequently and may be broken.
-2. Copy ``config.template.json`` to a new file, such as ``configuration.json`` and edit your options.
+2. Build your configuration file.  Please see the [Configuration](#Configuration) section.
 3. Execute Mecha with ``python main.py --config <your configuration file>``  (You may need to use the python3 alias)
 
 |    Flag         |    Description                     |
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ docker-compose run mechasqueak
 ```
 
 ## Run Locally
-> If you are not using pip, please see requirements.txt for the list of required libraries.  These will need to be installed manually if pip is not available.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -39,13 +42,28 @@ command:
 ```bash
 docker-compose run mechasqueak
 ```
+
+## Run Locally
+> If you are not using pip, please see requirements.txt for the list of required libraries.  These will need to be installed manually if pip is not available.
```
theunkn0wn1:
```md
> If you are not using pip, please see requirements.txt for the list of required libraries.  You will need to install these manually.
```

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+    return myRat
+````
+
+The first returns a simple, uninitialized rat object, while the second one is parametrized to run any test its used in three times, initialized three times; once for each platform.
+
+Most importantly, these fixtures make module calls to their appropriate sub objects, and can be used without breaking scope in other unit tests, and are disposed so that each iteration of parameters uses a 'clean' object.
+
+Here's an example of a valid and complete test:
+```python
+@pytest.mark.asyncio
```
theunkn0wn1: indentation required

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+        with pytest.raises(TypeError):
+            await Rats.get_rat_by_name(name=garbage)
+
+        with pytest.raises(TypeError):
+            await Rats.get_rat_by_name(name="foo", platform=garbage)
+```
+
+# Marking Tests
+
+Tests are categorized by module using ``pytest.mark`` so that specific tests may be excluded.  Each module's decorator is set using the ``pytestmark`` global:
```
theunkn0wn1: or test classes decorated.

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+
+```python
+# Replace 'category' with an unused one for your module, after it has been added to pytest.ini
+pytestmark = pytest.mark.category
+```
+
+Test categories are listed in [pytest.ini](./pytest.ini).  It may be necessary to run tests locally without using the ``database`` or ``api`` markers if you do not have these available.  This can be done by running py.test in the following manner:
```
theunkn0wn1:
```md
...markers if you do not have these available...
```
perhaps replace that with "if those test requisites are not available in your dev environment"?

TESTING.md
```diff
@@ -0,0 +1,110 @@
...
+```
+pytest -m "not (database or api)"
+```
+
+To exclude a single category:
+```
+pytest -m "not database"
+```
+
```
theunkn0wn1: also include an example of running a specific module?
```bash
pytest -m permissions
```

shatteredbeam:
Requested changes applied.

TESTING.md
theunkn0wn1: ((for some reason its not marked outdated, comment resolved ))

TESTING.md
```diff
@@ -0,0 +1,116 @@
...
+def rat_good_fx(request) -> Rats:
+    """
+    Testing fixture containing good and registered rats
+    """
+    params = request.param
+    myRat = Rats(params[2], name=params[0], platform=params[1])
+    return myRat
+````
+
+The first returns a simple, uninitialized rat object, while the second one is parametrized to run any test its used in three times, initialized three times; once for each platform.
```
MHajoha: > uninitialized rat object

Isn't quite accurate. The object is initialized. I think your meaning here was that it does not have an ID set?

TESTING.md
```diff
@@ -0,0 +1,116 @@
...
+# Marking Tests
+
+Tests are categorized by module using ``pytest.mark`` so that specific tests may be excluded.  Each module's decorator is set using the ``pytestmark`` global:
+
+```python
+# Replace 'category' with an unused one for your module, after it has been added to pytest.ini
+pytestmark = pytest.mark.category
+```
+
+Test categories are listed in [pytest.ini](./pytest.ini).  It may be necessary to run tests locally without using the ``database`` or ``api``.  This can be done by running py.test in the following manner:
```
MHajoha: I can only speak for myself, but the API tests do not use the API or any other external resource. I'm still unsure about the necessity of an `api` marker, though I suppose they are rather slow.
Anyways, I'd be happier if you could just throw out the 'locally' here :wink:

shatteredbeam: Requested changes made.  I reworded the rat plain fixture to be simply 'without attributes' along with some other typos that were missed.

PR #48, opened 2018-06-29
Resubmitting the DBM and burning docker

beepbeat:


tests/test_database_manager.py
```diff
@@ -42,54 +42,53 @@ class TestStuff(object):
     manager = DatabaseManager()
```
theunkn0wn1: This will cause a collection failure when the Db is not available, completely breaking the test process.

Instead of creating the instance during **class load** this should be done in a fixture

postgres-files/Dockerfile
```diff
@@ -1,4 +1,4 @@
 # Use alpine image
 FROM postgres:alpine
-# Bind a directory to store PostGres data and databases
-ENV PGDATA /psql-data
+COPY init.sh /docker-entrypoint-initdb.d/init.sh
+RUN chmod +x /docker-entrypoint-initdb.d/init.sh
```
theunkn0wn1: missing newline at EOF.

postgres-files/init.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
+	CREATE USER mecha3 LOGIN;
+	CREATE DATABASE mecha3;
+	GRANT ALL PRIVILEGES ON DATABASE mecha3 TO mecha3;
+EOSQL
```
theunkn0wn1: \n at EOF

pytest.sh
```diff
@@ -1,6 +1,4 @@
 export P_HOST
-psql -c 'CREATE ROLE mecha3 LOGIN CREATEDB' -U postgres -h $P_HOST
```
theunkn0wn1: do we really need a shell script here?

tests/test_database_manager.py
```diff
@@ -42,54 +42,53 @@ class TestStuff(object):
     manager = DatabaseManager()
 
     async def test_has_table(self):
-        if await self.manager._has_table("testtablehas"):
-            await self.manager._drop_table("testtablehas")
-        assert not await self.manager._has_table("testtablehas")
-        await self.manager._create_table("testtablehas", {"string1": "VARCHAR"})
-        assert await self.manager._has_table("testtablehas")
+        if await self.manager.has_table("testtablehas"):
```
theunkn0wn1: Your tests should use a fixture rather than a class / instance attribute

Modules/database_manager.py
```diff
@@ -11,19 +11,12 @@
...
+from utils.ratlib import Singleton
 
-    def __call__(cls, *args, **kwargs):
-        if cls not in cls._instances:
-            cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
-        return cls._instances[cls]
+log = logging.getLogger(f"mecha.{__name__}")
 
 
 # noinspection SqlNoDataSourceInspection
```
theunkn0wn1: I must ask why do we need a metaclass?
to quote Tim Peters: ```Metaclasses are deeper magic than 99% of users should ever worry about. If you wonder whether you need them, you don’t (the people who actually need them know with certainty that they need them, and don’t need an explanation about why).```

beepbeat: Thanks for handing me quote I already quoted. Secondly: Show me a cleaner way to achieve that, while not having similar code floating around

tests/test_database_manager.py
beepbeat: you come up with that now?

pytest.sh
beepbeat: not anymore, testing moved outside docker

postgres-files/init.sh
beepbeat: fixed

postgres-files/Dockerfile
beepbeat: fixed

tests/test_database_manager.py
theunkn0wn1: i raised it twice before

PR #49, opened 2018-06-30
feature/SPARK-24: Gracecful error messages

theunkn0wn1:
This Pr implements a high-level error handler in MechaClient.on_message that catches any exceptions raised during command invocation.

The objective is to print a psudo error message to the invoking user, notifying them the command failed to execute
and providing a reference code. 

This reference code is the first octet of a uuid4, generated randomly when this error handler is raised.
The full uuid will be written into the logfile directly above the stacktrace of the actual exception.
This way, the user gets a friendly message, while providing maintainers sufficient information to debug the issue.


Modules/graceful_errors.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+import random
+from typing import Dict
+from uuid import UUID
+
+from Modules.rat_board import IndexNotFreeError, RescueBoardException, RescueNotChangedException, \
+    RescueNotFoundException
+from Modules.rat_command import InvalidCommandException, NameCollisionException
+
+BY_ERROR: Dict[type(Exception), str] = {
```
beepbeat: I use and have seen RuntimeError too, so  you might want to find something for that too

Modules/graceful_errors.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    base = "Oh noes! {message} {cheese} encountered! please contact a tech! Reference code {uuid}"
+
+    ex_type = type(ex)
+
+    # gets the first octet from the uuid
+    octet = str(ex_uuid)[:8]
+
+    message = BY_ERROR[ex_type] if ex_type in BY_ERROR else "unknown error"
```
beepbeat: Maybe have an adjective for "idk this stuff" too? ("unknown error gouda") doesn't sound tasty. "unbearable" perhaps?

main.py
```diff
@@ -91,9 +97,20 @@ def __init__(self, *args, **kwargs):
...
-                                      channel=channel)
+            try:
+                await rat_command.trigger(message=sanitized_message,
+                                          sender=user,
+                                          channel=channel)
+            except CommandNotFoundException:
+                # command was not found
+                log.debug("Command invoked from \"{message}\" not found.")
+
+            except Exception as ex:
```
beepbeat: I might be blind, but does the exception get logged at all? ("raise" or "log.exception(ex)" )

pytest.ini
```diff
@@ -15,4 +15,5 @@ markers =
     user: marks a test as a user object test
     logging: marks a test as being a test for logging
     ratlib: marks a test as a ratlib test
-    setup_tests: setup test marker
\ No newline at end of file
+    setup_tests: setup test marker
+    graceful_error: 'graceful_error' tests
```
beepbeat: Missing \n at EOF ;)

Modules/graceful_errors.py
theunkn0wn1: Sure thing

Modules/graceful_errors.py
theunkn0wn1: but thats not a cheese. sounds like fun 😛

main.py
theunkn0wn1: see line 110
```py
            except Exception as ex:
                ex_uuid = uuid4()
                log.exception(ex_uuid)  # Hi there BeepBeat!
                error_message = graceful_errors.make_graceful(ex, ex_uuid)
                # and report it to the user
                await self.message(channel, error_message)
```

PR #51, opened 2018-07-01
Database module and facts, take III / SPARK-57

beepbeat:


Modules/rat_facts.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self):
+        """
+        retrieves the DBM object. should this fail, it will set the variable to None,
+            failing all subsequent calls
+        """
+        try:
+            self.dbm = DatabaseManager()
+        except:
+            self.dbm = None
```
MHajoha: This is not at all a safe way to do this for two reasons:
* Bare except clauses are a bad idea in most circumstances and your IDE should tell you this. You should only catch those exception which you expect to be thrown and whish to handle.
* If there is something wrong and the DBM can't be initialized, this constructor will run through quietly as if all were fine, leaving the fire to be noticed only when someone attempts to use a fact. Errors should be noticed as soon as possible, please raise an exception here if at all possible.

tests/conftest.py
```diff
@@ -224,3 +226,15 @@ def reset_rat_cache_fx(rat_cache_fx: RatCache):
...
     rat_cache_fx.flush()
+
+
+@pytest.fixture(scope="session")
+def dbm_fx():
+    """returns a DBM instance"""
+    return DatabaseManager()
+
+
+@pytest.fixture(scope="session")
```
MHajoha: Is there any reason for these fixtures to have the scope of session? This makes tests interdependent, as, if one test breaks the fixture object, it will be broken for all following tests as well. This should be avoided if at all possible.

tests/test_database_manager.py
```diff
@@ -0,0 +1,91 @@
...
+        await dbm_fx.create_table("testtablehas", {"string1": "VARCHAR"})
+        assert await dbm_fx.has_table("testtablehas")
+
+    async def test_select_row(self, dbm_fx):
+        if not await dbm_fx.has_table("testtableselect"):
+            await dbm_fx.create_table("testtableselect", {"string1": "VARCHAR"})
+        await dbm_fx.insert_row("testtableselect", ("thest",))
+
+        tmp = await dbm_fx.select_rows("testtableselect", "AND", {"string1": "thest"})
+        print(tmp)
```
MHajoha: This should be removed before merging :wink:

tests/conftest.py
beepbeat: The scope isn't supposed to be set there...

Modules/rat_facts.py
beepbeat: This may be a leftover to prevent pytest from failing on collection.

Modules/database_manager.py
```diff
@@ -0,0 +1,249 @@
...
+# noinspection SqlNoDataSourceInspection
+class DatabaseManager(metaclass=Singleton):
+
+    def __init__(self):
+        """
+        Connects to the DB, sets up the connection, retrieves the cursor.
+        Creates the default tables should they not exist.
+        """
+        # connect to PostgreSQL (PSQL) database
+        __config: dict = config.get("database")
```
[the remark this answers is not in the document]
beepbeat: Imo ["stuff"] should always be reserved for collections, not complex classes/objects. This just gives you a clear distinction between what is what. 

Also this it no supposed to mangle the name. Using "config" as a name will cause name shadowing, which you should prevent if not completely necessary. I just chose __ as a prefix to denote the fact that this is a completely and utterly internal variable.

tests/test_database_manager.py
beepbeat: Good eyes... Stupid debug stuff without a FIXME

Modules/database_manager.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+
+        @wraps(func)
+        async def wrapper():
+            """
+            Returns:
+                whatever the called function returns (probably None)
+            """
+            if not enabled:
+                raise ValueError("Module is disabled!")
```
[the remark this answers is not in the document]
beepbeat: That's sounds way better, it's just me who fell in love with decorators (and metaclasses)...

Modules/database_manager.py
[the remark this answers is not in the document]
theunkn0wn1: using `get` is an unnecessary function call and does lead to confusion. usage of indexing is much clearer here.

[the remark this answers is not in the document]
beepbeat: It may be an unnecessary function call but it emphasizes that this is *not* a collection, which is especially important in not strongly-typed languages like python.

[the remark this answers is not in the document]
theunkn0wn1: To quote from the docstring provided for `dict.get`
```Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.```

The only difference between using the function and using the index is that the function will return None if the key does not exist. I do not see where this function in any way implies the type returned. Further, the `database` key *is* a collection of subkeys.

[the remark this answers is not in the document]
Comment (beepbeat) [reply Modules/database_manager.py]: My argumentation always referred to implications about the object .get is executed on, not the returned type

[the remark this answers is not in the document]
Comment (MHajoha) [reply Modules/database_manager.py]: @beepbeat, our config **is** just a dict, so by your own logic you would use indexing.

`get` returning None if nothing is found is also something we don't want. When something is on fire, we want to know about it as soon as possible, i.e. we want it to raise an exception.

The Python convention for avoiding name shadowing or keywords is generally to append a `_` or, in this case, you could simply to rename the variable to better reflect its relationship to the one it shadows. 
Prefixing names with underscores has special uses, please don't make up your own.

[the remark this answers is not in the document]
Comment (beepbeat) [reply Modules/database_manager.py]: If it's a dict, I'll gladly change it to use indexing, for me the config was always a class.

Due to how it's used, it'll raise an exception a few lines later, as None does not expose any members. 

If that's a convention in python to append, I'm sorry I haven't read through all the 420+ PEPs. I'm used to prepend from my previous style convention I followed.

File: Modules/database_manager.py [hunk 6]
```diff
@@ -0,0 +1,249 @@
...
+        Returns: a list containing the rows returned
+
+        """
+        if await self.has_table(table_name):
+            if not condition:
+                condition = {}
+            cond_str = ""
+            for k, v in condition.items():
+                cond_str += f" {k} = '{v}' {connector}"
+            cond_str = cond_str[0:-len(connector) - 1]
```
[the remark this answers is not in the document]
beepbeat: That's a nice one. Tried it that way because that's how python encourages you to do it, but I couldn't figure it out.

Modules/database_manager.py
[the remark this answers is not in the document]
MHajoha: @beepbeat, neither have I and it's alright to not know the convention, that is why I am telling it to you.
(on a side note, the conventions are only in pep8, most peps are interpreter or standard library features.)

theunkn0wn1:
Thanks for opening this PR again, while i am still reviewing the code contribution i can note a few required points off the bat: 
Diff coverage must exceed 85% (currently 66%)
Issues Raised by CC must be addressed prior to this being merged
overall coverage has fallen outside acceptable bounds, probably due to the diff coverage being outside required bounds

tests/conftest.py
beepbeat: It actually has a reason.
1. It is used by a also session scoped fixture (setUp/tearDown of the DBM tests)
1. It's hard to break the DBM
1. This makes the tests faster as it re-uses the same instance for all tests

utils/abstract.py
```diff
@@ -0,0 +1,74 @@
+"""
+Abstract class from MHajoha's API branch
```
theunkn0wn1: Code written by someone else should probably be commited by that someone else.

tests/test_database_manager.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+# noinspection PyProtectedMember
+@pytest.mark.asyncio
+@pytest.mark.database
+@pytest.mark.usefixtures("prepare_dbm")
+class TestStuff(object):
+
+    @pytest.mark.skip(reason="Not a test")
+    def finalize(self, dbm_fx):
```
theunkn0wn1: skip is not needed, as pytest won't detect it as a test anyways.
pytest recognizes anything beginning with `test_` as a test by default

requirements.txt
```diff
@@ -2,5 +2,9 @@
 pytest
 pytest-cov
 pytest-asyncio
+pytest-ordering
```
theunkn0wn1: Why is this module necessary? by design tests should not depend on the state of other tests, therefore the order in which they are executed is irrelevant

pytest.ini
```diff
@@ -15,4 +15,5 @@ markers =
     user: marks a test as a user object test
     logging: marks a test as being a test for logging
     ratlib: marks a test as a ratlib test
+    facts: marks a test as a facts test
     setup_tests: setup test marker
```
theunkn0wn1: \n missing at EOF

odbcinst.ini
```diff
@@ -0,0 +1,15 @@
...
+CommLog=1
+UsageCount=1
+
+[PostgreSQL UNICODE]
+Description=PostgreSQL ODBC driver (Unicode version)
+Driver=psqlodbcw.so
+Setup=libodbcpsqlS.so
+Debug=0
+CommLog=1
+UsageCount=1
```
theunkn0wn1: \n missing at EOF

README.md
```diff
@@ -18,11 +18,13 @@ SPARK is currently **incomplete**.
...
+Before you can use the docker-compose, you need to create a volume called: ``p-data``
 
-If you would prefer to run the service locally, bear in mind you may need to run a local PostgreSQL Database for the fact module.
+If you would prefer to run the service locally, bare in mind you may need to run a local 
+PostgreSQL Database or the provided sql service via docker-compose for mecha's fact module.
 
-### Building via Docker
-We have provided a `docker-compose.yml` suitably configured for building and running Mecha
+###Building via Docker
+We have provided a `docker-compose.yml` suitably configured for building and running mechasqueak
```
theunkn0wn1: I would advise against making a one-line documentation change in this PR, as it is not within the scope of this PR
(this also appears to have caused a merge conflict)

utils/abstract.py
beepbeat: That's probably true, but then I shall ask you to decide between the DBM having an API (as a promise) or me copying over that class

tests/test_database_manager.py
beepbeat: you are right, this was a leftover.

requirements.txt
beepbeat: by design, they should not, but I want to make sure no other test executes while I mess with the connection.

README.md
beepbeat: this may have occured due to the 3-way merge this had to undergo when I nuked the git history

odbcinst.ini
beepbeat: resolved

README.md
beepbeat: according to git blame, you were the last person to touch that paragraph

theunkn0wn1: according to the diff, **your** pr changes this paragraph at this line.

beepbeat: yes, as my PR has changes from develop already merged into it, but: https://puu.sh/AWpCT/6362f40406.png

PR #52, opened 2018-07-01
SPARK-52: Make Rescue 's hash depdendent on an immutable property

theunkn0wn1:
This Pr changes `Rescue`'s hashing mechanism as to make it more closely abide the python hash contract.

Rather than build the hash based on the current state of mutable properties when the object is first hashed,
The hashing mechanism will use the hash of the internal `__rescue_id` UUID4

unless otherwise specified during Rescue creation, the `__rescue_id` will be generated on the spot.

 the `rescue_id` field is present in the constructor as to allow the ID to be set during Rescue Re-creation from the API.

Resolves SPARK-52


Modules/rat_rescue.py
```diff
@@ -33,30 +31,20 @@ class Rescue(object):
...
-                 code_red=False):
+                 mark_for_deletion: MarkForDeletion = MarkForDeletion(), lang_id: str = "EN",
+                 rats: List[Rat] = None, status: Status = Status.OPEN, code_red=False,
+                 rescue_uuid: Optional[UUID] =None):
         """
         creates a unique rescue
 
         Args:
 
+            rescue_uuid ():
```
MHajoha: This seems to have slipped throught the cracks.

Modules/rat_rescue.py
```diff
@@ -82,6 +70,10 @@ def __init__(self, uuid: UUID,
             irc_nickname (str): clients IRC nickname, may deffer from their
                 commander name.
             rats (list): identified (Rat)s assigned to rescue.
+            rescue_uuid(UUID): Internal uuid used for hashing
```
MHajoha: Perhaps the constructor arg and property name could be changed to clarify its purpose. `rescue_uuid` could refer to the api id as well. Something along the lines of `internal_id` of `hashing_id`.

Modules/rat_rescue.py
```diff
@@ -106,6 +98,8 @@ def __init__(self, uuid: UUID,
         self._lang_id = lang_id
         self._status = status
         self._hash = None
+        self.__rescue_id = rescue_uuid if rescue_uuid is not None else uuid4()
```
MHajoha: What's with the two underscores? This has special meaning in Python: It invokes name mangling. In short, whenever you refer to `self.__rescue_id` here, Python will replace that with `self._Rescue__rescue_id`. This is usually to prevent conflicts in subclasses, I don't believe we need such a thing here.

Modules/rat_rescue.py
```diff
@@ -150,25 +144,10 @@ def __eq__(self, other) -> bool:
...
-                self.outcome,
-                self.title,
-                self.first_limpet,
-                self.lang_id,
-                self.irc_nickname,
-            )
-
-            self._hash = reduce(xor, map(hash, attributes))
+            # compute the hash of the internally generated uuid and use that as our hash
+            # this resolves SPARK-52
```
MHajoha: I don't believe source code comments should be a place to mention which issue is resolved by which piece of code. We have PRs and commit messages for that.

MHajoha (changes requested):
A couple of things.

In addition, I believe that the `__eq__` should also be replaced with the same logic. The whole idea of this rather more involved solution was to satisfy both the hash-doesn't-change as well as the hash-and-`__eq__`-agree requirements, was it not?

Modules/rat_rescue.py
beepbeat: Isn't it marking it as "private" (as private as something can be in Python) too, therefore making PyCharm (and other IDEs) hinting you towards that fact ?

MHajoha: @beepbeat, well it's supposed to be private, just not mangled. The attribute is also not accessed from outside the class, which would fail as it couldn't find the non-mangled name.

theunkn0wn1: I have marked it as private as it is internal only, i named it as such specifically to invoke name mangling as i fail to see why we would need to expose this attribute as part of `Rescue`'s public interface

theunkn0wn1:
`Rescue.__eq__` is the only means of verifying if two `Rescue` objects are equal. If we replaced `.__eq__` with the same logic, even if two objects were otherwise exactly the same they would always come across as unequal. while checking if the internal id is equal would satisfy the python contract, it would be useless to us.

MHajoha:
**TL;DR** (but please do): Not changing `__eq__` as well still violates the hash contract as before and does not fix what this PR is supposed to fix.

@theunkn0wn1, isn't the whole point of this new UUID to be able to have a concrete way of deciding whether two rescues are the same, without relying on attributes which are mutable (client name, for instance), ones which may collide if we're unlucky (created_at and updated_at) or ones which may not exist (api id)?

This ID would fulfill that purpose, telling us: Do these objects refer to the same actual rescue, no matter the updates made to them? In that spirit, I believe that `__eq__` needs to change, too. Not to mention the fact that the alternative still violates part of the the hash contract, that equal objects _must_ have the same hash.

Even more severely, without also changing `__eq__`, this whole PR is useless, as we would be right where we started, because, to be blunt with you: The title and, to a degree, the description of SPARK-52 are simply incorrect. `Rescue` could be safely hashed and used in hashed collections before, as modification of the public properties of the rescue did not impact the hash once it had been computed. The issue is that it did not satisfy the exact same requirement which this PR disregards: That equal objects _must_ have the same hash and unequal objects should try to have different hashes. The current implementation of `Rescue.__eq__` uses the current state of the attributes (anything else would be madness), while the hash is only calculated once, albeit from the same attributes. This means that any change to the rescue object will cause the equality requirement to be broken.

I'll even go one step further: This PR makes the current, already undesireable, situation worse. The hash only being calculated once, but in the same way that `__eq__` does its check, meant that the requirement was satisfied at least for a small amount of time. (This, of course, isn't nearly enough to call the contract satisfied, that is why I agree that a change needs to be made.)
The current state of this PR, however, while it **keeps** the **already present** fulfillment of the requirement that hashes do not change, it completely throws out of the window even that shred of fulfilling the equality requirement.

theunkn0wn1:
Ok. So. If i change `__eq__` to abide by the python equality and hashing contract, i will need to implement some other way of asserting a rescue has been modified. The failing tests revolve around detecting if a call to `ratboard.modify` which currently depends on the old equality method to verify the rescue was, in fact, updated.

this check is in place due to a bug originating in mecha2 where some commands modifying rescues would report success even though there wasn't an actual change leading to unexpected behavior.

theunkn0wn1:
this could probably be done by adding a callback in each property setter, il have to investigate

MHajoha:
I suppose you could just check the attributes in question, manually. This also has the benefit of not relying on `Rescue.__eq__` working correctly.

Though, mentioning the modification of a rescue, wasn't there once an idea of tracking which attributes on a rescue have been changed? I seem to remember something along.the lines of `changedAttrs`. Do you plan on reimplementing that?

beepbeat:
My thoughts on this went towards having a Rescue.has_changes(Rescue) or has_changes(Rescue, Rescue) that checks similar to, what I think was, the "old" __eq__, providing the functionality while not violating python contracts.

theunkn0wn1:
That would require tracking the old state of the object, and dynamically building the delta on the fly.
While this is valid, i believe it would be less overhead to build the delta in real time, as properties are set.

MHajoha:
What @beepbeat suggests might not be a bad idea.

As for adding properties to `changed_attrs`, a custom property class seems to me the simplest option, otherwise the `self.changed_attrs += "me"` line would have to be added to every property accessor. Either way, the MarkedForDeletion etc. objects would have to communicate back that they have been changed, likely by giving them some sort of `owner: Rescue` attribute.

theunkn0wn1:
by making each child object responsible for notifying on change that would increase the total number of required function calls / attribute sets necessary. Methinks the best approach is for Rescue to be responsible for looking after its own properties. Or, rather, each of `Rescue`'s properties being responsible for notifying on change.

This would keep the individual dataclasses independent of `Rescue`

I think the simplest, or at least most practical solution would be to add a property
`Rescue.changed_attrs: Set[str]` that would be updated each time one of Rescue's property setters completes setting their respective properties

beepbeat:
Sounds like a way to go

theunkn0wn1:
Ok. ive made some changes to the data structure.

`RescueBoard.modify` has been refactored into `.update`, and is simply a call to notify the API that a rescue has been changed.

`Rescue.change` context manager
--------------------
`Rescue.change` has become an async context manager, which required adding a new project dependency. 

The generator will, on entry, make a logging call and reset `Rescue.modified_attrs`
We may want to place an assertion here to ensure someone didn't modify the rescue outside the tracked `.change` context as any untracked changes will lost.

This context manager will yield the Rescue's `RatBoard` object, may prove useful for Commands

On exit the context manager will check if the rescue was changed, if not it raises a warning.
if it was, it updates the `.updated_at` field.

If the rescue is assigned to a board, it will then call the board to update the case via `RatBoard.update`

Modules/rat_rescue.py
```diff
@@ -121,54 +122,15 @@ def __eq__(self, other) -> bool:
...
-                self.first_limpet,
-                self.lang_id,
-                self.irc_nickname,
-            )
-
-            self._hash = reduce(xor, map(hash, attributes))
+            # compute the hash of the internally generated uuid and use that as our hash
+            self._hash = hash(self.__hashing_uuid)
+
         return self._hash
```
MHajoha: This doesn't change much, but I think we can drop the lazy initialization now that the algorithm is so much simpler and the hashing id doesn't change.

MHajoha (changes requested):
Just two minor ideas.

Modules/rat_rescue.py
```diff
@@ -862,35 +880,30 @@ def unmark_delete(self) -> None:
...
+        if len(self.modified_attrs) != 0:
+            self.updated_at = datetime.utcnow()
+        else:
+            log.warning(f"rescue #{self.api_id}({self.client}) was not changed within .change!")
 
-    # TODO: to/from json
-    # TODO: track changes
-    # TODO: helper method for adding / editing quotes
+        if self.rat_board is not None:
+            await self.rat_board.update(rescue=self)
```
MHajoha: I have a bit of a headache about this method and the associated `RatBoard.update`.

Why does the board have to be involved in updating the rescue? I don't believe any data from `RatBoard` enters into it. It seems to me that this needlessly complicates the process. Apart from making things more straightforward, leaving the updating in `Rescue.change` would: 
* Allow the updating of rescues not attached to a board (closed ones) which is probably a needed feature.
* Less importantly, it would avoid making `modified_attrs` public, which I don't believe it needs to be.

In the same spirit, I also believe that yielding the board is confusing and not very useful at all.

Modules/rat_rescue.py
```diff
@@ -108,7 +121,12 @@ def __init__(self, uuid: UUID,
         self._board_index = board_index
         self._lang_id = lang_id
         self._status = status
-        self._hash = None
+        self._mecha_uuid = mecha_id
```
MHajoha: This is a nitpick, but could you decide on either `mecha_id` or `mecha_uuid`?
Furthermore, I think a it should have a getter so it can be used in more places.

MHajoha (changes requested):

Modules/rat_rescue.py
theunkn0wn1: *blinks* yeah that was supposed to be `mecha_uuid`. will fix

Modules/rat_rescue.py
theunkn0wn1: Methinks i did it this way as i originally intended the board to be the thing interfacing with the API rather than the `Rescue` objects. That said, upon consideration perhaps it is indeed a bit of a over-complication.

The question becomes how do we intend the flow of information to work?
- If the rescue is not on a board, then wouldn't it not be possible to update the rescue / method wouldn't do anything?

 - If we make Rescue directly update the API, the Rescue object will need a reference to the API. im thinking this could be accomplished by adding it to the `Rescue` constructor, with a default of `None`. Rescues would be created through their respective boards (which would include closed ones?) which would fill in the API reference.

theunkn0wn1: I think imma put this Pr on hold until after the Event system is completed, then i can remove the headache when it comes to who does what in terms of updating a rescue (will result in the board loosing its `update` method / retooling it for something else.

tests/test_rat_board.py
```diff
@@ -265,37 +239,50 @@ def test_next_free_mixed_board(self, keys: list, expected: int, rat_board_fx: Ra
...
+    async def test_update_garbage(self, garbage, rat_board_fx: RatBoard):
+        with pytest.raises(TypeError):
+            await rat_board_fx.update(garbage)
+
+    @pytest.mark.asyncio
+    async def test_update_wrong_board(self, rescue_sop_fx: Rescue, rat_board_fx: RatBoard):
+        """verifies what happens when someone tries to update a Rescue that is not on the board"""
+        rescue_sop_fx.rat_board = None
+        with pytest.raises(ValueError):
+            await rat_board_fx.update(rescue_sop_fx)
```
theunkn0wn1:
```suggestion
            await rat_board_fx.update(rescue_sop_fx)

```
`\n` at eol 😉

theunkn0wn1:
This PR has sat stale for too long, and has become stale. It currently has merge conflicts (when rebasing) and is likely out of date. Methinks its in our best interests for me to rewrite the Pr at this point

PR #53, opened 2018-07-02
SPARK-53 fix silly @require_channel decorator requirement

theunkn0wn1:
This was a bit of a doozy.

As described in SPARK-53 there was a bug that required the `@require_channel` and `@require_dm` to be used as follows:
```py
@require_channel()
async def my_command(...)
```

the empty `()` being an anoying requirement. This PR fixes that.

the valid usages for the decorators has been corrected, can can be used in either of the following forms
```py
@require_channel
async def my_command(...)
```
which will use the default access denied message.

```py
@require_channel(message="Heck no.")
async def my_command(...)
```
The kwarg **must be specified**, if the kwarg is not specified it will be ignored and the function may be have unexpectidly. 

Fixes SPARK-53

### DIFF
--- a/Modules/permissions.py
+++ b/Modules/permissions.py
@@ -154,7 +154,7 @@ def require_channel(func: Union[str, Callable] = None,
         ... async def my_command(context: Context):
         ...     pass
     """
-    # form of @decorate("message")
+    # form of @decorator("message") and @decorator(message=str)
     if isinstance(func, str):
         message = func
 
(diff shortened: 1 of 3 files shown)


Modules/permissions.py
```diff
@@ -131,7 +132,7 @@ def real_decorator(func):
     return real_decorator
 
 
-def require_channel(message: str = "This command must be invoked in a channel."):
+def require_channel(*args, **kwargs):
```
MHajoha: It should still be possible to have a nice signature without varargs here. Simply have one parameter which can be either the message or the function. The same goes for `require_dm`.

theunkn0wn1: Oh, and `len(args)` can also be zero, which is a key reason i cannot have a clean signature as python wouldn't accept it.
```py
@require_channel
async def...
```
would invoke `require_channel` with zero arguments

```py
@require_channel()
```
would invoke `require_channel` with exactly one argument: the function
```py
@require_channel(message="potato")
```
would invoke require_channel` with exactly one argument and exactly one kwarg

MHajoha: `require_channel(message: Union[str, Callable] = "default message")` would work fine. To go through the options:
* `@require_channel` would pass the function as the first and only argument (i.e. message).
* `@require_channel()` would use the default message.
* `@require_channel(message="message")` and `@require_channel("message")` both pass the provided message.

You would then merely need to test whether `message` is a function or a string.

MHajoha: For an example, see `pytest.fixture`.

theunkn0wn1:
Signature fixed.

PR #54, opened 2018-07-03
SPARK-54, SPARK-55 and SPARK-56: Rule decorator improvements

MHajoha:
This PR implements SPARK-55 and the SPARK-54, which it requires. It also implements SPARK-56.

It adds a `pass_match` parameter to the rule decorator to implement SPARK-55. It also adds a parameter for toggling case-sensitivity as well as one to switch whether the regex should match the entire message (minus the prefix on non-prefixless rules, see below) or just the first word (for command-like rules).

SPARK-56's prefixless rules are also implemented. This did, however, need some doing in the command triggering logic: `MechaClient.on_message` now only checks that the message does not come from Mecha itself. `rat_command.trigger` tries to find a command, a regular old rule or a prefixless rule, in that order. If it can't find anything, it returns, possibly logging a warning. These changes also remove the need for `CommandNotFoundException`.

To facilitate simpler tests, this PR also adds callable mocks somewhat similar to those from unittest. These can be inquired as to if and how they were called. There is also an async version which can be treated as a coroutine function.

### DIFF
--- a/Modules/permissions.py
+++ b/Modules/permissions.py
@@ -225,10 +225,10 @@ def require_permission(permission: Permission,
         log.debug(f"Wrapping a command with permission {permission}")
 
         @wraps(func)
-        async def guarded(context: Context):
+        async def guarded(context: Context, *args):
             if context.user.identified and context.user.hostname in _by_vhost.keys() \
                     and _by_vhost[context.user.hostname] >= permission:
-                return await func(context)
+                return await func(context, *args)
             else:
                 await context.reply(override_message if override_message
                                     else permission.denied_message)
@@ -283,7 +283,7 @@ def require_channel(func: Union[str, Callable] = None,
         """
 
         @wraps(wrapped)
-        async def guarded(context: Context) -> Any:
+        async def guarded(context: Context, *args) -> Any:
             """
             Enforces channel requirement
 
@@ -294,7 +294,7 @@ def require_channel(func: Union[str, Callable] = None,
                 Any: whatever the called function returned
             """
             if context.channel is not None:
-                return await wrapped(context)
+                return await wrapped(context, *args)
             else:
                 log.debug(f"channel was None, enforcing channel requirement...")
                 await context.reply(message)
@@ -354,7 +354,7 @@ def require_dm(func: Union[str, Callable] = None,
         """
 
         @wraps(wrapped)
-        async def guarded(context: Context) -> Any:
+        async def guarded(context: Context, *args) -> Any:
             """
             Enforces channel requirement
 
@@ -365,7 +365,7 @@ def require_dm(func: Union[str, Callable] = None,
                 Any: whatever the called function returned
             """
             if context.channel is None:
-                return await wrapped(context)
+                return await wrapped(context, *args)
             else:
                 log.debug(f"channel was None, enforcing channel requirement...")
                 await context.reply(message)
--- a/Modules/rat_command.py
+++ b/Modules/rat_command.py
@@ -229,8 +229,9 @@ def command(*aliases):
 
 _RuleTuple = NamedTuple("_RuleTuple", underlying=Callable, full_message=bool, pass_match=bool)
 
-def rule(regex: str, *, case_sensitive: bool=False, full_message: bool=False, pass_match: bool=False,
-         prefixless: bool=False):
+
+def rule(regex: str, *, case_sensitive: bool=False, full_message: bool=False,
+         pass_match: bool=False, prefixless: bool=False):
     """
     Decorator to have the underlying coroutine be called when two conditions apply:
     1. No conventional command was found for the incoming message.
--- a/Modules/rat_command.py
+++ b/Modules/rat_command.py
@@ -14,7 +14,7 @@ This module is built on top of the Pydle system.
 
 import logging
 import re
-from typing import Callable
+from typing import Callable, Dict, Pattern, NamedTuple
 
 from pydle import BasicClient
 
@@ -55,7 +55,7 @@ class NameCollisionException(CommandException):
 
 
 _registered_commands = {}
-_rules = {}
+_rules: Dict[Pattern, "_RuleTuple"] = {}
 
 # character/s that must prefix a message for it to be parsed as a command.
 prefix = config['commands']['prefix']
@@ -105,10 +105,17 @@ async def trigger(message: str, sender: str, channel: str):
     if words[0].casefold() in _registered_commands.keys():
         return await _registered_commands[words[0].casefold()](context)
     else:
-        for key, value in _rules.items():
-            match = key.match(words[0])
+        for pattern, (coro, full_message, pass_match) in _rules.items():
+            if full_message:
+                match = pattern.match(words_eol[0])
+            else:
+                match = pattern.match(words[0])
+
             if match is not None:
-                return await value(context, match)
+                if pass_match:
+                    return await coro(context, match)
+                else:
+                    return await coro(context)
         else:
             raise CommandNotFoundException(f"Unable to find command {words[0]}")
 
@@ -180,15 +187,28 @@ def command(*aliases):
     return real_decorator
 
 
-def rule(regex: str, case_sensitive: bool=False):
+_RuleTuple = NamedTuple("_RuleTuple", underlying=Callable, full_message=bool, pass_match=bool)
+
+def rule(regex: str, case_sensitive: bool=False, full_message: bool=False, pass_match: bool=False):
     """
     Decorator to have the underlying coroutine be called when two conditions apply:
     1. No conventional command was found for the incoming message.
     2. The command matches the here provided regular expression.
 
     Arguments:
-        regex (str): Regular expression to match the command.
-        case_sensitive (bool): Whether to match case-sensitively (using the re.IGNORECASE flag).
+        regex (str):
+            Regular expression to match the command.
+        case_sensitive (bool):
+            Whether to match case-sensitively (using the re.IGNORECASE flag).
+        full_message (bool):
+            If this is True, will try to match against the full message. Otherwise,
+            only the word will be matched against.
+        pass_match (bool):
+            If this is True, the match object will be passed as an argument toward the command
+            function.
+
+    Please note that *regex* can match anywhere within the string, it need not match the entire
+    string. If you wish to change this behaviour, use '^' and '$' in your regex.
     """
     def decorator(coro: Callable):
         if case_sensitive:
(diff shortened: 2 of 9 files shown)


tests/mock_callables.py
```diff
@@ -0,0 +1,97 @@
...
+    def __eq__(self, other):
+        if isinstance(other, self.klass):
+            return True
+        elif isinstance(other, InstanceOf):
+            return self.klass is other.klass
+        else:
+            return NotImplemented
+
+
+class CallableMock(object):
```
theunkn0wn1: if we need something akin magicmock from `unittest` we can make use of the [pytest-mock](https://github.com/pytest-dev/pytest-mock) library. this may work better than implementing our own solution.

tests/mock_callables.py
```diff
@@ -0,0 +1,97 @@
...
+    Examples:
+        >>> InstanceOf(str) == "hello"
+        True
+        >>> InstanceOf(dict) == 7
+        False
+    """
+    __slots__ = ("klass",)
+
+    def __init__(self, klass: type):
+        self.klass = klass
```
theunkn0wn1: I take it this is some from of pun?

Modules/permissions.py
```diff
@@ -118,10 +118,10 @@ def real_decorator(func):
         log.debug(f"Wrapping a command with permission {permission}")
 
         @wraps(func)
-        async def guarded(context: Context):
+        async def guarded(context: Context, *args):
```
theunkn0wn1: this will cause a merge conflict with #53

Modules/rat_command.py
```diff
@@ -76,18 +70,82 @@ class NameCollisionException(CommandException):
...
-    # remove command prefix
-    raw_command: str = message.lstrip(prefix)
+    Args:
+        string: Any string.
 
+    Returns:
+        (list of str, list of str):
+            A 2-tuple of (words, words_eol), where words is a list of the words of *string*,
+            seperated by whitespace, and words_eol is a list of the same length, with each element
+            including the word and everything up to the end of *string*
```
theunkn0wn1: perhaps an example in the docstring could be valuable since we have begun adopting doctest?

tests/mock_callables.py
MHajoha: Much of pytest-mock is simply a wrapper around unittest, in fact, `mocker.stub()` seems to just return a `MagicMock` object, which you didn't want me using.

I also am not a huge fan of `MagicMock`, so I've implemented this much simpler and in my opinion easier to use module. It also makes it so much easier to have mock coroutine functions.

tests/mock_callables.py
MHajoha: What? No, why would this be a pun? `klass` is to avoid the keyword `class` :D

Modules/permissions.py
MHajoha: I've just given that a review. Once it's merged, I'll update my branch, there's no issue.

Modules/rat_command.py
MHajoha: In a private member, I wasn't going to bother with extensive documentation. But I guess it can't hurt. Will add.

main.py
```diff
@@ -81,12 +81,6 @@ def __init__(self, *args, **kwargs):
             # self-stimulated positive feedback loop.
             log.debug(f"Ignored {message} (anti-loop)")
             return None
-
-        if not message.startswith(rat_command.prefix):
-            # prevent bot from processing commands without the set prefix
-            log.debug(f"Ignored {message} (not a command)")
```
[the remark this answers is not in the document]
MHajoha: Eh, don't think it's necessary but it also can't hurt. Sure

Modules/rules.py
```diff
@@ -0,0 +1,107 @@
...
+            else:
+                return fun, ()
+    else:
+        return None, ()
+
+
+def clear_rules():
+    global _prefixless_rules, _rules
+    _prefixless_rules = []
+    _rules = []
```
theunkn0wn1: `list.clear()` can be called here to clear the rules, this would remove the need for the `global` keyword

tests/test_rat_command.py
```diff
@@ -35,16 +32,16 @@ def Setup_fx(bot_fx):
...
-        Ensures the proper exception is raised when a command is not found.
-        :return:
+        Ensures that nothing happens and `trigger` exits quietly when no command can be found.
         """
-        with pytest.raises(CommandNotFoundException):
+        try:
             await Commands.trigger(message="!nope", sender="unit_test",
                                    channel="foo")
+        except BaseException as e:
+            pytest.fail("trigger raised " + type(e).__name__)
```
theunkn0wn1: why catch the exception, methinks pytest will give better introspection if we let it catch the exception instead.

unhandled / unexpected exceptions in tests automatically fail a test.

Modules/rat_command.py
```diff
@@ -76,18 +70,88 @@ class NameCollisionException(CommandException):
...
+
+def _split_message(string: str) -> Tuple[List[str], List[str]]:
+    """
+    Split up a string into words and words_eol
 
+    Args:
+        string: Any string.
+
+    Returns:
+        (list of str, list of str):
```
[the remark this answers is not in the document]
MHajoha: This is the usual format Google uses for type hints. I don't actually think pycharm correctly displays typing library stuff in docstrings.

Modules/rat_command.py
```diff
@@ -166,54 +214,66 @@ def command(*aliases):
...
-        return wrapper
+        return func
     return real_decorator
 
 
-def rule(regex: str):
+_RuleTuple = NamedTuple("_RuleTuple", underlying=Callable, full_message=bool, pass_match=bool)
+
+
+def rule(regex: str, *, case_sensitive: bool=False, full_message: bool=False,
```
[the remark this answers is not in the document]
MHajoha: It makes sure that all the parameters after it must be referenced via keyword, to force clarity.

Modules/rules.py
MHajoha: True, neat idea

tests/test_rat_command.py
MHajoha: You're right, I think I changes this one about a couple of times, but in the end I went for the explicit fail to make it clear that `trigger` not raising an exception is what was being tested.

`with pytest.does_not_raise:` is something I would have preferred but it doesn't exist.

Modules/rat_command.py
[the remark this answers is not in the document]
MHajoha: You'll note that the signature does have the typing type hint

Modules/rules.py
```diff
@@ -0,0 +1,132 @@
...
+from typing import Callable, NamedTuple, Pattern, List, Tuple, Optional
+
+log = logging.getLogger(__name__)
+
+_rules: List["Rule"] = []
+_prefixless_rules: List["Rule"] = []
+
+
+class Rule(NamedTuple):
+    pattern: Pattern
```
theunkn0wn1: Are these intentionally class attributes as opposed to instance attributes?

Modules/rat_command.py
```diff
@@ -76,18 +68,58 @@ class NameCollisionException(CommandException):
...
+    if message.startswith(prefix):
+        if words[0].casefold() in _registered_commands.keys():
+            # A regular command
+            command_fun = _registered_commands[words[0].casefold()]
+            extra_args = ()
+            log.debug(f"Regular command {words[0]} invoked.")
+        else:
+            # Might be a regular rule
+            command_fun, extra_args = get_rule(words, words_eol, prefixless=False)
+            if command_fun:
```
theunkn0wn1: Value is always Truthy.

Modules/rules.py
```diff
@@ -0,0 +1,132 @@
...
+        else:
+            match = pattern.match(words[0])
+
+        if match is not None:
+            if pass_match:
+                return fun, (match,)
+            else:
+                return fun, ()
+    else:
+        return None, ()
```
theunkn0wn1: This expression is always truthy, which [causes issues in other modules when a truthy check is performed.](https://github.com/FuelRats/pipsqueak3/pull/54/files#diff-f3ecc38ebed1a7b33077a674020998efR84)
```py
In [1]: def foo():
   ...:     return None, ()
   ...:
   ...:

In [2]: retn = foo()

In [3]: retn
Out[3]: (None, ())

In [4]: bool(retn)
Out[4]: True

In [5]: bool(foo())
Out[5]: True

In [6]: if foo():
   ...:     print("truthy!")
   ...:
truthy!
```

My suggestion here is to simply return `None` and let the caller check the return with an explicit None check

Modules/rat_command.py
theunkn0wn1: oh wait no its not 🤦‍♂️ 
variable unpacking `/tableflip`

Modules/rules.py
MHajoha: This is the 'new' syntax for NamedTuple. It looks at the class annotations and constructs a NamedTuple from it. This is equivalent to `Rule = NamedTuple("Rule", pattern=Pattern, ...)`, however it allows me to define additional methods.

Modules/rules.py
MHajoha: I disagree that this made it less clear. The meaning of the return value is: No rule was found and no arguments should be appended to the call. [Returning None in place of the tuple will make this less clear](http://thefinestartist.com/effective-java/43) in my opinion, and returning just a `None` will obfuscate client code:
```python
result = get_rule(...)
if result is not None:
  fun, extra_args = result
```
instead of 
```python
fun, extra_args = result
```
to achieve the same result.

MHajoha:
You're right, I missed `get_rule` after I moved it out of `trigger`. Will add.

MHajoha:
Sorry for the late response - I'm afraid I don't quite get what you mean. What exactly would `invoke` do, and where would it sit? Do you intend to make the rule invocation a subscriber to on_message?

theunkn0wn1:
I would like `invoke` to sit somewhere in your rules module.
The function is that `invoke` can become a event subscriber (probably `on_message_raw`, not decided yet).

This also moves the rule handling logic out of rat_command where it does not belong.

Since Event is still pending a function call to your emit would be appropriate, i can refactor that out in the Event system PR.

MHajoha:
As I noted in IRC, the rules would require two events and two invocation functions. One for 'regular' rules (on_command.on_not_implemented) and one for 'prefixless' rules (on_message). It seems to me that this would make things more complicated than they need to be, especially since commands, rules and prefixless rules all receive the same arguments, plus optionally the match argument. The code for this would essentially need to be duplicated for all three.

I honestly don't think events are a good fit for this, basing core functionality on them is opening a can of worms, such as the issue that only either a command or a rule must be invoked. I believe events are better suited for the command side of things, e.g. saving previous messages for !grab. (I believe that is what you built the events for in the first place?)

PR #55, opened 2018-07-04
feature/spark-58

nathanctech:
This PR modified add_rat() to return a Rat object of the rat added to the rescue. This is to ensure !assign works correctly and could be used for others.

This resolves SPARK-58. SPARK-52 may need to be merged first before this one is.

### DIFF
--- a/Modules/rat_rescue.py
+++ b/Modules/rat_rescue.py
@@ -745,7 +745,7 @@ class Rescue(object):
         else:
             raise TypeError(f"expected type list got {type(value)}")
 
-    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) -> None:
+    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) -> Rat:
         """
         Adds a rat to the rescue. This method should be run inside a `try` block, as failures will
         be raised as exceptions.
@@ -759,7 +759,7 @@ class Rescue(object):
             guid (UUID or str): api uuid of the rat, used if the rat is not found in the cache
                 - if this is a string it will be type coerced into a UUID
         Returns:
-            None:
+            Rat: the added rat object
 
         Raises:
             ValueError: guid was of type `str` and could not be coerced.
@@ -774,7 +774,7 @@ class Rescue(object):
             # we already have a rat object, lets verify it has an ID and assign it.
             if rat.uuid is not None:
                 self.rats.append(rat)
-                return
+                return rat
             else:
                 raise ValueError("Assigned rat does not have a known API ID")
 
@@ -784,22 +784,23 @@ class Rescue(object):
                      await RatCache().get_rat_by_name(name))
             if found[0]:
                 self.rats.append(found[0])
-                return
+                return found[0]
             elif found[1]:
                 # a generic match (not platform specific) was found
                 # TODO throw a warning so the invoking method can handle this condition
                 log.warning("A match was found, but it was not the right platform!")
                 self.rats.append(found[1])
-                return
+                return found[1]
 
             else:
                 # lets make a new Rat!
                 if self.rat_board:  # PRAGMA: NOCOVER
-                    raise NotImplementedError  # TODO fetch rat from API
+                    pass  # TODO fetch rat from API
                 # TODO: fetch rats from API handler, use that data to make a new Rat instance
 
                 rat = Rat(name=name, uuid=guid)
                 self.rats.append(rat)
+                return rat
 
     def mark_delete(self, reporter: str, reason: str) -> None:
         """
--- a/Modules/rat_rescue.py
+++ b/Modules/rat_rescue.py
@@ -745,7 +745,9 @@ class Rescue(object):
         else:
             raise TypeError(f"expected type list got {type(value)}")
 
-    async def add_rat(self, name: str = None, guid: UUID or str = None,
+    async def add_rat(self,
+                      name: str = None,
+                      guid: UUID or str = None,
                       rat: Rat = None) -> Optional[Rat]:
         """
         Adds a rat to the rescue. This method should be run inside a `try` block, as failures will
@@ -771,7 +773,7 @@ class Rescue(object):
 
             ```
         """
-        assigned_rat: Rat = None
+        assigned_rat: Optional[Rat] = None
 
         if isinstance(rat, Rat):
             # we already have a rat object, lets verify it has an ID and assign it.
@@ -811,7 +813,17 @@ class Rescue(object):
             if found:
                 self.rats.append(found)
                 assigned_rat = found
-            # if the specified UUID wasn't found, return None for "no rat added"
+            # else:  TODO: placeholder for fetching rats from the API handler
+                # if the specified UUID wasn't found, we need to search the API
+
+        elif isinstance(guid, str):
+            # attempt to coerce into a UUID
+            parsed_guid = UUID(guid)
+            found = await RatCache().get_rat_by_uuid(parsed_guid)
+            if found:
+                self.rats.append(found)
+                assigned_rat = found
+            # else: TODO: placeholder for fetching rats from the API handler
 
         return assigned_rat
 
--- a/Modules/rat_rescue.py
+++ b/Modules/rat_rescue.py
@@ -745,7 +745,7 @@ class Rescue(object):
         else:
             raise TypeError(f"expected type list got {type(value)}")
 
-    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) -> Rat:
+    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) -> Optional[Rat]:
         """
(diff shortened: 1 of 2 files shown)


Modules/rat_rescue.py
```diff
@@ -745,7 +745,8 @@ def rats(self, value):
         else:
             raise TypeError(f"expected type list got {type(value)}")
 
-    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) -> None:
+    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) \
+            -> Optional[Rat]:
```
theunkn0wn1: i assume this is on a newline due to pep8? i would rather you place the newline after a paramater than during the return type hint.

Modules/rat_rescue.py
```diff
@@ -770,11 +771,13 @@ def rats(self, value):
...
             ```
         """
+        found_rat: Rat = None
+
         if isinstance(rat, Rat):
             # we already have a rat object, lets verify it has an ID and assign it.
             if rat.uuid is not None:
                 self.rats.append(rat)
-                return
+                found_rat = rat
```
theunkn0wn1: `found_rat` is a bit of a misnomer in this context, as the `Rat` was provided as a argument. may i suggest renaming this variable `assigned_rat`?

Modules/rat_rescue.py
```diff
@@ -784,22 +787,33 @@ def rats(self, value):
...
                 # a generic match (not platform specific) was found
                 # TODO throw a warning so the invoking method can handle this condition
                 log.warning("A match was found, but it was not the right platform!")
                 self.rats.append(found[1])
-                return
+                found_rat = found[1]
 
             else:
                 # lets make a new Rat!
                 if self.rat_board:  # PRAGMA: NOCOVER
```
theunkn0wn1: We may want to comment this whole logical branch out, as it doesn't serve a purpose until we have the API to develop against (which i hope is soon)

tests/test_rescue.py
```diff
@@ -511,6 +512,67 @@ def test_mark_for_deletion_setter_bad_types(garbage, rescue_plain_fx: Rescue):
     assert name in rat_cache_fx.by_name
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("uuid,name", [(uuid4(), "foo"), (uuid4(), "bar"), (uuid4(), "potato")])
+async def test_add_rat_returns_rat_by_object(uuid: uuid4, name: str, rescue_plain_fx: Rescue):
```
theunkn0wn1: invalid type hint. `uuid4` is a classmethod. `UUID` is the type.

tests/test_rescue.py
```diff
@@ -511,6 +512,67 @@ def test_mark_for_deletion_setter_bad_types(garbage, rescue_plain_fx: Rescue):
...
+    """
+    Verifies `Rescue.add_rat` returns a proper `Rat` object when given a valid Rat object
+    """
+    result_rescue = rescue_plain_fx
```
theunkn0wn1: no reason to assign a local to the fixture, it is already local 😉

tests/test_rescue.py
```diff
@@ -511,6 +512,67 @@ def test_mark_for_deletion_setter_bad_types(garbage, rescue_plain_fx: Rescue):
...
+
+    test_rat = Rat(uuid, name)
```
theunkn0wn1: may i suggest using `rat_good_fx`?

tests/test_rescue.py
```diff
@@ -511,6 +512,67 @@ def test_mark_for_deletion_setter_bad_types(garbage, rescue_plain_fx: Rescue):
...
+
+    result = await result_rescue.add_rat(rat=test_rat)
+
+    assert test_rat == result
```
theunkn0wn1: methinks an assertion the test rat *is* the result is more appropriate, no?

Modules/rat_rescue.py
nathanctech: Although this check may never be hit, I'd prefer leaving it there. I even wrote a test for it >.>

Modules/rat_rescue.py
nathanctech: Yeah, will change.

tests/test_rescue.py
nathanctech: I copied another test (one of yours?) that has it set that way.... >.>

tests/test_rescue.py
nathanctech: Huh, neat...the more I learn about fixtures the more I like them.

Modules/rat_rescue.py
```diff
@@ -745,7 +745,8 @@ def rats(self, value):
         else:
             raise TypeError(f"expected type list got {type(value)}")
 
-    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) -> None:
+    async def add_rat(self, name: str = None, guid: UUID or str = None,
```
theunkn0wn1: grr, while this would take more vertical space methinks each param should be on a newline. the place you choose to split is just awkward (nitpick)

Modules/rat_rescue.py
```diff
@@ -770,11 +771,13 @@ def rats(self, value):
 
             ```
         """
+        assigned_rat: Rat = None
```
theunkn0wn1: type is `Optional[Rat]` 😜

Modules/rat_rescue.py
```diff
@@ -784,22 +787,33 @@ def rats(self, value):
...
+                #    pass  # TODO fetch rat from API
                 # TODO: fetch rats from API handler, use that data to make a new Rat instance
 
                 rat = Rat(name=name, uuid=guid)
                 self.rats.append(rat)
+                assigned_rat = rat
+
+        if isinstance(guid, UUID):
+            # lets check if we already have this rat in the cache
+            found = await RatCache().get_rat_by_uuid(guid)
```
theunkn0wn1: "the absence of evidence is not evidence of absence"
 - rat might not be in the cache 😉

tests/test_rescue.py
```diff
@@ -511,6 +512,43 @@ def test_mark_for_deletion_setter_bad_types(garbage, rescue_plain_fx: Rescue):
...
 
 
+@pytest.mark.asyncio
+async def test_add_rat_returns_rat_by_object(rat_good_fx: Rat, rescue_plain_fx: Rescue):
+    """
+    Verifies `Rescue.add_rat` returns a proper `Rat` object when given a valid Rat object
+    """
+    result = await rescue_plain_fx.add_rat(rat=rat_good_fx)
+
+    assert result == rat_good_fx
```
theunkn0wn1: This should be an identity check rather than an equality check.

Modules/rat_rescue.py
theunkn0wn1: If the rat is not in the cache, we may need to query the API to see if the Rat exists. at the very least this branch should be accounted for, but unused as that depends on a feature that is not currently merged

Modules/rat_rescue.py
```diff
@@ -784,22 +789,43 @@ def rats(self, value):
...
+            if found:
+                self.rats.append(found)
+                assigned_rat = found
+            # else:  TODO: placeholder for fetching rats from the API handler
+                # if the specified UUID wasn't found, we need to search the API
+
+        elif isinstance(guid, str):
+            # attempt to coerce into a UUID
+            parsed_guid = UUID(guid)
```
theunkn0wn1: code duplicity. i would suggest checking if `guid` is a `str`, and attempt the coercion above line 810, that way you can reuse existing machinery

Modules/rat_rescue.py
nathanctech: It's just left unhandled right now...I could just add a TODO comment if that's what you were looking for (currently the comment says not found rats return None, which is the current behavior).

theunkn0wn1: i think that would be advisable, since most of us use an IDE that recognizes TODO items. As much as i know todos are ugly, i find thats better than ignoring an obvious missed branch.
That or make another JIRA ticket noting this branch was missed intentionally and make it blocked by the API ticket

Modules/rat_rescue.py
nathanctech: The code above this isn't duplicating since guid can be one of two types, I want to handle them as separate "branches" to make the code easier to follow. That's my personal preference, though.

Modules/rat_rescue.py
```diff
@@ -745,7 +745,10 @@ def rats(self, value):
         else:
             raise TypeError(f"expected type list got {type(value)}")
 
-    async def add_rat(self, name: str = None, guid: UUID or str = None, rat: Rat = None) -> None:
+    async def add_rat(self,
+                      name: str = None,
+                      guid: UUID or str = None,
```
theunkn0wn1: methinks all three of these paramaters should be of type `Optional[type]`
eg:
```py
name: Optional[str],
guid: Optional[UUID, str],
rat: Optional[Rat]

PR #57, opened 2018-07-14
SPARK-57: Database Manager and Facts, take IV.I 

beepbeat:


theunkn0wn1:
ok... we still are having issues with rebasing your PR. Also, curiously, you have merge commits for two ancient PRs on this branch, which is probably the source of the conflict. 

This probably occured because your forks develop branch is out of sync with the upstream, which will cause issues when rebasing. 

Suggested course of action: 
ensure your forks develop is up-to-date
copy the **end state** of this PR's files into an untracked directory
delete your local `feature/database_manager` branch
create a new `feature/database_manager` branch from develop
copy the files back in from the untracked directory
commit
force push

theunkn0wn1:
Branch is now mergable via rebase, il review the content in the following days.
image

theunkn0wn1:
OK... this took a lot more time to get running on my dev box than expected, most of which being user error.

Things that i think need to be changed based on that experience:

1. We should have a method for connecting via a DSN
2. Something done goof with the current server connection string, not sure what (might be PEBKAC)
3. we should have strong documentation for setting up a local test env with the ODBC drivers and whatnot.

beepbeat:
1. DSN can be done
1. Nope, just if you use a DSN, you need a different one
1. ODBC drivers are __usally__ easily installed, but MSODBC thinks, it is allowed to be special by only accepting predefined DSNs.

Modules/rat_command.py
```diff
@@ -63,6 +64,9 @@ class NameCollisionException(CommandException):
 # Pydle bot instance.
 bot: BasicClient = None
 
+# facts manager instance
+facts_m: FactsManager = FactsManager()
```
theunkn0wn1: please rename this to `facts_manager`, `facts_m` doesn't exactly describe what this field represents

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        checks for existence of a fact
+        Args:
+            name: name of the fact to check
+            lang: language to check
+
+        Returns: True if it exists, False if it exists
+
+        """
+        result = await self.dbm.select_rows("fact", "AND", {"lang": lang, "name": name})
```
theunkn0wn1: This introduces duplicity with `get_fact`.
I suggest changing `result` to await `self.get_fact`

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+import re
+import pyodbc
+import datetime
+
+internal_facts = ("!go", "!assign", "!add")
+# any fact name in here will not be handled by the handler
+
+
+class Fact:
+    name = None
```
theunkn0wn1: Im fairly certain these should not be class attributes unless `Fact` is a singleton?

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class FactsManager(Singleton):
+
+    @property
+    def enabled(self):
+        return self.dbm.enabled
+
+    @enabled.setter
+    def enabled(self, value):
```
theunkn0wn1: Then don't expose a setter if you don't want people setting it 😉 
Python will raise an `AttributeError` if someone tries setting a property that does not have a defined setter method.

beepbeat: It's merely a debugging aid, but I can remove it too ;)

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+        self.dbm = DatabaseManager()
+
+    async def get_fact(self, name: str, lang: str = "en") -> Fact or None:
+        """
+        retrieves a fact from the DB.
+        Args:
+            name: name of the fact to retrieve
+            lang: language to retrieve
+
+        Returns: message of the fact, or None should it not be found for the specified language
```
theunkn0wn1: Fairly certain this returns a `Fact` object, not the Facts message

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+                                      {"name": fact.name,
+                                       "lang": fact.lang,
+                                       "message": fact.message,
+                                       "author": fact.author},
+                                      ("name", "lang",)
+                                      )
+        except ValueError:
+            return False
+        else:
+            time = datetime.datetime.now(tz=datetime.timezone.utc)
```
theunkn0wn1: `datetime.datetime.utcnow`?

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        delete's the given fact
+        Args:
+            name: name of the fact to delete
+            lang: language to delete
+
+        Returns: True on success, False if it failed and None if the DBM is disabled
+
+        """
+        if not self.dbm.enabled:
```
theunkn0wn1: I believe we agreed this would raise?

Modules/rat_command.py
beepbeat: done

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Responds with the message of the fact, should a match be found and not listed as an
+        internal fact.
+
+        Args:
+            context: the context as passed by rat_command
+
+        Returns: True if it was a fact, False if not. Will also return False,
+            should the DBM be disabled.
+
```
theunkn0wn1: This docstring is confusing. the description section describes it as `Responds with the message of the fact`, which i interpret as a return.

Later in the `Returns` block it returns a boolean...

Modules/rat_facts.py
```diff
@@ -0,0 +1,187 @@
...
+        name = result.group("name")
+        lang = None
+
+        if not name:
+            name = result.group("name2")
+        else:
+            lang = result.group("lang")
+
+        if not lang:
```
theunkn0wn1: If this is intended to be a `None` check then `None` should be checked for explicitly.
(Falsy check where a none check was expected)

Modules/rat_facts.py
beepbeat: no, utcnow gives timezone-naive not timezone-aware datetimes.
Sadly.

theunkn0wn1: ah. darn.

Modules/rat_facts.py
beepbeat: amended to clarify

Modules/rat_facts.py
beepbeat: fixed

Modules/rat_facts.py
beepbeat: fixed

Modules/rat_facts.py
beepbeat: removed

Modules/database_manager.py
```diff
@@ -0,0 +1,432 @@
...
+                raise ValueError("Suspicion of SQL-Injection. Statement: SELECT * FROM {table_name}"
+                                 f" WHERE {cond_str}. Aborting")
+            # throw out the result of the query
+            return self._execute(f"SELECT * FROM {table_name} WHERE {cond_str};").fetchall()
+        else:
+            # let's raise them to be responsible adults
+            raise ValueError(f"Table {table_name} does not exists!")
+
+    async def has_table(self, name: str) -> bool:
+        """
```
theunkn0wn1: As I recall, docstrings done in the ABC need not be re-defined in the implementation.

Modules/database_manager.py
```diff
@@ -0,0 +1,432 @@
...
+             condition: conditions, connected by "equals"
+             skip_double_dash_test: skip the crude SQLInjection test if it breaks your request,
+                     implement your OWN CHECK!
+
+         Returns: None
+
+         """
+
+    @abstractmethod
+    async def delete_row(self, table_name: str, connector: str, condition=None,
```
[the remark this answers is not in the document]
beepbeat: Dict[str]

Modules/database_manager.py
```diff
@@ -0,0 +1,432 @@
...
+
+        """
+
+
+# noinspection SqlNoDataSourceInspection
+class DatabaseManager(DBMApi, Singleton):
+    _enabled = True
+
+    # for testing only, as a makeshift "session" scope
+    marker = None
```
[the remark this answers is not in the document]
beepbeat: this is to prevent trying to access the var without it being there. There is no downside to it but a byte or so reserved to have a representation of it.

Modules/database_manager.py
```diff
@@ -0,0 +1,432 @@
...
+    cursor = None
+
+    @property
+    def enabled(self):
+        return self._enabled
+
+    @enabled.setter
+    def enabled(self, value: bool):
+        if value == self._enabled:
```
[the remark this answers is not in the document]
beepbeat: 1. Does it really matter?
1. Equality seems more safe as I only care for the value anyways.
1. It works, you can't prove it broken, never touch a "working as intended" system.

Modules/database_manager.py
```diff
@@ -0,0 +1,432 @@
...
+        else:
+            self.connection.close()
+            self._enabled = False
+
+    def _execute(self, *args, **kwargs):
+        if self.enabled:
+            for retry in range(3):
+                try:
+                    return self.cursor.execute(*args, **kwargs)
+                except pyodbc.ProgrammingError as e:
```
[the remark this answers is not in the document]
beepbeat: 1. this is a widely used var name for exceptions (similar to i for iterative loop counters).
1. it exists for only ~5 lines anyways, so it's quite clear what it does from it's (very small) context.

Modules/database_manager.py
beepbeat: they do not need to be redefined, but it can't hurt to have them there again, as it's the function/method/procedure you will be pointed at by your IDE / the first to look at if you want to understand it a bit better, need docs. also I''m pretty sure it missing might break ctrl+q.

Modules/database_manager.py
```diff
@@ -0,0 +1,432 @@
...
+    @property
+    def enabled(self):
+        return self._enabled
+
+    @enabled.setter
+    def enabled(self, value: bool):
+        if value == self._enabled:
+            # no need to change it if no change is required
+            return
+        if value:
```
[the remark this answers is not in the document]
beepbeat: 1. On the one hand you could argue, that passing None will also disable it (can be intended)
1. On the other hand, passing None should trigger a lethal response, as it isn't really a valid value for the request.

Modules/database_manager.py
```diff
@@ -0,0 +1,432 @@
...
+from utils.ratlib import Singleton
+from abc import abstractmethod, ABC
+
+log = logging.getLogger(f"mecha.{__name__}")
+
+
+class DBMApi(ABC):
+    @abstractmethod
+    async def select_rows(self, table_name: str, connector: str, condition: dict = None,
+                          skip_double_dash_test: bool = False) -> list or None:
```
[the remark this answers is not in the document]
theunkn0wn1: in which case either stick with `Optional[List]` or use `Optional[List[Any]]`

Modules/database_manager.py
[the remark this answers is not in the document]
theunkn0wn1: Then apply a type hint please :)

Modules/database_manager.py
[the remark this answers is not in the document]
theunkn0wn1: 1: the type hint is `bool`, not `Optional[bool]` therefor 2 is accurate. `None` should trigger a lethal response. likewise, since a daft programmer could throw anything into `value` a Truthy/Falsy check may lead to unintended regressions

Modules/database_manager.py
theunkn0wn1: Doesn't break ctrl+q in pycharm at any rate, its smart and copies the docstring from the ABC.

but yeh, i suppose it doesn't hurt until the docstrings don't add up. right now they are duplicated, but what of the future?

Modules/database_manager.py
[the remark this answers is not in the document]
theunkn0wn1: 1. yeup.
2. think of comparing a value against None via equality: its incorrect (also pep8 disbars it)
3. `/shrug` indeed its not broken but that doesn't mean its functioning as intended

Modules/database_manager.py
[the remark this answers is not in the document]
theunkn0wn1: 1. so is `exc` which is what i find frequently in python code samples.
2. null argument is null. See [rule #1](https://github.com/FuelRats/tech-rat-code-standards/blob/master/README.md)

Modules/database_manager.py
```diff
@@ -0,0 +1,380 @@
...
+            # would raise the above catched exceptions after failing three times
+        else:
+            log.warning("Tried to execute a DB call with the DB access being disabled.")
+            raise RuntimeError("Check if the DB-module is enabled before executing code!"
+                               " Currently it is NOT!")
+
+    # noinspection PyAttributeOutsideInit
+    def init_connection(self):
+        # connect to PostgreSQL (PSQL) database
+        config_: dict = config["database"]
```
theunkn0wn1: nitpick: type hint should be `typing.Dict`

Modules/database_manager.py
```diff
@@ -0,0 +1,380 @@
...
+        connect_str = None
+        if config_["method"].lower() == "server":
```
theunkn0wn1: (minor) use `str.casefold` for caseless comparisons

Modules/database_manager.py
```diff
@@ -0,0 +1,380 @@
...
+            return
+        # we init the connection, as we are supposed to do
+        try:
+            self.init_connection()
+        except pyodbc.Error:
+            # something went terribly wrong and excrements hit the ventilation device
+            # using internal here to skip disconnection part as it might (and probably would) call
+            # stuff on a NoneType
+            self._enabled = False
+            return
```
theunkn0wn1: should the exception be re-raised? unless im mistaken this will silently disconnect the database
If this is intentional, we should at least log the exception

tests/test_rat_facts.py
```diff
@@ -0,0 +1,138 @@
...
+        assert value
+
+        # kick the fact out of the DB
+        await facts_fx.delete_fact("test1", "de")
+
+        # and now the fact should be gone
+        value = await facts_fx.is_fact("test1", "de")
+        assert not value
+
+    async def test_rule(self, facts_fx):
```
theunkn0wn1: I believe we have a fixture `bot_fx` which provides a instance of `tests.mock_bot`.

theunkn0wn1:
Looks good now, unfortunately recent changes have created a merge conflict 🙁 
Please resolve them and Il give final approval within the week

theunkn0wn1:
For some reason your CI build is failing, seems to be something to do with `coverage.py` .

We cannot merge PRs that break the CI build process.

Also, can we get the CC issues resolved? they should be trivial to correct😉

Modules/database_manager.py
```diff
@@ -0,0 +1,383 @@
...
+            raise ValueError(f"{type(value)} is not a valid type for setting a bool!")
+        if value == self._enabled:
+            # no need to change it if no change is required
+            return
+        if value:
+            try:
+                self.init_connection()
+                self._enabled = True
+            except pyodbc.Error as ex:
+                log.error(f"Connection reinitialization failed! {ex}")
```
theunkn0wn1: Shouldn't the exception be raised after the logging call? If we want this error not to be fatal the exception should be handled in the caller methinks

beepbeat: I can put a "raise" right there, no problem

Modules/rat_facts.py
UncleClapton: funny enough we encountered the problems with `datetime.utcnow` in mecha2 as well. Caused a major problem with trusting times set by mecha in the quote object.

theunkn0wn1: Yeah. i remember fixing that too. In one place timezone aware timestamps were written but in another timezone-naive. What we should probably do is have a sanity check here to ensure that the data is correctly formatted prior to interpreting it

beepbeat:
I should probably test it, but the DB/ODBC should kill you for trying to use naive timestamps or make them timezone aware to the local timezone (still a problem), as the column is defined as "timestamp with timezone"

Modules/database_manager.py
```diff
@@ -0,0 +1,386 @@
...
+                raise
+        else:
+            self.connection.close()
+            self._enabled = False
+
+    def _execute(self, *args, **kwargs):
```
theunkn0wn1: We can simplify this enabled check.

```py
if not self.enabled:
            log.warning("Tried to execute a DB call with the DB access being disabled.")
            raise RuntimeError("Check if the DB-module is enabled before executing code!"
                               " Currently it is NOT!")

# contents of else on line 169

This would save one layer of complexity

File: Modules/rat_command.py [hunk 23]
```diff
@@ -51,6 +52,12 @@ class NameCollisionException(CommandException):
 # character/s that must prefix a message for it to be parsed as a command.
 prefix = config['commands']['prefix']
 
+# Pydle bot instance.
+bot: BasicClient = None
```
theunkn0wn1: Ahem. This was removed in #66 intentionally.

Modules/rat_command.py
```diff
@@ -51,6 +52,12 @@ class NameCollisionException(CommandException):
...
+
+# facts manager instance
+facts_manager: FactsManager = FactsManager()
```
theunkn0wn1: Methinks this should be created in MechaClient?

Modules/rat_command.py
```diff
@@ -78,7 +85,18 @@ class NameCollisionException(CommandException):
                 log.debug(
                     f"Rule {getattr(command_fun, '__name__', '')} matching {ctx.words[0]} found.")
             else:
-                log.debug(f"Could not find command or rule for {prefix}{ctx.words[0]}.")
+                # might be a fact
```
theunkn0wn1: I question if this handling logic should occur in the facts manager instance, rather than in `rat_command`

This logic really isn't relevant to command handling, and is specific to facts handling therefore should occur inside the facts module

Modules/rat_facts.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Returns: True if it was a fact, False if not. Will also return False,
+            should the DBM be disabled.
+
+        """
+        if not self.dbm.enabled:
+            await context.reply("Sadly, Facts are currently not available. Please check back later.")
+            return False
+
+        regex = re.compile(r"!(?P<name>[a-zA-Z\d]+)-(?P<lang>[a-zA-Z]{2})|!(?P<name2>[a-zA-Z\d]+)")
```
theunkn0wn1: If for no other reason other than Me not knowing enough of ReGex to read that, could you please document this regex so us non-regex magicians can understand it? ((not blocking))

tests/conftest.py
```diff
@@ -238,6 +240,29 @@ def reset_rat_cache_fx(rat_cache_fx: RatCache):
...
+    """returns a FactsManager"""
+    return FactsManager()
+
+@pytest.fixture
+def dbm_fx():
+    """returns a DBM instance"""
+    return DatabaseManager()
+
+
+@pytest.fixture
```
theunkn0wn1: Duplicated fixture is Duplicated

tests/test_rat_facts.py
```diff
@@ -0,0 +1,148 @@
...
+        import tests.mock_bot
```
theunkn0wn1: use `bot_fx` rather than importing `mock_bot` directly.

Modules/rat_command.py
beepbeat: the code being there instead of in the facts module reduces (unneeded) nesting. It may be specific to facts, but it is also very specific to figuring out if the invoked "command" is indeed a fact or not.

Modules/rat_command.py
beepbeat: It doesn't really matter, as every constructor call will return the same instance #singleton

Modules/rat_command.py
theunkn0wn1: 👍

Modules/rat_command.py
theunkn0wn1: It unnecessarially introduces another instance of the facts manager, which will make testing harder.

As it is already provided as part of the `MechaClient` instance, it is provided in the `context` argument of all command executions. 

During testing we should only need to monkeypatch this object in one place, if at all.

by having it at the module level in rat_command as well as an instance-attribute of the Mechaclient, it creates two places where a patch may be necessary which makes testing potentially more difficult

tests/test_rat_facts.py
```diff
@@ -0,0 +1,148 @@
...
+        assert value
+
+        # kick the fact out of the DB
+        await facts_fx.delete_fact("test1", "de")
+
+        # and now the fact should be gone
+        value = await facts_fx.is_fact("test1", "de")
+        assert not value
+
+    async def test_rule(self, facts_fx, bot_fx):
```
theunkn0wn1:
```suggestion
    async def test_rule(self, facts_fx, bot_fx, async_callable_fx, monkeypatch):
```
Part one of my suggestion

tests/test_rat_facts.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # field to store the reply in
+        bot_fx.rat_facts_reply = ""
+
+        # mocking the Context class to better control what happens
+        class MockContext:
```
theunkn0wn1: This class is entirely unnecessary, a context_fx fixture can be used, or just a plain instance of `Context`.

tests/test_rat_facts.py
```diff
@@ -0,0 +1,148 @@
...
+            words = ("!go",)
+
+            # noinspection PyMethodParameters
+            def reply(msg: str):  # Note: no self here, works without
+                bot_fx.rat_facts_reply = msg
+
```
theunkn0wn1: we can create the context locally,
```suggestion
        context = await Context.from_message(bot_fx, "#unit_test", "some_recruit", "!go")
        monkeypatch.setattr(context, "reply", async_callable_fx)  # patch the reply function
```

tests/test_rat_facts.py
```diff
@@ -0,0 +1,148 @@
...
+        # noinspection PyTypeChecker
+        # shouldn't reply, as this is an "internal" fact
+        await facts_fx.handle_fact(MockContext)
+
+        # checking it did not reply, or replied with an empty string
+        assert bot_fx.rat_facts_reply == ""
```
theunkn0wn1: and assert our callable was not called.
```suggestion
        assert not async_callable_fx.was_called
```

Modules/rat_facts.py
beepbeat:
```suggestion
        regex = re.compile(r"!(?P<name>[a-zA-Z\d]+)-(?P<lang>[a-zA-Z]{2})|!(?P<name2>[a-zA-Z\d]+)")
        # Tries to split the inputted word (e.g !test-tr) into a group called "name" and a group called "lang" should this fail, for example because no language has been supplied, the it'll store the name of the fact in a group called "name2"
```

tests/test_rat_facts.py
```diff
@@ -0,0 +1,148 @@
...
+        class MockContext:
+            words = ("!go",)
+
+            # noinspection PyMethodParameters
+            def reply(msg: str):  # Note: no self here, works without
+                bot_fx.rat_facts_reply = msg
+
+        # noinspection PyTypeChecker
+        # shouldn't reply, as this is an "internal" fact
+        await facts_fx.handle_fact(MockContext)
```
theunkn0wn1: pass our actual context object to handle fact,
```suggestion
        await facts_fx.handle_fact(context)
```

Modules/rat_facts.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Responds in IRC with the message of the fact, should a match be found and not listed as an
+        internal fact.
+
+        Args:
+            context: the context as passed by rat_command
+
+        Returns: True if it was a fact, False if not. Will also return False,
+            should the DBM be disabled.
+
```
theunkn0wn1: Missing type check.

Modules/rat_facts.py
```diff
@@ -0,0 +1,209 @@
...
+            result = result[0]
+            return Fact(
+                result[0],
+                result[1],
+                result[2],
+                result[3],
+                timestamp
+            )
+        else:
+            return None
```
theunkn0wn1: [minor] no return implicitly returns `None`, else statement is redundant

config/config.template.json
```diff
@@ -65,5 +65,20 @@
...
+      "username": "mecha3",
+      "password": "mecha3"
+    },
+    "dsn": {
+      "name": "",
+      "username": "",
+      "password": ""
+    }
   }
 }
```
theunkn0wn1:
```suggestion
}

```
missing '\n' at EOL

requirements.txt
```diff
@@ -2,5 +2,9 @@
 pytest
 pytest-cov
 pytest-asyncio
+pytest-ordering
 pure-sasl
 coloredlogs
+pytest
```
theunkn0wn1:
```suggestion
```
redundant pytest requirement, its defined on line2

tests/test_database_manager.py
```diff
@@ -0,0 +1,179 @@
...
+# noinspection PyProtectedMember
+@pytest.mark.asyncio
+@pytest.mark.database
+@pytest.mark.usefixtures("prepare_dbm")
+class TestStuff(object):
+
+    def finalize(self, dbm_fx):
+        dbm_fx.enabled = False
+
+    async def test_enabled_fail(self, dbm_fx):
```
theunkn0wn1:
```suggestion
    @pytest.mark.parametrize("garbage", [None, 42, "nope.exe", sys')
    async def test_enabled_fail(self, dbm_fx, garbage):
```

be sure to import `sys` at top of file. the way your test is currently written only line 51 gets executed. Consider revising.

tests/test_database_manager.py
```diff
@@ -0,0 +1,179 @@
...
+        with pytest.raises(ValueError):
+            dbm_fx.enabled = None
+            dbm_fx.enabled = 42
+            dbm_fx.enabled = "nope.exe"
+            dbm_fx.enabled = dbm_fx
+            import sys
+            dbm_fx.enabled = sys
```
theunkn0wn1: nuke everything from line 51 -> 56 and replace it with `dbm_fx.enabled = garbage`

tests/test_database_manager.py
```diff
@@ -0,0 +1,179 @@
...
+        tmp = await dbm_fx.select_rows("testtabledelete", "AND", {"string1": "thest"})
+        assert len(tmp) == 0
+
+    async def test_double_dash_checks(self, dbm_fx: DatabaseManager):
+        with pytest.raises(ValueError):
+            await dbm_fx.delete_row("testtabledelete", "AND", {"test": "OR TRUE; -- "})
+            await dbm_fx.update_row("testtableupdate", "AND", {"string1": " OR FALSE; -- "})
+            await dbm_fx.insert_row("testtableinsert", {"string1": "stuff"}, ("string1",))
+            await dbm_fx.select_rows("testtableselect", "AND", {"test; -- ": "stuff"})
+
```
theunkn0wn1: only line 115 executes. presuming each of these calls is supposed to raise a valueerror.

tests/test_database_manager.py
```diff
@@ -0,0 +1,179 @@
...
+        # store the old cursor and connection
+        old_cursor = dbm_fx.cursor
+        old_conn = dbm_fx.connection
+
+        # this is a fake cursor
+        class I_am_a_duck():
+            def execute(self, *args, **kwargs):
+                raise pyodbc.Error("('08S01', '[08S01] SQLExecDirectW unable due to the connection"
+                                   " lost (35) (SQLExecDirectW)')")
```
theunkn0wn1: use a monkeypatch and an function definition here, rather than rolling a mock class

beepbeat: We had a discussion about this, which started as we couldn't make this work with Monkeys iirc.

theunkn0wn1: I am certain a monkeypatch will work here.

```py
async def duck_execute(*args, **kwargs)
    raise pyodbc.Error("('08S01', '[08S01] SQLExecDirectW unable due to the connection"
                                   " lost (35) (SQLExecDirectW)')")

monkeypatch.setattr(dbm_fx.cursor, "execute", duck_execute)
```

shatteredbeam:
Heyo.  Thought I'd give this some input as well.

```py
 def __init__(self):

        """
        Connects to the DB, sets up the connection, retrieves the cursor.
        Creates the default tables should they not exist.
        """

        # if we are not supposed to connect on start-up, we just sit tight and do nothing
        if not self.enabled:
            return
        # we init the connection, as we are supposed to do
        try:
            self.init_connection()
        except pyodbc.Error:
            # something went terribly wrong and excrements hit the ventilation device
            # using internal here to skip disconnection part as it might (and probably would) call
            # stuff on a NoneType
            self._enabled = False
            return
```
Does this really need to be under ``__init__`` ? Why are we placing this here instead of using a connect method?

```py
    def _execute(self, *args, **kwargs):
        if not self.enabled:
            log.warning("Tried to execute a DB call with the DB access being disabled.")
            raise RuntimeError("Check if the DB-module is enabled before executing code!"
                               " Currently it is NOT!")
        for retry in range(3):
            try:
                return self.cursor.execute(*args, **kwargs)
```

I have some concerns here about the use of the internal method.  it's just a pass through for cursor.execute, which is a potential security issue.  

```py
self.connection.setdecoding(pyodbc.SQL_WCHAR, encoding='utf-8')
        self.connection.setdecoding(pyodbc.SQL_CHAR, encoding='utf-8')
        self.connection.setencoding(encoding='utf-8')
```

UTF-8 compliance! My logging engine thanks you.

on line 227,  You're using the ``_execute`` function directly and just create a method for table creation, without any error catching.  It's a missed opportunity for logging.

on line 237, same.

I don't see ANY usage of logging at all, except in regards to the enable flag.*  We should have a log string sent when a database connection changes (connected, disconnected, refused, etc). Along with any error message returned.

Especially here, during init on line 260:
```py
        except pyodbc.Error:
            # something went terribly wrong and excrements hit the ventilation device
            # using internal here to skip disconnection part as it might (and probably would) call
            # stuff on a NoneType
            self._enabled = False
            return
```

There should be something other than a silent fail.  Tracking this down post-deployment would be time consuming.  I suggest a log entry such as 'Database Disabled due to' + ErrorType + Error Description.

```py
  async def drop_table(self, name: str) -> None:
```
Let's just remove the drop table method entirely.  I can't think of any situation in which a sanctioned table drop would be used from mecha.  It would all be done server side through proper channels, if a table drop was warranted.  That said, having this method is simply an attack surface that we don't need.

on line 370 (and other locations) I applaud your efforts at combating sql injection, but checking for a double dash alone isn't going to do anything, as it requires another whitespace character anyway to be detected as a newline.  I believe this needs to be done at the user input level, and the driver level, but not directly here.

In conclusion, I believe this methodology has some merit but it feels lacking.  It needs significantly more logging output for troubleshooting (logs are cheap, debugging is expensive) and a much greater acknowledgement to security.  If we should implement checking at this level, or higher in the stack isn't my call to make, but ANY transaction through the database should pass through before reaching your API.  In that way, you don't need to make ad-hoc checks for injection.

If you would like some assistance, I'm more than happy to help you with your PR, just let me know.

theunkn0wn1:
Thank you for all your hard work, i really appreciate the effort.

Unfortunately I cannot accept this pull request and I believe it is beyond repair. At several points critical issues in the design of this database module have been raised, and unfortunately i do not see how this PR can be repaired short of another rewrite. I have assigned @shatteredbeam to do this rewrite, no further effort should be required on your part.

Once again, thank you for contribution.

PR #59, opened 2018-07-16
SPARK-60: Implement color codes

nathanctech:
Implements the ability to use color codes in text as well as other formatting. Implements the `Colors` class in ratlib which has constants for all supported colors and other formatting. Also includes helper functions:

- `color()`
- `bold()`
- `italic()`
- `underline()`
- `reverse()`

`color()` accepts an additional color for the background color. This resolves SPARK-60.

### DIFF
--- a/utils/ratlib.py
+++ b/utils/ratlib.py
@@ -41,7 +41,7 @@ class Status(Enum):
     """The rescue is open, but is marked inactive"""
 
 
-class COLORS:
+class Colors:
     """
     Contains mIRC-style color codes (the standard)
     Reference: https://www.mirc.com/colors.html
@@ -62,6 +62,14 @@ class COLORS:
     PINK = '\x0313'
     GREY = '\x0314'
     LIGHT_GREY = '\x0315'
+
+    # codes for formatting
+    FORMAT_BOLD = '\x02'
+    FORMAT_UNDERLINE = '\x1F'
+    FORMAT_ITALIC = '\x1D'
+    FORMAT_COLOR = '\x03'
+    FORMAT_REVERSE = '\x16'
+    FORMAT_RESET = '\x0F'
 
 
 class Singleton(object):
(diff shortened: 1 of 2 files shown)


utils/ratlib.py
```diff
@@ -41,6 +41,37 @@ class Status(Enum):
     """The rescue is open, but is marked inactive"""
 
 
+class Colors:
```
MHajoha: Could you turn this into an enum? That would make things a tad more typesafe, you'd just have to grab the color codes with `color.value`.

theunkn0wn1: I would agree here, after reviwing enums and coming to terms with it >.>
Colours should be an enum.

theunkn0wn1:
@MHajoha we can rename/extract ratlib once we get the API module and the DBM merged in. Until then, we are kinda stuck with it as otherwise it would create probably create massive merge conflicts

utils/ratlib.py
```diff
@@ -41,6 +41,37 @@ class Status(Enum):
...
+    LIGHT_BLUE = '\x0312'
+    PINK = '\x0313'
+    GREY = '\x0314'
+    LIGHT_GREY = '\x0315'
+
+    # codes for formatting
+    FORMAT_BOLD = '\x02'
+    FORMAT_UNDERLINE = '\x1F'
+    FORMAT_ITALIC = '\x1D'
+    FORMAT_COLOR = '\x03'
```
theunkn0wn1: if you have the `FORMAT_COLOR` as a member of the enum, you do not also need to include it in the color enumerations.

I would suggest making `WHITE` thru `LIGHT_GREY` be ints rather than strings. would simplyify logic below.

utils/ratlib.py
```diff
@@ -128,3 +159,73 @@ def try_parse_uuid(suspect: str) -> UUID:
...
+    Args:
+        text: The text to colorize
+        text_color: a Colors.COLOR
+        bg_color: if specified, the background color
+
+    Returns:
+        str: colorized string
+    """
+    if bg_color is not None:
+        bg_color_code = bg_color.replace('\x03', '')
```
theunkn0wn1: This step would not be required if the colors were ints

utils/ratlib.py
```diff
@@ -128,3 +159,73 @@ def try_parse_uuid(suspect: str) -> UUID:
...
+        return f'{text_color},{bg_color_code}{text}{Colors.FORMAT_RESET}'
```
theunkn0wn1: you would need to preceed this with `FORMAT_COLOR`, but would still be cleaner IMHO

utils/ratlib.py
```diff
@@ -128,3 +159,73 @@ def try_parse_uuid(suspect: str) -> UUID:
...
+    """
+    Makes the text bold.
+    Args:
+        text: The text.
+
+    Returns:
+        str: the bolded text
+
+    """
+    return f'{Colors.FORMAT_BOLD}{text}{Colors.FORMAT_RESET}'
```
theunkn0wn1: resetting formatting here will have unintended consequences, such as if thi string were to then be wrapped in a color block the colour would be terminated prematurely. In other words, this method does not stack.

Demonstration:
```py
@command("debug-demo-clear")
async def cmd_dclear(context: Context):
    bold_str = bold("<bold>My boldened text</bold>")
    fstring = color(f"my beautyful {bold_str} text!", Colors.GREEN)
    await context.reply(fstring)
```
output: 
image
expected: the whole output string is in green.

Root cause: `colors.FORMAT_RESET` at the end of your formatting methods.

I believe the proper way to toggle a formatter is to repeat the format char without arguments.

for bold you would wrap text in 
```py
f"{Colors.FORMAT_BOLD}text{Colors.FORMAT_BOLD}"
```

utils/ratlib.py
theunkn0wn1: One more thing, i would suggest moving the bold/underline/italic/color/reverse/reset format keys to their own enum, as they are not technically colors

utils/ratlib.py
```diff
@@ -128,3 +165,75 @@ def try_parse_uuid(suspect: str) -> UUID:
 
     else:
         return result
+
+
+# color functions
+def color(text: str, text_color: Colors, bg_color=Optional[Colors]) -> str:
```
theunkn0wn1: I believe you will find that is not how `Optional` is supposed to be used.
In python, `Optional` is part of the type hint, not the value 😉 
`color`'s signature should look something like
```py
def color(text: str, text_color: Colors, bg_color:Optional[Colors] = None) -> str:
    ...
```

utils/ratlib.py
```diff
@@ -128,3 +165,75 @@ def try_parse_uuid(suspect: str) -> UUID:
...
+    """
+    Colorizes the given string with the specified color.
+    Args:
```
theunkn0wn1: Newline between function description and the next block, otherwise it will not be parsed correctly

utils/ratlib.py
```diff
@@ -41,6 +42,42 @@ class Status(Enum):
...
+    YELLOW = '08'
+    LIGHT_GREEN = '09'
+    CYAN = '10'
+    LIGHT_CYAN = '11'
+    LIGHT_BLUE = '12'
+    PINK = '13'
+    GREY = '14'
+    LIGHT_GREY = '15'
+    # this code needs to be suffixed to the colors above to actually display a color
+    FORMAT_COLOR = '\x03'
```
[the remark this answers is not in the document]
nathanctech: I only put it here so it's easily found, but that makes sense to me. I'll move it.

utils/ratlib.py
nathanctech: ...doh. Will fix.

PR #62, opened 2018-07-22
Spark 50: Event system

theunkn0wn1:
This PR implements an event system. (edited for consistency) 

SPARK-50 was intended to create a subscribe-able  callback for `on_message`, however during the development of that approach various limitations were reached. The first being it would require adding an instance attribute to the MechaClient, as well as multiple functions for the handling of registering subscribers and handling the callback event. Given the resources required to do this, and the lack of expandability of this approach, i determined a more general approach is required.
# the public interface (`event_system`)
`Modules.event_system` is the public interface for the events package, which exposes both the `Event` class and all pre-defined events.
## `NotImplemented` sentinel
this sentinel value can be returned by any event subscriber, however only renders effect during primary event emission where it will cause the not_handled subscribers to be invoked. If it is returned by a not_handled event subscriber it will be ignored 

# the `Event` Class
The event class both tracks events and their subscribers, as well as providing the machinery required to create new ones. 

## instance attributes
the Event class defines two instance attributes.
`Event.name` refers to the name of the event
`Event.subscribers` is a list of callable objects, the subscribers to this event instance.
`Event.not_handled_subscribers` list of callable objects, subscribers here are called when any normal subscriber returns `NotImplemented`

## public methods
`Event.emit` emits the event to all subscribers, `*args` and`**kwargs` are passed to subscribers.
`Event.__call__` convenience wrapper for `Event.emit`

## public decorators
`Event.subscribe` - subscribes the decorated function to the Event, returns the original function.
`Event.on_not_implemented`  - subscribe to the NotImplemented subevent

## Event Decloration
events can be declared in the following form:
```py
event = Event("my_rediculus_event_that_doesnt_deserve_a_base_function")
```
It is worth noting that instances of events are Singletons.
```py
>>> from Modules.event_system import Event
>>> foo = Event("Foo")
>>> bar = Event("Bar")
>>> snafu = Event("Foo")
>>> foo is bar
False
>>> foo is snafu
True
>>> snafu is not bar
True
```

## Subscription 
Functions can subscribe to receive callbacks from events via the `Event.subscribe(name)` decorator.
```py
@event.subscribe
async def subscribe_demo_coro(*args, **kwargs):
    ...
```
Functions can also subscribe, as needed, to the `on_not_Implemented` subevent.
```py
@event.on_not_implemented
async def potato(*args, **kwargs):
    ...
```
where `event` is an instance of `Event`.

the target object must be callable and must be an `async def` function, and its signature should match the form the event emits.

Subscribing functions get the same arguments as the event did upon its invocation, they work with the same information.

### class / instance method subscription
Due to technical limitations, namely being unable to get `self`/`cls` into the decorator, class and instance methods must be manually subscribed via a partial.

In your class's `__init__` function, you will need to generate a partial for your instance/class method, and invoke `Event.subscribe(partial)` directly.
Example, 
```py
class Potato(Singleton):
    def __init__(self):
        on_message.subscribe(partial(self.test))

    async def test(self, message, user, channel):
        log.debug(self)
        log.warning("potato test got called!")
```

TL;DR (but please do)
---------------------

- `Event` is a new class that defines a subscribe-able event
- `Event.subscribe` can be used to define subscribers of declared events
- events get emitted to subscribers via  `Event.emit` or via `event()`
- subscribers get the same arguments as those passed to `Event.emit`

Todo list:
---------------------

- [x] Remove `Event` decorator
- [x] make `@Event.subscribe` a instance method decorator
- [x] add event consumption/cancellation mechanism
- [x] package events
- [x] pre-define all pydle events (and our own) for general use
- [x] tests
- [x] documentation
- [x] integration with existing machinery


Modules/event.py
```diff
@@ -0,0 +1,157 @@
...
+    @classmethod
+    def _register(cls, event: 'Event'):
+        """
+        Registers a new event
+
+        Args:
+            event(Event): Event to register
+
+        Raises:
+            ValueError: attempted to register an event has already been registered
```
beepbeat: English is weird, but this is weirder.

pytest.ini
```diff
@@ -15,4 +15,5 @@ markers =
     user: marks a test as a user object test
     logging: marks a test as being a test for logging
     ratlib: marks a test as a ratlib test
-    setup_tests: setup test marker
\ No newline at end of file
+    setup_tests: setup test marker
+    event_system: Event system tests
```
beepbeat: !\n@EOF

Modules/event.py
```diff
@@ -0,0 +1,157 @@
...
+        log.debug(f"event {self.name} invoked...")
+        if self.decorated_coro is not None:
+            await self.decorated_coro(*args, **kwargs)
+        log.debug(f"emitting event to subscribers...")
+        await self._emit(*args, **kwargs)
+
+    def __get__(self, instance, owner):
+        # i honestly have no idea what black magic this is.
+        # https://stackoverflow.com/questions/30104047/how-can-i-decorate-an-instance-method-with-a-decorator-class
+        return partial(self.__call__, instance)
```
MHajoha: Let's have an example on how this 'black magic' works:
```python
class Bubble(object):
  def bath(self): ...

  @Event
  def gum(self): ...

my_bubble = Bubble()
```
When you call `Bubble.bath`, you have to provide the `self` argument yourself, as expected. You are calling a function like any other, `self` has no special significance here.

### Bound Methods ###
When, however, you access `my_bubble.bath`, you aren't getting a function. You are getting a **bound method**. The bound method behaves similarly to the function, with one key difference: The first argument is set to `my_bubble`. You can actually see what it was bound to by getting `my_bubble.bath.__self__`.
In essence, a method acts as follows:
```python
class MethodType(object):
  ...
  def __call__(self, *args, **kwargs):
    return self.__func__(self.__self__, *args, **kwargs)
  ...
```
In our example, `__self__` would be `my_bubble` and `__func__` would be `Bubble.bath`.

### The Problem ###
The problem arises for `Bubble.gum`.
Here, the function `gum` is replaced by an `Event` object, so the class `Bubble` now has an attribute `gum` of type `Event`. This means that no binding will take place. `Bubble.gum is my_bubble.gum` evaluates to True. The decorated function, however, is expecting a traditional `self` argument, i.e. the instance it is called on.

Your solution is to make `Event` a delegate. When `my_bubble.gum` is accessed, `my_bubble.gum.__get__` is called with `instance=my_bubble` and `owner=Bubble`. In `__get__`, you use `partial` to manually bind the first argument of `Event.__call__` to the instance `gum` was accessed on, i.e. `my_bubble` and returns the result. This means that `my_bubble.gum(7)` causes the underlying function to be called with `(my_bubble, 7)`.

### But why so complicated? ###
This was the happy path, and it works fine. Now, let's look at calling `Bubble.gum()` directly on the class. `__get__` will still be called, only with `instance=None`. This breaks the usual idea of being able to specify your own `self` argument, as it will be bound to None. You'll have to check for that and handle it.

Oh, and class as well as static methods won't work at all, just as a note.

This all strikes me as massively overcomplicated. Why should events be created by decorator at all? Why should it be called to trigger the event, that is simply misleading. Let me suggest three arguably simpler APIs:
```python
>>> bob = Event("bob")
... @bob.subscribe
... def bob1():
...   print("bob1")
... @Event("bob").subscribe
... def bob2():
...   print("bob2")
... bob.emit()
bob1
bob2
```
Or, without any classes:
```python
>>> @subscribe("bob")
... def bob1():
...   print("bob1")
... @subscribe("bob")
... def bob2():
...   print("bob2")
... emit("bob")
bob1
bob2
```
Or why allow arbitrary events at all, won't that only lead to typos? How about an enum?
```python
>>> @subscribe(Event.BOB)
... def bob1():
...   print("bob1")
... @subscribe(Event.BOB)
... def bob2():
...   print("https://youtu.be/FU-tuY0Z7nQ")
... emit(Event.BOB)  # Or Event.BOB.emit() ?
bob1
https://youtu.be/FU-tuY0Z7nQ
```

theunkn0wn1:
Ive added some additional functionality to this PR:

#### Instances of event are now callable (wrapper for `Event.emit`)
```py
>>> # setup stuff
>>> from Modules.event_system import Event
>>> from asyncio import get_event_loop
>>> loop = get_event_loop()
>>> 
>>> # define an event
>>> demo_event = Event("demo_event")
>>> 
>>> #define a subscriber
>>> @demo_event.subscribe
>>> async def foo(*args, **kwargs):
>>>     print("called and awaited!")
# call the event
>>> loop.run_until_complete(demo_event())
called and awaited!
[None]
```

#### Output of event invocation is now  a list
The output of `asyncio.gather` is a list of results, no reason to cast to a set here

#### Added `Event.subscribe_on_not_handled` decorator
Same basic functionality as `Event.subscribe`, but writes into a different instance-attribute.
This decorator allows coroutines to be called when any **conventional** subscriber returns `NOT_HANDLED`

theunkn0wn1:
@MHajoha i believe this is ready for review 😃

Modules/events/event.py
```diff
@@ -0,0 +1,181 @@
...
+    async def emit(self, *args: Any, **kwargs: Any) -> List[Any]:
+        """
+        Emit the event to all subscribers
+
+        Args:
+            *args (tuple): tuple of positional arguments to subscribers
+            **kwargs (dict): dict of keyword arguments to provide to subscribers
+
+        Returns:
+
```
MHajoha: Please add documentation on the method's return value ;)

Modules/events/event.py
```diff
@@ -0,0 +1,181 @@
...
+         >>> "my_other_event" in Event.events
+         True
+
+    Notes:
+        - Events will need to be invoked by something.
+
+        - Event subscribers must be async functions, and should be tested to work with
+            the event's signature (variable)
+
+        - exceptions raised by subscribers will not stop an event from being emitted
```
MHajoha: Wasn't the event already emitted when a subscriber has the chance to throw an exception? In that case, maybe 'will not stop other subscribers from being called' might be clearer.

Modules/events/event.py
```diff
@@ -0,0 +1,181 @@
...
+from functools import partial
+from typing import Callable, List, Dict, Any, Union
+
+# typedef
+Subscriptions = List[Callable]
+log = logging.getLogger(f"mecha.{__name__}")
+
+# If a subscriber returns this value, the event will canceled.
+# return this value to consume an event
+CANCEL_EVENT = object()
```
MHajoha: Could you make it clear what 'cancelling' an event means? I assume it prevents other subscribers from being called, but it isn't clear. Also, it would be great if this info were added to `Event.subscribe`.

Modules/events/event.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        if event.name in cls.events:
+            raise ValueError(f"name {event.name} is already registered as an event")
+        cls.events[event.name] = event
+
+    def subscribe(self, coro: Union[Callable, partial]) -> Callable:
+        """
+        Subscribe to this event
+
+        This method can be used as a decorator around **static methods**
```
MHajoha: Probably this decorator needs to be below the `@staticmethod` in this case. Might be worthwhile information to add.

Modules/mechaclient.py
```diff
@@ -0,0 +1,225 @@
+"""
+mechaclient.py - {summery}
+
+{long description}
```
MHajoha: {suggested changes}


;)

Modules/mechaclient.py
```diff
@@ -0,0 +1,225 @@
...
+            log.exception(ex_uuid)
+            error_message = graceful_errors.make_graceful(exc, ex_uuid)
+
+            # check to see if we are gracefully handling errors
+            if self.graceful_errors:
+                # and report it to the user
+                await self.message(channel, error_message)
+            else:
+                # we are not handling things gracefully, raise.
+                raise
```
MHajoha: grafik

Well, `exc`. But still.

Modules/events/event.py
```diff
@@ -0,0 +1,181 @@
...
+
+        Args:
+            event(Event): Event to register
+
+        Raises:
+            ValueError: attempted to register an event has already been registered
+        """
+        if event.name in cls.events:
+            raise ValueError(f"name {event.name} is already registered as an event")
+        cls.events[event.name] = event
```
MHajoha: Is `Event.events` used at all still?

theunkn0wn1:
```py

    @classmethod
    def _register(cls, event: 'Event'):
        """
        Registers a new event

        Args:
            event(Event): Event to register

        Raises:
            ValueError: attempted to register an event has already been registered
        """
        if event.name in cls.events:
            raise ValueError(f"name {event.name} is already registered as an event")
        cls.events[event.name] = event

```
yes. I believe i use this to keep track of the actually registered events...
now that i think about it this should be made private

Modules/mechaclient.py
theunkn0wn1: uhhh.... im not sure what you are telling me here. sowhat does `e` have to do with all this?

Modules/events/event.py
theunkn0wn1: sure thing! il add further documentation and revise what exists.

Modules/mechaclient.py
MHajoha: Sorry, I couldn't help myself. The `exc` is missing from after the `raise`.

theunkn0wn1: `raise` will implicitly raise the caught exception
i don't believe i need to explicitly tell python to `raise exc`

MHajoha: Urgh, so using `raise` without arguments re-raises the currently handled exception? I didn't know that. Could you do me a favour and add the `exc` anyhow? I might not be the only one bamboozled by this syntax.

theunkn0wn1:
```py
try:
    0/0
except ZeroDivisionError:
    print("someone divided by zero (Again...)")
    raise
Traceback (most recent call last):
someone divided by zero (Again...)
  File "J:\worspaces\pycharm\fuelrats\pipsqueak3\mecha_py_3_7\lib\site-packages\IPython\core\interactiveshell.py", line 2961, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-479d368a3351>", line 3, in <module>
    0/0
ZeroDivisionError: division by zero
```

MHajoha: I believe you, don't worry. Though I believe explicitly providing the exception to raise would prevent misunderstandings, like this one.

theunkn0wn1: hmm, i beg to differ. if i included an Exception class / object here it might be mis-read as me raising a *new* exception. or python would see it as a an exception during exception handling which would be worse on the stack trace.

By just having a bare `raise` it should be clear that i am raising the exception that was caught as im not raising anything else. IMHO

theunkn0wn1:
Im holding back on merging this til #77 is resolved, as this event system would be better implemented with contextvars

Eearslya:
With #77 closed as invalid, what is holding this back now?

nathanctech:
This PR is getting quite stale...any movement on this?

theunkn0wn1:
at this point this feature is entirely unnecessary.

As discovered somewhat recently Pydle has a working mechanism for adding new features without needing an entire extra module for it. for example in #97 the new commands and rule support mechanisms were added without needing an event subscription, instead `CommandSupport` added a new pydle callback `on_prefixed_message` which became part of the Mechaclient via featurization.


It took a little getting used to, but i believe this is a better approach than this Event system implementation at least for what this Event system was aimed to achieve.

If you guys still want this in mecha, versus using Featurization, gimme a squeak

PR #64, opened 2018-07-24
Use Fuelrats Pydle fork (temporary) - Enables python 3.7 support

theunkn0wn1:
While the source repo still has 3.7 support pending (as proposed) We can start developing against it


theunkn0wn1:
Official disapproval from the management

PR #65, opened 2018-09-13
SPARK-66: Clean up startup code and move it into a coroutine

MHajoha:
This PR makes the following changes:
* Rewriting startup to happen in a coroutine.
* Remove the use of `ClientPool`, instead using `client.connect(...)` directly.
Functionality should have stayed the same. In fact, apart from the minor changes in logging during startup, one shouldn't be able to tell that anything changed.

### DIFF
--- a/main.py
+++ b/main.py
@@ -12,11 +12,11 @@ See LICENSE.md
 This module is built on top of the Pydle system.
 
 """
-
+import asyncio
 import logging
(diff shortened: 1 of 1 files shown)


shatteredbeam:
Clientpool is only necessary if we are connecting more than once.

Looks good to me, too.

PR #66, opened 2018-09-19
Refactor/spark 67

theunkn0wn1:
This PR implements SPARK-67

# Main Effects:
- The creation of `Context` has been moved to `MechaClient.on_message`
`rat_command.trigger()` 's signature has been changed to `trigger(ctx: Context)`

- `rat_command.bot` has been nuked from orbit. We don't need it any more

## Secondary Effects
- `rat_command._split_message` was moved to `context._split_message`
This was done as to keep it to where it is used. it is not generic enough to warrant moving to our shared library file. 

- `Context` now has a new instance property, `prefixed`
This boolean flag is used to indicate the message started with the bots command prefix
This was necessary as not to change the established behavior of `_split_message` and `Context` respectively

- `Context` now has a new async classmethod: `Context.from_message`
This method is used to process a raw irc message, channel, and sender into a Context object
It needs to be async due to the WHOIS lookup call necessary for the `Context` constructor.

### Reasoning

The primary motivation for this was removing `rat_command.bot` module attribute, it was a ugly hack that can now be removed in favor of something cleaner.

Other events, such as `on_message_raw` can make use of the context variable, which is currently locked into `rat_command.trigger()` meaning they have to use the `MechaClient.on_message(target, sender, message)` signature and do their own message processing. 

By moving the Context creation into `on_message` the context can be used by any function it calls. This standardizes any event dependent on an IRC message by giving it its own class.

### DIFF
--- a/Modules/context.py
+++ b/Modules/context.py
@@ -139,7 +139,7 @@ class Context(object):
             Context
         """
         # check if the message has our prefix
-        prefixed = prefix in message
+        prefixed = message.startswith(prefix)
 
         # before removing it from the message
         message = message.lstrip(prefix)
--- a/Modules/context.py
+++ b/Modules/context.py
@@ -127,7 +127,8 @@ class Context(object):
     @classmethod
     async def from_message(cls, bot: 'MechaClient', channel: str, sender: str, message: str):
         """
-        Creates a context from a message
+        Creates a context from a IRC message
+
         Args:
             bot (MechaClient): MechaClient instance
             message (str):  raw message
--- a/Modules/context.py
+++ b/Modules/context.py
@@ -10,9 +10,15 @@ Licensed under the BSD 3-Clause License.
 
 See LICENSE.md
 """
-from typing import Optional
+from typing import Optional, Tuple, List, TYPE_CHECKING
 
 from Modules.user import User
+from config import config
+
+if TYPE_CHECKING:
+    from main import MechaClient
+
+prefix = config['commands']['prefix']
 
 
 class Context(object):
@@ -20,7 +26,13 @@ class Context(object):
     Command context, stores the context of a command's invocation
     """
 
-    def __init__(self, bot: 'MechaClient', user: User, target: str, words: [str], words_eol: [str]):
+    def __init__(self, bot: 'MechaClient',
+                 user: User,
+                 target: str,
+                 words: [str],
+                 words_eol: [str],
+                 prefixed: bool = False
+                 ):
         """
         Creates a new Commands Context
 
@@ -30,12 +42,30 @@ class Context(object):
             target(str): channel of invoking channel
             words ([str]): list of words from command invocation
             words_eol ([str]): list of words from command invocation to EOL
+            prefixed (bool): marker if the message is prefixed
         """
         self._user: User = user
         self._bot: 'MechaClient' = bot
         self._target: str = target
         self._words: [str] = words
         self._words_eol: [str] = words_eol
+        self._prefixed: bool = prefixed
+
+    @property
+    def prefixed(self):
+        """
+        Flag marking if the created context is a command/prefixed invocation
+        Returns:
+
+        """
+        return self._prefixed
+
+    @prefixed.setter
+    def prefixed(self, value: bool):
+        if not isinstance(value, bool):
+            raise ValueError
+
+        self._prefixed = value
 
     @property
     def user(self) -> User:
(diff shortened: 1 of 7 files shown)


Modules/context.py
```diff
@@ -94,6 +124,34 @@ def channel(self) -> Optional[str]:
...
+            bot (MechaClient): MechaClient instance
+            message (str):  raw message
+            sender (str): IRC nickname of the sender
+            channel (str): IRC channel the message was sent from
+
+        Returns:
+            Context
+        """
+        # check if the message has our prefix
+        prefixed = prefix in message
```
MHajoha: This will give false positives for messages such as `R@SIGNAL I'm stranded!!!`. In fact, it looks to me like any message including a `!` would be picked up as a command. `messase.startswith(prefix)` should be a better option.

Modules/rat_command.py
```diff
@@ -52,86 +51,46 @@ class NameCollisionException(CommandException):
...
-    :param message: triggers message to invoke
-    :param sender: author of triggering message
-    :param channel: channel of triggering message
-    :return: bool command
+
+    Args:
+        ctx (Context):
+
+    Returns:
+
```
MHajoha: *ahem*

Modules/rat_command.py
```diff
@@ -52,86 +51,46 @@ class NameCollisionException(CommandException):
...
-        return
-
-    words, words_eol = _split_message(message.lstrip(prefix))
-    if message.startswith(prefix):
-        if words[0].casefold() in _registered_commands.keys():
+
+    if ctx.words_eol[0] == "":
+        return  # empty message, bail out
+
+    if ctx.prefixed:
```
MHajoha: I know it doesn't change anything, but just for my peace of mind, could you change this to `elif`?

theunkn0wn1:
Ready for a final review methinks

main.py
```diff
@@ -144,7 +144,9 @@ def api_handler(self) -> object:
     """
     Initializes and connects the client, then passes it to rat_command.
     """
-    client_args = {"nickname": config["irc"]["nickname"]}
+    client_args = {"nickname": config["irc"]["nickname"],
+                   # "loop":loop
```
MHajoha: If it's not needed any longer, could you remove this comment?

main.py
```diff
@@ -92,12 +96,8 @@ def __init__(self, *args, **kwargs):
             sanitized_message = sanitize(message)
             log.debug(f"Sanitized {sanitized_message}, Original: {message}")
             try:
-                await rat_command.trigger(message=sanitized_message,
-                                          sender=user,
-                                          channel=channel)
-            except CommandNotFoundException:
-                # command was not found
-                log.debug("Command invoked from \"{message}\" not found.")
+                ctx = await Context.from_message(self, channel, user, message)
```
MHajoha: Didn't you already make the context instance at line 87?

main.py
```diff
@@ -163,9 +165,9 @@ def api_handler(self) -> object:
     client = MechaClient(**client_args)
     await client.connect(hostname=config['irc']['server'],
                          port=config['irc']['port'],
-                         tls=config['irc']['tls'])
+                         tls=config['irc']['tls'],
+                         )
```
MHajoha: Stray comma & newline. Does Python not complain when it gets here?

tests/test_permissions.py
```diff
@@ -25,17 +25,29 @@
...
     Commands._flush()
     Commands.bot = bot_fx
+
+
+@pytest.fixture
+def restricted_command_fx(async_callable_fx, Setup_fx):
+    # ugly hack to utilize async_callable_fx is ugly
+    @require_permission(permissions.OVERSEER)
+    async def restricted(context):
+        await async_callable_fx(context)
```
MHajoha: You should be able to do `restricted = require_permission(permissions.OVERSEER)(async_callable_fx)`.

tests/test_permissions.py
```diff
@@ -25,17 +25,29 @@
...
 
 # registration is done in setUp
-@require_permission(permissions.OVERSEER)
-async def restricted(context):
-    await context.reply("Restricted command was executed.")
+
+
+# @require_permission(permissions.OVERSEER)
+# async def restricted(context):
+#     await context.reply("Restricted command was executed.")
```
MHajoha: Is this likely to be needed again soon? Commented-out code can get confusing quickly imo.

main.py
theunkn0wn1: No, its legal. It also translates to less of a headache if we ever want to add an additonal kwarg here, as the comma is already in place. functionally this makes no change (I think it found its way there while debugging something)

tests/test_permissions.py
theunkn0wn1: Il have a look. good point.

tests/test_permissions.py
theunkn0wn1: didn't think of that. 👍

main.py
theunkn0wn1: Yeah, but it belongs here. fixed

PR #67, opened 2018-10-03
Make the MockBot a subclass of the MechaClient

theunkn0wn1:
This PR makes the `MockBot` class a subclass of the `MechaClient`.

The primary motivation here is to enable the usage of the mock bot's overriden methods while calling methods on the MechaClient instance. For instance, in SPARK-50 i need to call Mechaclient methods to verify their function, which in turn call methods overriden in MockBot. For these tests, i am testing the implementation of the Mechaclient method, not the pydle back end. Therefore being able to easily mock out the `whois` method is of value.

Since the MockBot is provided as a facility to mock certain Mechaclient calls for the sake of testing, it makes sense to have it simply extend and override the corresponding methods in the Mechaclient.

This change does not break existing tests, nor requires any rework of them.

This change facilitates testing the MechaClient

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,10 +19,9 @@ from uuid import uuid4, UUID
 
 import pytest
 
-# Set argv to keep cli arguments meant for pytest from polluting our things
-
-
 from Modules.rat_cache import RatCache
+
+# Set argv to keep cli arguments meant for pytest from polluting our things
 
 sys.argv = ["test",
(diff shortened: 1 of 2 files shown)


theunkn0wn1:
Points addressed, i have broken `connect()`, it now raises a `RuntimeWarning`.
I added the missing call to the super.
I don't think we need to override the `connected` property, i think we can revisit it later if it becomes necessary

PR #68, opened 2018-10-10
SPARK-69 & SPARK-72: Allow CallableMock to mimic an existing callable's signature and to raise a previously set exception

MHajoha:
# SPARK-69

`CallableMock` now take an optional constructor parameter `function_to_mimic`. Nothing will change for instances without this parameter set.

If it is set, however, the `_bound_calls` field will hold a list of `inspect.BoundArguments` instances, one for each call. These can be accessed via the property `CallableMock.bound_calls` and is also considered by `was_called_with`.

From the `CallableMock` docstring:
```rest
    Additionally, a :class:`CallableMock` can be made to mimic an existing function or method, such
    that args and kwargs can be matched to the arguments that the function expects. This also works
    with many other callables including bound methods, but may fail for certain built-in functions
    and methods.

    Note that calling a mock mimicking a function will raise a :exc:`TypeError` if the provided
    arguments don't fit the method. The same restriction also applies to the `was_called_with`
    method on those mock instances.

    >>> def target_function(my_arg): ...
    >>> fun = CallableMock(target_function)
    >>> fun("banana")
    >>> fun.was_called_with("banana")
    True
    >>> fun.was_called_with(my_arg="banana")
    True

    This is not possible when that function is not provided, as args and kwargs will be kept
    separate:

    >>> fun = CallableMock()
    >>> fun("banana")
    >>> fun.was_called_with("banana")
    True
    >>> fun.was_called_with(my_arg="banana")
    False
```

`_Call` has also been corrected to `Call`, as its instances are passed outside of the module.

# SPARK-72

It is now possible to specify an exception to be raised when the callable is called. This is done by setting the `exception_to_raise` field, which expects an exception instance. If it is set, `return_value` will be ignored and the exception instance will be thrown instead.

### DIFF
--- a/tests/mock_callables.py
+++ b/tests/mock_callables.py
@@ -57,6 +57,20 @@ class CallableMock(object):
     >>> fun.was_called_with(InstanceOf(int), 2, 3, her=InstanceOf(str))
     True
 
+    Either a return value or an exception to raise during a call can be set, using the fields
+    `return_value` and `exception_to_raise`, respectively. In case both are set, the exception will
+    be raised.
+
+    >>> fun = CallableMock()
+    >>> fun.return_value = 42
+    >>> fun()
+    42
+    >>> fun.exception_to_raise = ValueError("meaning of life is incomputable")
+    >>> fun()
+    Traceback (most recent call last):
+      ...
+    ValueError: meaning of life is incomputable
+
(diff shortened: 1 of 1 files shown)


theunkn0wn1:
Do these bindings also work for async functions?

MHajoha:
I don't see why they shouldn't. Async functions aren't much different from regular ones. Though I haven't specifically tested it.

PR #70, opened 2018-10-28
SPARK-71 Fixed faulty command prefix processing in Context

theunkn0wn1:
This PR resolves SPARK-71 by correcting Context's prefix processing.

Context will now only strip exactly the command prefix from the start of the message leaving any subsequent characters (such as the command-prefix being used improperly)

This PR also adds a test module for regression tests, which can be used to house further regression tests as more bugs are detected and fixed.
Further, I have added a pytest marker for regression tests, so they can be run independently from the rest of the test suite.

End result: if our command prefix is `!`, commands such as `!bang` will now behave as expected, 
```
<unknown> !!bang
<Mechasqueak3-tests[BOT]> boom!
<unknown> good.
```

### DIFF
--- a/Modules/context.py
+++ b/Modules/context.py
@@ -134,8 +134,9 @@ class Context(object):
         # check if the message has our prefix
         prefixed = message.startswith(prefix)
 
-        # before removing it from the message
-        message = message.lstrip(prefix)
+        if prefixed:
+            # before removing it from the message
+            message = message[len(prefix):]
 
         # build the words and words_eol lists
         words, words_eol = _split_message(message)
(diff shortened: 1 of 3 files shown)


theunkn0wn1:
Im just gunna note that this is an edge case, and would only occur if one of us developers do something daft and tried to register a command like this.

That being said, Mecha makes no effort to prevent this kind of behavior nor does it behave properly when someone does something like this. This PR corrects mecha's behavior so, in the event we want to do something like this, its possible. ((perhaps we can abuse this for debug /super-special commands?))

I Discovered this bug while experimenting with the Test Fuzzing library [Hypothesis](https://hypothesis.works)

PR #71, opened 2018-11-03
Remove whois usage from mecha3

theunkn0wn1:
This PR removes WHOIS queries from mecha3.

Basically mecha3 can get all the data it needs about a user from what is already provided by the Server
And, by extension, Pydle. Further, pydle already has its own builtin User database we can query into to obtain relevant data.

The one downside to using Pydle's database *without* ever doing a whois query against the server is certain
fields are never initalized and remain as default. these fields are `account` and `identified`. 

Nothing much can be done about this, besides doing a whois query at some point should it become necessary. 😕

A user can be assumed to be identified and valid if they have a valid vhost, as it is not possible to obtain a fuelrats.com vhost without going through nickserv. Therefore we can throw out the identified check entirely, which negates the point of using whois queries in the current implementation

### DIFF
--- a/Modules/user.py
+++ b/Modules/user.py
@@ -150,7 +150,7 @@ class User(object):
         data = bot.users.get(nickname.casefold(), None)
 
         # if we got a object back
-        
+
         if data:
             return cls(**data)
 


Modules/user.py
```diff
@@ -174,47 +134,25 @@ def nickname(self) -> str:
...
-                       secure=data['secure'],
-                       account=data['account'],
-                       nickname=nickname,
-                       username=data['username'],
-                       hostname=data['hostname'],
-                       realname=data['realname'],
-                       server=data['server'],
-                       server_info=data['server_info'])
+        # if we got a object back
+        # ((inb4 falsy check comes back to bite))
```
MHajoha: I feel like the second one of these comments shouldn't be here ;)

Modules/user.py
```diff
@@ -174,47 +134,25 @@ def nickname(self) -> str:
...
+        if data:
+            return cls(**data)
```
MHajoha: It seems dangerous to me to throw the contents of the user dict into `User`'s constructor without testing first that at least the required arguments are present.

theunkn0wn1: if the required arguments are not present, the constructor will fail. if invalid arguments are provided the constructor will fail. "It is better to seek forgiveness than ask permission"

Modules/user.py
theunkn0wn1: methinks its fine where it is, in (( out of context notation )) 😉

theunkn0wn1: im not seriously concerned, and im actually checking against a Empty dictionary or None. I just have a bit of a inherent distrust towards falsy checks 😅

Also, `(())` notation, while used in our drills, actually originates from elsewhere.


Anyways, if it bothers you this much il just remove the comment
```suggestion
        
```

PR #73, opened 2018-12-04
[Testing/bugfix]Manually specify our testing directory

theunkn0wn1:
Apparently Pydle's test suite gets detected by PyTest in some configurations, which leads to bad things.
I added a line to the `pytest.ini` configuration file as to manually specify our `tests/` directory as the test discovery root. This prevents pytest from trying to run dependencies' test suite(s).


shatteredbeam:
This is fine.  No objections.

PR #77, opened 2018-12-23
Rise of the ContextVar, fall of `Context`

theunkn0wn1:
This PR removes the `Modules.Context` dataclass in its entirety, it is nolonger necessary.

Rationale
---------------------------
Our present `Context` object is designed to store information relevant to `MechaClient.on_message()` Invocations, which it does in an acceptable fashion. It stores the bot's instance, the words of the message, and the sender as a `User` object. However, it is **horribly inflexible**.

For instance, if we want to expand the `Context` class to work for other event types, or store other data in the execution context it would require making global changes. Further, it is inefficient passing around a Context object into every routine involved in the handling of the event. Especially since only a small portion of the functions in the callchain actually /do/ anything with the Context object. Those that do generally don't use more than one or possibly two attributes in the Context object. 

Thus, passing around a `Context` object is bad for memory-management as it requires passing a useless object around, it increases cognative complexity (unused argument) and is inflexible (it is purpose-built with only one of many events in mind). **We would be better served with a more flexible solution**

Proposal
--------------------

Therefore, I propose introducing `ContextVar`s.
These were introduced in python 3.7, and provide async-safe thread-local storage. You can read up on them [here](https://www.python.org/dev/peps/pep-0567/) and [here](https://docs.python.org/3/library/contextvars.html)

In essence, `ContextVar`s provide us with the means of defining our context-sensitive variables in one place that can be imported where needed. This has the benefit of not needing to pass around the objects unnecessarily, as the functions only need to import what they need.

 Further, these objects are fully compatible with the asyncio event loop, and provides builtin isolation between asynchronous tasks. This means the bot can continue handing multiple concurrent irc messages without these variables becoming corrupted by one another

Finally, we can define new `ContextVar`s without needing to write additional tests, nor needing to change existing code except for the bits that will make use of the added var. This makes the approach far more expandable and easier to change should our future needs demand.

Implementation
------------------------

file `Modules.context` has been moved to `Modules.context._context`
- this was done so I could turn `Modules.context` into a package, to simply the public interface
`Context.from_message` has been moved to the module-scope
- this function has also been adapted to make use of the new `ContextVar`s
class `Modules.context.Context` has been removed, it is not necessary
Several new Context Variables have been introduced, they are defined in file `Modules.context._context` and exposed as part of the public interface `Modules.context`. See below.
## the new vars
`target`:`str` the target of a message
`channel`:`str` the target of a message, if it was a channel
`user`:`User` the User that invoked the message
`words`:`List[str]` the list of words for a `on_message` invocation
`words_eol`:`List[str]` list of words to eol
`prefixed`:`bool` whether or not a `on_message` invocation possesses the command prefix.
`message`:`str` the raw message of a `on_message` invocation
`bot`:`Mechaclient` the `MechaClient` instance.
`sender` :`str` the raw sender of a message.

Notes
------------------
This PR requires python 3.7 to work, so this Pr also has the side effect of raising the version bar. There is another Pr in the work that also makes use of 3.7 features so this isn't a problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
     docker:
       # specify the version you desire here
       # use `-browsers` prefix for selenium tests, e.g. `3.6.1-browsers`
-      - image: circleci/python:3.6.6
+      - image: circleci/python:3.7.1
 
       # Specify service dependencies here if necessary
       # CircleCI maintains a library of pre-built images
--- a/Modules/context.py
+++ b/Modules/context.py
@@ -127,34 +127,32 @@ class Context(object):
         """
         return self.target if self.bot.is_channel(self.target) else None
 
-    @classmethod
-    async def from_message(cls):
-        """
-        Creates a context from a IRC message
 
-        Args:
+def from_message():
+    """
+    Populates the context from a IRC message
+    """
+    _message = message.get()
+    _bot = bot.get()
+    # check if the message has our prefix
+    _prefixed = _message.startswith(prefix)
 
-        Returns:
-            Context
-        """
-        # check if the message has our prefix
-        prefixed.set(message.get().startswith(prefix))
+    if _prefixed:
+        prefixed.set(_message.startswith(prefix))
+        # before removing it from the message
+        _message = _message[len(prefix):]
+        message.set(_message)
 
-        if prefixed.get():
-            # before removing it from the message
-            message.set(message.get()[len(prefix):])
+    # build the words and words_eol lists
+    _words, _words_eol = _split_message(_message)
+    words.set(_words)
+    words_eol.set(_words_eol)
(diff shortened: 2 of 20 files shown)


theunkn0wn1:
sigh the more I think about it the less sense this makes.
Why do we need to use thread-local storage? 
We don't really.
Whats wrong with passing an actual object about?
Expandability, but i suppose thats a problem we need a more appropriate solution for.
Why are these globally accessible?
by spec.
should they be?
Not really.

What problem does this solve?
Expandability
What problems does this create?
 - Added complexity of using thread-local storage
 - potentially a whole new class of bugs
 - scoping and encapsulation issues

Methinks this is more a solution looking for a problem, than a solution solving a problem.

I suppose we will need to figure something better out for the dozen or so different contexts we have.

PR #78, opened 2018-12-24
Feature/database manager

shatteredbeam:
New package: Database Manager.

A singleton class using psycopg2 to query postgreSQL.  Initial tests may fail on this PR due to requirement of an online database for testing, and this is expected.

Simply instantiate the DatabaseManager class, and it will create a connection pool to the specified server, either with arguments passed during creation or by pulling configuration information from file and send your requests via DBM.query(query, values), returning a list of tuples (one row per tuple).

The query value is required to be a composed sql.SQL object, which is part of the psycopg2 library.  It does *not* accept string queries for security reasons.  All queries must be pre-composed and passed, with appropriate values, to the query function.  This is the main mitigation to prevent injection/substitution issues.  An example of this format is in commands\debug.py as a basic fact search.

This package is intended to be a requirement for the upcoming Fact Manager, but can be used generally as well.

'Smoke Testing' has been performed up to 50 concurrent queries, all handled gracefully, mostly in thanks to allowing the psycopg2 library to handle connection pooling, and being able to create/destroy cursors for each query as needed.

Future plans for this module include removing dependency on the Singleton class, and adding support to be used as a context manager.

Please note, this is NOT based on or derivative of any previous database-related pull requests.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,7 @@
 
 [run]
 branch = False
-source = Modules,ratlib
+source = Modules,ratlib,database
 
 [report]
 exclude_lines =
--- a/Pipfile
+++ b/Pipfile
@@ -4,9 +4,10 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-pydle = {editable = true, ref = "asyncio", git = "git://github.com/Shizmob/pydle.git"}
+pydle = {editable = true,ref = "asyncio",git = "git://github.com/Shizmob/pydle.git"}
 pure-sasl = "*"
 coloredlogs = "*"
+psycopg2 = "*"
 
 [dev-packages]
 pytest = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-pydle = {editable = true,ref = "asyncio",git = "git://github.com/Shizmob/pydle.git"}
+pydle = ">=0.8.5"
 pure-sasl = "*"
 coloredlogs = "*"
 psycopg2 = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -15,4 +15,4 @@ pytest-cov = "*"
 pytest-asyncio = "*"
 
 [requires]
-python_version = "3.6"
+python_version = "3.7"
--- a/commands/debug.py
+++ b/commands/debug.py
@@ -16,6 +16,8 @@ import logging
 from Modules.context import Context
 from Modules.permissions import require_permission, TECHRAT, require_channel
 from Modules.rat_command import command
+from database import DatabaseManager
+from psycopg2 import sql
 
 log = logging.getLogger(f"mecha.{__name__}")
 
@@ -39,3 +41,15 @@ async def cmd_debug_whois(context):
 @require_permission(TECHRAT)
 async def cmd_superping(context: Context):
     await context.reply("pong!")
+
+
+@command("debug-fact")
+@require_permission(TECHRAT)
+async def cmd_query(context: Context):
+    # assuming `database` is a instance of the DBM
+    database = DatabaseManager()
+    sql_query = sql.SQL("SELECT message from fact WHERE name=%s AND lang=%s")
+    sql_values = (context.words[1], context.words[2])
+    data = await database.query(sql_query, sql_values)
+    for item in data:
+        await context.reply(str(item))
--- a/config/testing.json
+++ b/config/testing.json
@@ -23,6 +23,13 @@
     "base_logger": "testmecha",
     "log_file": "logs/unit_tests.log"
   },
+  "database": {
+    "host": "localhost",
+    "port": 5432,
+    "dbname": "mecha3",
+    "username": "mecha3",
+    "password": "mecha3"
+  },
   "commands": {
     "prefix": "!"
   },
--- a/database/__init__.py
+++ b/database/__init__.py
@@ -1,0 +1,13 @@
+"""
+__init__.py
+
+Copyright (c) 2018 The Fuel Rat Mischief,
+All rights reserved.
+
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+"""
+__all__ = ["database_manager"]
+
+from .database_manager import DatabaseManager
--- a/database/__init__.py
+++ b/database/__init__.py
@@ -10,5 +10,3 @@ See LICENSE.md
 """
 from .database_manager import DatabaseManager
 __all__ = ["DatabaseManager"]
-
-
--- a/database/__init__.py
+++ b/database/__init__.py
@@ -8,6 +8,6 @@ Licensed under the BSD 3-Clause License.
 
 See LICENSE.md
 """
-__all__ = ["database_manager"]
+__all__ = ["DatabaseManager"]
 
 from .database_manager import DatabaseManager
--- a/database/__init__.py
+++ b/database/__init__.py
@@ -8,6 +8,7 @@ Licensed under the BSD 3-Clause License.
 
 See LICENSE.md
 """
+from .database_manager import DatabaseManager
 __all__ = ["DatabaseManager"]
 
-from .database_manager import DatabaseManager
+
--- a/database/database_manager.py
+++ b/database/database_manager.py
@@ -59,8 +59,9 @@ class DatabaseManager(Singleton, object):
 
     async def query(self, query: sql.SQL, values: tuple) -> list:
         """
-        Send a query to the connected database.  Pulls a connection from the pool and creates a cursor,
-        executing the composed query with the values.  Requires a composed SQL object (See psycopg2 docs)
+        Send a query to the connected database.  Pulls a connection from the pool and creates
+        a cursor, executing the composed query with the values.
+        Requires a composed SQL object (See psycopg2 docs)
         Args:
             query: composed SQL query object
             values: tuple of values for query
@@ -77,7 +78,8 @@ class DatabaseManager(Singleton, object):
 
         # Pull a connection from the pool, and create a cursor from it.
         with self._dbpool.getconn() as connection:
-            # If we could set these at connection time, we would, but they must be set outside the pool.
+            # If we could set these at connection time, we would,
+            # but they must be set outside the pool.
             connection.autocommit = True
             connection.set_client_encoding('utf-8')
             # Create cursor, and execute the query.
--- a/database/database_manager.py
+++ b/database/database_manager.py
@@ -62,6 +62,7 @@ class DatabaseManager(object):
         Send a query to the connected database.  Pulls a connection from the pool and creates
         a cursor, executing the composed query with the values.
         Requires a composed SQL object (See psycopg2 docs)
+
         Args:
             query: composed SQL query object
             values: tuple of values for query
--- a/database/database_manager.py
+++ b/database/database_manager.py
@@ -28,7 +28,7 @@ class DatabaseManager(object):
             self._initialized = True
 
             # Utilize function arguments if they are provided,
-            # otherwise retrieve from config.json and use those values.
+            # otherwise retrieve from config file and use those values.
             self._dbhost = dbhost if dbhost is not None else config['database'].get('host')
             assert self._dbhost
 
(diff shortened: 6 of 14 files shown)


.coveragerc
```diff
@@ -1,7 +1,7 @@
 
 [run]
 branch = False
-source = Modules,ratlib
+source = Modules,ratlib,database
```
Eearslya: Should this be a top-level folder, or a subfolder of `Modules`?

tests/test_database_manager.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    assert test_dbm._initialized
+    assert test_dbm._dbhost
+    assert test_dbm._dbpass
+    assert test_dbm._dbname
+    assert test_dbm._dbport
+    assert test_dbm._dbuser
+
+
+def test_private_settings(test_dbm):
```
Eearslya: Empty test?

tests/test_database_manager.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    Assign values to
+    """
+
+
+def test_db_connection(test_dbm_pool):
```
Eearslya: Is this testing our code, or testing `pyscopg`'s code?
Ditto for `test_db_connection_close`, `test_db_connection_autocommit`, and `test_db_connection_client_encoding`.

.coveragerc
shatteredbeam: It's a package, not part of the Modules package, so it has its own top folder.

`from database import Databasemanager`
vs
`from Modules.database import DatabaseManager`

tests/test_database_manager.py
shatteredbeam: Good catch.  I'll remove it.

tests/test_database_manager.py
shatteredbeam: I guess the scope has a little bit of overshadowing, but I felt it necessary to test the DBM's implementation and assert that these are actually valid, and being used.  When it comes to using live/production database I would like these to cause an error if they leave the library or become depreciated.

tests/test_database_manager.py
shatteredbeam: Removed in commit c876eb96d4bda6b308d55d5a6e50e7c73445a21a

.coveragerc
theunkn0wn1: Methinks this should be a sub-package of the `Modules` package. If nothing less its for consistency with the rest of the implementation.

Pipfile
```diff
@@ -4,14 +4,15 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-pydle = {editable = true, ref = "asyncio", git = "git://github.com/Shizmob/pydle.git"}
+pydle = {editable = true,ref = "asyncio",git = "git://github.com/Shizmob/pydle.git"}
```
theunkn0wn1: we can adjust this now that Pydle has a feature update.

```suggestion
pydle = ">=0.8.5"
```

commands/debug.py
```diff
@@ -39,3 +41,15 @@
...
+@command("debug-fact")
+@require_permission(TECHRAT)
+async def cmd_query(context: Context):
+    # assuming `database` is a instance of the DBM
+    database = DatabaseManager()
+    sql_query = sql.SQL("SELECT message from fact WHERE name=%s AND lang=%s")
+    sql_values = (context.words[1], context.words[2])
+    data = await database.query(sql_query, sql_values)
+    for item in data:
+        await context.reply(str(item))
```
theunkn0wn1: I believe we agreed a command of this form would not reach production.

config/testing.json
```diff
@@ -23,6 +23,13 @@
     "base_logger": "testmecha",
     "log_file": "logs/unit_tests.log"
   },
+  "database": {
+    "host": "localhost",
+    "port": 5432,
+    "dbname": "circle_test",
+    "username": "root",
+    "password": "mecha"
+  },
```
theunkn0wn1: please also append this region to the template config file, don't want that getting out of date.

database/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+Copyright (c) 2018 The Fuel Rat Mischief,
+All rights reserved.
+
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+"""
+__all__ = ["DatabaseManager"]
+
+from .database_manager import DatabaseManager
```
theunkn0wn1: import statement should be top-level.

database/database_manager.py
```diff
@@ -0,0 +1,93 @@
...
+from config import config
+from psycopg2 import sql, pool
+from utils.ratlib import Singleton
+import logging
+import psycopg2
+
+log = logging.getLogger(f"mecha.{__name__}")
+
+
+class DatabaseManager(Singleton, object):
```
theunkn0wn1: After long discussion, i don't believe inheriting from `Singleton` is a good idea.

database/database_manager.py
```diff
@@ -0,0 +1,93 @@
...
+            self._dbpool = psycopg2.pool.SimpleConnectionPool(5, 10, host=self._dbhost,
+                                                              port=self._dbport,
+                                                              dbname=self._dbname,
+                                                              user=self._dbuser,
+                                                              password=self._dbpass)
+
+            if self._dbpool:
+                log.info("SQL Database Connected.")
+        except (Exception, psycopg2.DatabaseError) as error:
+            log.warning("Error connecting to SQL database.", error)
```
theunkn0wn1: fatal errors should not be suppressed. Responsibility for handling a failed connection should be differed to the caller.

database/database_manager.py
```diff
@@ -0,0 +1,93 @@
...
+            connection.set_client_encoding('utf-8')
+            # Create cursor, and execute the query.
+            with connection.cursor() as cursor:
+                cursor.execute(query, values)
+                result = cursor.fetchall()
+
+        # Release connection back to the pool.
+        self._dbpool.putconn(connection)
+
+        return list(result)
```
theunkn0wn1: What type is `result` and why are we casting it to a list?

tests/test_database_manager.py
```diff
@@ -0,0 +1,140 @@
...
+
+@pytest.fixture
+def test_dbm() -> DatabaseManager:
+    """
+    Test fixture for Database Manager.
+
+    A DATABASE CONFIGURATION AND CONNECTION IS REQUIRED FOR THESE TESTS.
+    """
+    database = DatabaseManager()
+    return database
```
theunkn0wn1: Shouldn't the fixture also handle cleaning up the DBM instance during teardown?
Should the DBM be function scoped or test-session scoped?
Shouldn't this fixture be in `conftest.py`?

Fixtures should follow established naming convention, namely being suffixed with `_fx` to signify they are actually fixtures.

tests/test_database_manager.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    return test_dbm._dbpool
+
+
+def test_dbm_singleton():
+    """
+    Verify that DBM is indeed a singleton and not spawning a new instance.
+    """
+    class Pirate(DatabaseManager):
+        pass
```
theunkn0wn1:
```suggestion
        ...
```
Nitpick

tests/test_database_manager.py
theunkn0wn1: Methinks i see where @Eearslya is coming from here, are we testing our code here or the psql library?
The behavior of dependencies alone should not be tested, unless we intend to run their test suite as well. It is beyond the scope of our testing

tests/test_ratlib.py
```diff
@@ -24,7 +24,7 @@
      'Um die Rat zu deiner Freundesliste hinzuzufügen drücke'),
     ('которые отдают топливо, нажмите здесь: http://t.fuelr.at/kgbfoamru',
      'которые отдают топливо, нажмите здесь: http://t.fuelr.at/kgbfoamru'),
-    ('a;A;B;b;C;c;1\'2\'3', 'aABbCc123'),
+    ('a;A;B;b;C;c;123', 'aABbCc123'),
```
theunkn0wn1: ??? what does this fix?

utils/ratlib.py
```diff
@@ -18,7 +18,7 @@
 
 MIRC_CONTROL_CODES = ["\x0F", "\x16", "\x1D", "\x1F", "\x02",
                       "\x03([1-9][0-6]?)?,?([1-9][0-6]?)?"]
-STRIPPED_CHARS = ';\''
+STRIPPED_CHARS = ';'
```
theunkn0wn1: ???

database/database_manager.py
theunkn0wn1: Its worth noting that this line lead to downstream errors during testing where the database fails to connect. Granted in this situation any subsequent calls fail but they do so for the wrong reason.
```
test setup failure
test_dbm = <database.database_manager.DatabaseManager object at 0x7fb37a42ef60>

    @pytest.fixture()
    def test_dbm_pool(test_dbm) -> psycopg2.pool.SimpleConnectionPool:
        """
        Test fixture for Database Manager's connection pool.
    
        A DATABASE CONFIGURATION AND CONNECTION IS REQUIRED FOR THESE TESTS.
        """
>       return test_dbm._dbpool
E       AttributeError: 'DatabaseManager' object has no attribute '_dbpool'

tests/test_database_manager.py:40: AttributeError
```

tests/test_ratlib.py
shatteredbeam: No further reason to strip quotes or semi-colons, as we're not vulnerable to this type of injection due to composed strings.

utils/ratlib.py
shatteredbeam: No further reason to strip quotes or semi-colons, as we're not vulnerable to this type of injection due to composed strings.

commands/debug.py
shatteredbeam: That is why it is in the debug module.  It will be needed for further work with the Fact Manager.

database/database_manager.py
shatteredbeam: This will be fixed in another iterative pass, when I move the DBM module to a context manager.

database/__init__.py
shatteredbeam: Fixed in ade5999

tests/test_database_manager.py
shatteredbeam: Fixed in 79daed8 and 6da7c08.

No teardown is necessary.  Connections are closed upon destroying the connection pool by the library.

config/testing.json
shatteredbeam: Added in ade5999.

commands/debug.py
shatteredbeam: This will be handled in a future pass.

.coveragerc
shatteredbeam: You directed me to make this a top level package in the first place?

database/database_manager.py
shatteredbeam: result's return type is a list.

`async def query(self, query: sql.SQL, values: tuple) -> list:`

.coveragerc
theunkn0wn1: I believe i directed you to make it a package, I do not believe i specified for it to be top level but implied it should match existing filestructure (under `Modules`)

database/database_manager.py
theunkn0wn1: unless Github's diff is deceiving me i don't see a `raise` statement in here. the exception raised here should not be suppressed.

.coveragerc
theunkn0wn1: I gave this further thought, scratch the above.

The Database module itself isn't a end-functionality module like commands facts are, methinks its present location is fine. the Facts manager should end up under `Modules` as it actually provides functionality that extends to Mecha's public behavior.

database/database_manager.py
shatteredbeam: with the log.warning on line 58, this is the output for a captured DatabaseError:

`E           psycopg2.DatabaseError: ('Unable to connect to database.', OperationalError('could not connect to server: Conne
ction refused (0x0000274D/10061)\n\tIs the server running on host "localhost" (::1) and accepting\n\tTCP/IP connections on
port 5432?\ncould not connect to server: Connection refused (0x0000274D/10061)\n\tIs the server running on host "localhost"
 (127.0.0.1) and accepting\n\tTCP/IP connections on port 5432?\n'))`

No other errors are directly captured, and will raise.

PR #81, opened 2019-01-03
Integrate pylint into CI build (take II)

theunkn0wn1:
This PR adds Pylint to our CI build, with the end goal of removing CodeClimate from the build system.

Rational
-------------
Im not the biggest fan of CodeClimate, i find its interface hard to use at the best of times and configuring its linting rules to be a PITA at the worst of times. 

Further, i find its inspections lacking insight, and is relatively inflexible. 

Instead, I introduce pylint into the mix to cover the linting rules. It is a well maintained and widely used QA tool in python, and boasts a wide array of options, inspections, and other utilities that may prove useful. Further, unlike codeclimate, Pylint can be run locally with consistent results. 

## what do we use Codeclimate for?
We use CodeClimate for a QA pass, specifically for linting and asserting code coverage. 

Pylint appears to be superior in terms of linting capabilities and is more readily configurable and is locally usable. 
The one thing it doesn't appear to do is code coverage, but i am actively investigating other Github integrations to reproduce that check. (the candidates I am presently looking at actually have better features in that capacity, too.)

CI changes
-----------
Linting is now part of our Circleci build process.
Linting will execute prior to running the test suite, if it picks up anything fatal the test run is aborted.
Tests only run if Linting passes.
 - the rationale here is Pylint in theory should be faster, and is more likely to pick up fatal errors that our test suite may miss. For example, https://github.com/FuelRats/pipsqueak3/pull/81/commits/0d17158e0c5f433d0666a16a7a4f3794ea68a5a7 was found as a result of Pylint rather than our test suite. 

Long term plans
---------------
Right now pylint is only set to fail a build if there is something blatently incorrect with the codebase, errors and fatals only.
In the long run, i would like to raise its potency by failing on warnings. However, some fine-tuning of pylint is required before I dare enable such a feature.
- cleaning up our codebase to fall within pylint's configuration
- tuning pylint's configuration to fall within our specs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@
 #
 version: 2
 jobs:
-  build:
+  tests:
     environment:
       # tell pipenv
       PIPENV_VENV_IN_PROJECT: true
@@ -29,7 +29,6 @@ jobs:
       # Specify service dependencies here if necessary
       # CircleCI maintains a library of pre-built images
       # documented at https://circleci.com/docs/2.0/circleci-images/
-      # - image: circleci/postgres:9.4
 
     working_directory: ~/repo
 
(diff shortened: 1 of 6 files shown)


theunkn0wn1:
this fresh in: by using some clever CircleCi tricks, i reduced the number of times Circle needs to clone the repo & install dependencies from O(n) to O(1) by using workspace caching. All jobs in a workflow will now use the same repo & venv.

this had the side effect of reducing workflow time by 10 seconds, but saves us much more if we add more jobs / parallelize the testing later on.

Eearslya:
While we're fixing CI, what's up with the circleci task that never seems to complete?

theunkn0wn1:
It doesn't complete because it nolonger exists. i renamed the build task to `tests`.

Eearslya:
Pylint seems unwilling to fully adopt PEP8 coding standards, as errors such as E301 and E303 (regarding blank lines before/after classes/functions/methods) are not supported, and will not be supported.

This causes a minor discrepancy, as pylint will not catch these types of errors, but CodeClimate will reject a build because of them. This can be annoying to outside contributors, as our linting suite passes, but the build is rejected.

To that end, I recommend we add in `pycodestyle`. `pycodestyle` is the replacement for the `pep8` package. We can simply run this over the project before or after pylint. This will verify the PEP8 standard more strictly, and is just as easily configurable, all we'd need to do is change the `[pep8]` section in `setup.cfg` to `[pycodestyle]`.

theunkn0wn1:
I plan on dropping codeclimate in its entirety.

This PR is waiting on me to review and merge the galaxy changes (unless those are WIP) so i don't change the requirements mid-pullrequest

theunkn0wn1:
Galaxy requires some rewriting, going to merge this now.

PR #85, opened 2019-01-08
Added CodeCov and pycodestyle

theunkn0wn1:
This PR makes a couple changes, Specifically throwing out CodeClimate and replacing its remaining functions with `pycodestyle` and [CodeCov](https://codecov.io/gh/theunkn0wn1/pipsqueak3/branch/ci%2Fcodecov)

If this Pr is accepted, Codeclimate will be disabled and removed from the CI Environment, and required statusus will be updated to the below list.

Changes to status reports
----------------
three new CI build steps are added:
- `ci/circleci: linting` this job contains our linting steps. both `pylint` and `pycodestyle`
     - `pycodestyle` is configured to fail the build it it finds anything, `pylint`'s config remains unchanged
- `ci/circleci:setup_workspace` this job doesn't actually run tests, but is used to create the shared workspace the other jobs depend upon. 
- `ci/circleci: tests` Our test suite.
The above three status codes will replace the existing `ci/circleci` and `ci/circleci:build` steps
Additionally, instead of CodeClimate's coverage reports we now have the following two statuses:
- `codecov/project` Project's overall Coverage. Will pass as long as the coverage does not significantly drop (threshold set at `5%`)
- `codecov/patch` Delta coverage of affected files. 

CodeCov will also create a message on the PR rendering a report on coverage changes, though it will only create it if the coverage changes. (eg documentation PRs shouldn't trigger it). It will create at most one message.


TESTING.md
```diff
@@ -35,7 +35,14 @@ with pytest.raises(SomeErrorType):
 ```
 If the above function raised _SomeErrorType_, this test would pass.  Raising a different error or if nothing is raised, fails the test.  If you are not asserting an exception was raised, it should not be caught.
 
-
+* mark asynchronous  tests with `@pytest.mark.asyncio`, otherwise they will xpass
```
shatteredbeam: Perhaps we should consider using xpass_strict = true in pytest settings?

setup.cfg
```diff
@@ -1,4 +1,2 @@
-;config file for the pep8 analysis plugin
-[pep8]
-max-line-length = 100
-exclude=tests
+[pycodestyle]
+max-line-length = 110
```
shatteredbeam: If we are changing the max line length to 110, please update the contributing page to reflect this change as well.

shatteredbeam (changes requested):
Overall this looks great.  Please take a look at the few items I've marked.

TESTING.md
theunkn0wn1: unfortunately it won't count as an xpass,
it will only count as an xpass (thus `xpass_struct`) if the test was marked to xfail to begin with,

setup.cfg
theunkn0wn1: its to give some leeway, contributors should still aim for 100 but ive set it not to go off til over 110.
that gives contributers +- 10 characters before builds start failing

theunkn0wn1: reduced to +2.

TESTING.md
shatteredbeam: Makes sense.

.circleci/config.yml
```diff
@@ -154,18 +167,25 @@ jobs:
       - run:
           name: Run pylint over project
           # run pylint, pass it through pylint-exit so Circle doesn't throw a fit
-          command: pipenv run pylint --rcfile pylint.ini Modules database utils || pipenv run python pylint-exit $?
+          command: pipenv run pylint --rcfile pylint.ini Modules database || pipenv run python pylint-exit $?
```
Eearslya: Why was `utils` removed from pylint?

theunkn0wn1: ... it wasn't supposed to be 0.o

theunkn0wn1: resolved in 29ea71a

Eearslya (approved):
LGTM

.coveragerc
```diff
@@ -1,6 +1,6 @@
 
 [run]
-branch = False
+branch = True
```
shatteredbeam: Are we still enabling branch coverage?

theunkn0wn1: that comment is against outdated code, try refreshing.
I re-disabled branch coverage as agreed upon, it is deferred til SPARK-87

shatteredbeam (approved):

MHajoha (approved):

PR #86, opened 2019-01-08
Fact Manager

shatteredbeam:
Fact Manager inherits the recent Database Manager class, creating methods for access and modification of facts stored within a facts table. 

New features for this version of the Fact Manager include:
* Increased performance, using psycopg2 as the main library.  No need for an ODBC connection.
* Specific methods for editing, adding, and deleting facts.
* A Fact object class
* A complete history for each fact object, using the ``FactManager.fact_history()`` method.
* 'Marked For Delete' functionality, similar to how it is done for case management.
* (Planned) Aliases and management for them, to set a base fact and point other names at it.

This PR provides a new table schema for the facts themselves:

```sql
CREATE TABLE fact2
(
  name character varying NOT NULL,
  lang character varying NOT NULL,
  message character varying NOT NULL,
  aliases tsvector,
  author character varying,
  edited timestamp with time zone,
  editedby character varying,
  mfd boolean,
  CONSTRAINT fact2_pkey PRIMARY KEY (name, lang)
)
```
As well as a logging table, to which events such as additions, deletes and edits can be logged for use outside of mecha3, or as part of retrieving the history of a fact object.

Planned, but not included in this PR, is a one-time conversion script to convert the existing fact database to this new schema.

At this time, Alias functionality is not completed.

Please note that the HumanFriendly library has been added as a dependency.  The fact command file, due to changes in how commands are being processed, is not included., but makes heavy use of this library.


main.py
```diff
@@ -111,11 +112,34 @@ def rat_cache(self) -> object:
         return self._rat_cache
 
     @property
-    def database_mgr(self) -> object:
+    def fact_mgr(self) -> FactManager:
```
Eearslya: Any particular reason for the abbreviation? I see that's how it used to be written, but it doesn't need to stay that way.

tests/test_fact.py
```diff
@@ -0,0 +1,204 @@
...
+See LICENSE.md
+"""
+import pytest
+import datetime
+from datetime import timezone
+
+pytestmark = pytest.mark.fact_class
+
+
+@pytest.mark.parametrize("strings", ["Shatt", "Ion4", "290ugs03", "ThISsTRiNG is AwFuL"])
```
Eearslya: Ditto the comment regarding whether getters/setters are necessary.

tests/test_fact.py
```diff
@@ -0,0 +1,204 @@
...
+    assert test_fact.complete
+
+    # unset
+    test_fact.message = ""
+    assert not test_fact.complete
+
+
+def test_fact_edited(test_fact_empty_fx):
+    """
+    Test that edited generates a datetime automagically, and returns it.
```
Eearslya: This test doesn't appear to match its documentation. It sets the edited value manually, then asserts that the set value exists?

tests/test_fact_manager.py
```diff
@@ -0,0 +1,439 @@
...
+
+@pytest.mark.asyncio
+async def test_fm_add(test_fm_fx):
+    """
+    Verify the a fact added to the database is searchable and has all required properties.
+
+    add_values = (fact.name, fact.lang, fact.message, None,
+                              fact.author, fact.edited, fact.editedby, fact.mfd)
+    """
+    fm = test_fm_fx
```
Eearslya: Why rename the fixture parameter instead of simply using it?

Ditto for most tests below.

tests/test_fact_manager.py
```diff
@@ -0,0 +1,439 @@
...
+    await fm.edit_message('test', 'en', 'Shatt', 'This fact has been edited during testing.')
+
+    result = await fm.find('test', 'en')
+
+    assert result.message == 'This fact has been edited during testing.'
+
+
+@pytest.mark.asyncio
+async def test_edit_message_failure(test_fm_fx):
```
Eearslya: Appears to be a duplicate of `test_edit_message_bad_fact`?

tests/test_fact_manager.py
```diff
@@ -0,0 +1,439 @@
...
+        await fm.edit_message('nopeavi', 'nope', 'Kovacs', 'This is not a fact')
+
+
+@pytest.mark.asyncio
+async def test_fact_exists(test_fm_fx):
+    """
+    Verify a fact looked up with the exists method returns true.
+    """
+    fm = test_fm_fx
+    assert bool(await fm.exists('stats', 'en'))
```
Eearslya: Casting the result to a `bool` seems to defeat the test's purpose of ensuring the method returns a `bool`. Almost any value aside from `None` or `False` will cast to `True`.

tests/test_fact_manager.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+@pytest.mark.asyncio
+async def test_fact_exists_no(test_fm_fx):
+    """
+    Verify a fact looked up with the exist method returns false if not found.
+    """
+    fm = test_fm_fx
+    assert not bool(await fm.exists('asdf', 'asdfa'))
```
Eearslya: Ditto the above `bool` comment.

tests/test_fact_manager.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+@pytest.mark.asyncio
+async def test_facthistory_invalid_query(test_fm_fx):
+    """
+    Verify fact history raises no errors on an invalid fact lookup.
+    """
+    fm = test_fm_fx
+
+    history = await fm.facthistory('s205gw', 'sdjn3nw')
```
Eearslya: No assertions?

tests/test_fact_manager.py
```diff
@@ -0,0 +1,439 @@
...
+    assert found_fact.complete is True
+
+
+@pytest.mark.asyncio
+async def test_exist_exception_handling(test_fm_fx, monkeypatch):
+    """
+    Verify the error is caught if psycopg2.ProgrammingError is raised.
+    """
+
+    async def boomstick(*args, **kwargs):
```
Eearslya: Duplicated helper method, consider refactoring into an `AsyncCallableMock` fixture?

utils/ratlib.py
```diff
@@ -168,6 +171,17 @@ def try_parse_uuid(suspect: str) -> UUID:
         return result
 
 
+# duration functions
+def duration(time: datetime.timedelta) -> str:
+    """
+    Converts a datetime object into a human readable duration string.
+    """
+    if not isinstance(time, datetime.timedelta):
```
Eearslya: This method appears never to be tested? Also, documentation on arguments and return value.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        try:
+            raw_results = await self.query(query, (True,))
+        except (psycopg2.ProgrammingError, psycopg2.DatabaseError) as error:
+            # ProgrammingError is a query failure, DatabaseError is database unavailable.
+            log.exception(f"Error getting MFD list.")
+            raise error
+        else:
+            result = [f"{item[0]}-{item[1]}" for item in raw_results]
+
+        return result
```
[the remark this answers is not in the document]
shatteredbeam: Incorrect.  ``self.query()`` always returns something.  worst case you'd get is any empty list.

tests/test_fact.py
shatteredbeam: Are you asking if parametrize is necessary here?  Unclear.

tests/test_fact_manager.py
shatteredbeam: Nitpick?

Modules/fact_manager.py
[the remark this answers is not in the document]
Eearslya: Except `result` is only set by the `else` block. The query's return value is stored in `raw_results`.

tests/test_fact.py
Eearslya: Commented the wrong line, this was meant to reference all of the tests for the getters/setters, and whether or not the methods they test are necessary to begin with.

tests/test_fact_manager.py
shatteredbeam: ```Verify fact history raises no errors on an invalid fact lookup.```

We are verifying that no error is thrown when looking up an invalid fact, so no assertion is necessary.

tests/test_fact_manager.py
shatteredbeam: Mock is unit test.

Modules/fact_manager.py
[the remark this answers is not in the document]
shatteredbeam: because the else: statement will only resolve if the try: block doesn't raise.  If an error is raised, then there won't be any assignment to result anyway.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        try:
+            result = await self.query(query, (name, lang))
+        except (psycopg2.ProgrammingError, psycopg2.DatabaseError) as error:
+            # Check for offline database, or issues with the query.
+            log.exception(f"Could not establish existence of {name}-{lang}")
+            raise
+
+        # We are only getting a single integer as a response, so we can unpack it by index.
+        # it will always return a single integer.
+        return result[0][0]
```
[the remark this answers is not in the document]
shatteredbeam: which will typecast into a bool if > 0.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+            fact_name: Name of fact this applies to.
+            fact_lang: langID of fact this applies to.
+            author: the user that triggered the log entry.
+            msg: Message to be written to log.
+            new_field: Contents to place in the 'new' field for the log row.
+            old_field: Contents to place in the 'old' field for the log row.
+
+        Returns: Nothing.
+        """
+        log_query = sql.SQL(f"INSERT INTO {self._FACT_LOG} VALUES "
```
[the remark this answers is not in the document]
shatteredbeam: https://www.sqlstyle.guide/

tests/test_fact_manager.py
Eearslya: Fair. Withdrawn.

grumble *GitHub won't let me mark my own comments as resolved...* grumble

tests/test_fact_manager.py
Eearslya: We have our own definition, under `tests/mock_callables.py`.

Modules/fact_manager.py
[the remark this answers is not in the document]
Eearslya: Exactly, which means that `result` will be `None` instead of an empty list.

Modules/fact_manager.py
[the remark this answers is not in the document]
Eearslya: Not sure what you're referencing, I think I was unclear. I meant to parrot the above comment regarding explicitly defining column names in INSERTs.

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+"""
+
+import datetime
+
+
+class Fact(object):
```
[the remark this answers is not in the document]
theunkn0wn1: redundant as of py3, and pylint does raise at this

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+        self._lang = lang
+        self._message = message
+        self._aliases = aliases
+        self._author = author
+        self._editedby = editedby
+        self._edited = edited if edited else datetime.datetime.now(datetime.timezone.utc)
+        self._mfd = mfd
+
+    @property
+    def name(self) -> str:
```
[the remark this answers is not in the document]
theunkn0wn1: i think you are entirely off-base.

it is to good OOP to always implement getters for public objects. Setters and deleters are optional

tests/test_fact_manager.py
theunkn0wn1: further, boolean casts are redundant as python will already cast them for use in the assertion statement.

tests/test_fact_manager.py
theunkn0wn1: i don't believe our callable mocks have the ability to raise exceptions

Modules/fact_manager.py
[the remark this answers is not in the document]
Eearslya: Aaand..you're absolutely right, and I'm the blind one. Rescinded.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        # Proclaim loudly into the void that we are loaded.
+        log.info("Fact Manager Initialized.")
+        super().__init__()
+
+    async def add(self, fact: Fact):
+        """
+        Adds a new fact to the database.  This will result in a ProgrammingError being thrown
+        if a fact violates the (name, lang) Primary Key relationship.
+
+        VERIFY IF THE FACT EXISTS WITH self.EXISTS() FIRST.
```
nathanctech: I wouldn't leave it up to the caller to determine if a fact exists. There's a warning in the comments but nothing code-wise preventing a duplicate from being entered other than an exception. Perhaps raise an error and return to the user cleanly?

shatteredbeam: You're not wrong - but remember - this is the Fact Manager - not the command side.  on the command side (ie user entry side) you'd want to run the .exists check first, and then continue with execution, delivering the message to the user at that point.  I will update the documentation to reflect this.

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+"""
+
+import datetime
+
+
+class Fact(object):
+    def __init__(self, name=None, lang='en', message=None,
```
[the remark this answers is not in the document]
shatteredbeam: This is where we default to 'en' should no language be entered on creation.

Modules/fact.py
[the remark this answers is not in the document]
shatteredbeam: fixed in 036fd0

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        except (psycopg2.ProgrammingError, psycopg2.DatabaseError) as error:
+            # Check for offline database, or issues with the query.
+            log.exception(f"Could not establish existence of {name}-{lang}")
+            raise
+
+        # We are only getting a single integer as a response, so we can unpack it by index.
+        # it will always return a single integer.
+        return result[0][0]
+
+    async def facthistory(self, fact_name: str, fact_lang: str) -> list:
```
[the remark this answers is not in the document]
shatteredbeam: Fixed in 036fd0ec

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+
+        try:
+            await self.query(log_query, query_data)
+        except (psycopg2.ProgrammingError, psycopg2.DatabaseError):
+            log.exception("Unable to write transaction log to table.")
+            raise
+
+    async def mfd(self, name: str, lang: str) -> bool:
+        """
+        Toggles the MFD flag on a fact.
```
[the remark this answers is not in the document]
shatteredbeam: fixed in 036fd0ec.

tests/test_fact_manager.py
shatteredbeam: Duplicate test removed in 036fd0ec.

tests/test_fact_manager.py
shatteredbeam: AFAIK, monkeypatch's scope is the same as the parent def. when you add a helper method in this fashion.

utils/ratlib.py
shatteredbeam: Tests added in 036fd0ec

Modules/fact.py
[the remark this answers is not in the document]
theunkn0wn1: in the `name` and `message` setters we enforce these fields to be non-nullable via the `isinstance` checks.

Therefore, should it not be reasonable to require values to be passed into their constructors?

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+                 aliases=None, author=None, edited=None,
+                 editedby=None, mfd=False):
+        self._name = name
+        self._lang = lang
+        self._message = message
+        self._aliases = aliases
+        self._author = author
+        self._editedby = editedby
+        self._edited = edited if edited else datetime.datetime.now(datetime.timezone.utc)
```
theunkn0wn1: It may be wise to actually verify that the `datetime` object passed here is timezone aware.

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+        self._mfd = mfd
+
+    @property
+    def name(self) -> str:
+        """
+        Name of fact, ie 'prep'
```
theunkn0wn1: newline between summmery and `returns:` blocks

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        return self._name
+
+    @property
+    def lang(self) -> str:
+        """
+        language code, ie 'en'
+        Returns: language of fact.
+
+        This may be may be longer than two, in the case of template facts.
```
theunkn0wn1:
```suggestion
        This may be may be longer than two characters, in the case of template facts.
```

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+        Returns: bool marked for deletion
+        """
+        return self._mfd
+
+    @name.setter
+    def name(self, value: str):
+        """
+        Sets fact name.
+        Args:
+            value: str
```
theunkn0wn1: missing summarization on value [nitpick]

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+
+        self._name = value
+
+    @lang.setter
+    def lang(self, value: str):
+        """
+        Sets fact language ID.
+        Args:
+            value: str language ID
+        Returns: Nothing
```
theunkn0wn1: newlines between blocks. Helpful hint, place the carrot over the function and hit `ctrl`+`q` 😉

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        if not isinstance(value, str):
+            raise TypeError("Fact.lang must be of string type.")
```
theunkn0wn1: should we also validate that the string is not empty?

Modules/fact.py
```diff
@@ -0,0 +1,225 @@
...
+
+Provides a Fact object for the Fact Manager Class.
+
+Copyright (c) 2018 The Fuel Rat Mischief,
+All rights reserved.
+
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+"""
```
theunkn0wn1: May i suggest you look into `dataclass`es, I think this object can be better implemented as one.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+                await self.query(add_query, add_values)
+            else:
+                # Lets handle this TypeError here, because if .complete is False, we didn't
+                # get enough information to create it without violating NOT NULL fields.
+                raise TypeError("Attempted commit on incomplete Fact.")
+
+        except (psycopg2.DatabaseError, psycopg2.IntegrityError) as error:
+            # Database is not available, or fact already exists and wasn't checked.
+            # Raise error, generating graceful cheese error and log the exception.
+            log.exception(f"Unable to add fact '{fact.name}!", error)
```
theunkn0wn1: We talked about this, `log.exception` **should not** be the exception passed as an argument!

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        Returns: Nothing
+        """
+        fact = await self.find(name, lang)
+
+        # Check the fact's property to verify it is marked for deletion
+        if fact.mfd:
+            await self._destroy(name, lang)
+        else:
+            log.exception("Attempted deletion of fact not marked for delete.")
+            raise psycopg2.ProgrammingError(f"{name}-{lang} is not marked for deletion.")
```
theunkn0wn1: or does not exist.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        # await our raw result from query
+        try:
+            rows = await self.query(query, (name, lang))
+        except (psycopg2.DatabaseError, psycopg2.ProgrammingError) as error:
+            # Check for offline database, or query errors
+            log.exception("Unable to find fact due to exception.")
+            raise
+
+        # unpack query into a fact object, or return an empty Fact if there is no result.
+        return Fact(*rows[0]) if rows else Fact()
```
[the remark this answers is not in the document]
theunkn0wn1: I think here `None` is more suited. Not finding a fact isn't necessarily an error.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+            await self.query(log_query, query_data)
+        except (psycopg2.ProgrammingError, psycopg2.DatabaseError):
+            log.exception("Unable to write transaction log to table.")
+            raise
+
+    async def mfd(self, name: str, lang: str) -> bool:
+        """
+        Toggles the 'marked for deletion' flag on a fact.
+
+        >>> mfd_value = await self.mfd("test", "en")
```
theunkn0wn1: missing check on return value

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+
+        Args:
+            name: name of fact to update
+            lang: lang of fact to update
+
+        Returns: True/False that fact was set to.
+        """
+
+        query = sql.SQL(f"SELECT mfd FROM "
+                        f"{self._FACT_TABLE} WHERE name=%s AND lang=%s")
```
theunkn0wn1: is it not a taboo to use string interpolation on a SQL query?

shatteredbeam: Those placeholders are replaced by with their values when the query is performed, by the Database manager.  working as intended.

theunkn0wn1: reffering to the fstring.
```py
f"{self._FACT_TABLE} 
```

shatteredbeam: fstring is not taboo, as it resolves before the object is composed.

```
string_something = 'foo'
sql.SQL("SELECT from {string_something} WHERE field=%s")
```

raw composed sql:  "SELECT FROM foo WHERE field=%s"

Passed to query is the raw string above.  Once composed, its immutable except for the placeholder %s, which is only modified by the cursor.execute method to insert values into said placeholder.

``cursor.execute(sql, ('bar',))``

Final string: "SELECT FROM foo WHERE field='bar'"

theunkn0wn1: field being inserted isn't user accessible. Im overthinking as always
🤷‍

tests/test_fact_manager.py
shatteredbeam: fixed in 83935a7

tests/test_fact_manager.py
shatteredbeam: 83935a7

Modules/fact.py
shatteredbeam: 83935a7

Modules/fact.py
shatteredbeam: 83935a7

Modules/fact.py
shatteredbeam: fixed in 83935a7

Modules/fact.py
shatteredbeam: fixed in 83935a7

Modules/fact_manager.py
shatteredbeam: fixed in 83935a7

Modules/fact_manager.py
shatteredbeam: fixed in 83935a7

Modules/fact_manager.py
shatteredbeam: fixed in 83935a7

Modules/fact_manager.py
theunkn0wn1: what actually happens when someone tries to add an existing fact? does it explode?

Modules/fact_manager.py
```diff
@@ -0,0 +1,337 @@
...
+        """
+        Internal method to destroy a fact.  This is not to be invoked outside the delete
+        method.
+        Args:
+            name: name of fact to destroy
+            lang: langID of fact to destroy
+
+        Returns: Nothing.
+        """
+        del_query = sql.SQL(f"DELETE FROM {self._FACT_TABLE} WHERE name=%s AND lang=%s")
```
theunkn0wn1: Should these queries be stored in the module / class scope? seems inefficient to keep creating them for /every/ call.
(referring to the various sql.SQL objects created by this codebase)

Modules/fact_manager.py
```diff
@@ -0,0 +1,337 @@
...
+                        f"ORDER BY ts DESC LIMIT 5")
+
+        query_data = (fact_name, fact_lang)
+
+        try:
+            result = await self.query(query, query_data)
+        except (psycopg2.ProgrammingError, psycopg2.DatabaseError) as error:
+            # ProgrammingError is a query failure, DatabaseError is database unavailable.
+            log.exception(f"Unable to retrieve history for {fact_name}-{fact_lang}")
+            raise
```
theunkn0wn1: While valid, we should re-raise the exception explicitly as to not confuse developers not familiar with this syntax.
```suggestion
            raise error
```

Modules/fact_manager.py
```diff
@@ -0,0 +1,337 @@
...
+        query = sql.SQL(f"SELECT name, lang, message, aliases, author, edited, editedby, mfd from "
+                        f"{self._FACT_TABLE} where name=%s AND lang=%s")
+
+        # await our raw result from query
+        try:
+            rows = await self.query(query, (name, lang))
+        except (psycopg2.DatabaseError, psycopg2.ProgrammingError) as error:
+            # Check for offline database, or query errors
+            log.exception("Unable to find fact due to exception.")
+            raise
```
theunkn0wn1: see previous.

Modules/fact_manager.py
```diff
@@ -0,0 +1,337 @@
...
+
+        # unpack query into a fact object, or return None if there is no result.
+        return Fact(*rows[0]) if rows else None
+
+    async def log(self, fact_name: str, fact_lang: str, author: str, msg: str,
```
theunkn0wn1: I fear this name is too similar to the `LOG` field we use to store the module's logger. Consider revising.

Modules/fact_manager.py
theunkn0wn1: appears not resolved, the doctest does not test `mfd_value`

Modules/fact_manager.py
```diff
@@ -0,0 +1,337 @@
...
+        try:
+            raw_results = await self.query(query, (True,))
+        except (psycopg2.ProgrammingError, psycopg2.DatabaseError) as error:
+            # ProgrammingError is a query failure, DatabaseError is database unavailable.
+            log.exception(f"Error getting MFD list.")
+            raise error
+        else:
+            result = [f"{item[0]}-{item[1]}" for item in raw_results]
+
+        return result
```
theunkn0wn1: Else block is redundant. path can only reached if no exception is thrown (caught or otherwise.)

tests/test_fact_manager.py
MHajoha: Our Mocks can raise exceptions, using `CallableMock.exception_to_raise`.

Modules/fact_manager.py
```diff
@@ -0,0 +1,337 @@
...
+from psycopg2 import sql
+
+
+from Modules.fact import Fact
+from database import DatabaseManager
+
+log = logging.getLogger(f"mecha.{__name__}")
+
+
+class FactManager(DatabaseManager):
```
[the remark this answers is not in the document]
theunkn0wn1: I catagorically reject this argument. 
tge `FactManager` is a specialized implementation of the`DatabaseManager` and therefore should inherit from it. 
1. Can the fact manager do anything without a database connection? no
2. would it be more clearly written `self._dbm.<parent method>` no.
3. Is the DBM useful standalone? no, it requires extension to be useful.
4. Should `FactManager` be able to access the parent's properties: yes. it is an extending implementation. I fail to see the problem this produces.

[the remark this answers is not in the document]
Eearslya: I can see both sides of the argument, but I'm leaning towards not inheriting.

>FactManager is a specialized implementation of the DatabaseManager

It really isn't, though. FactManager *uses methods from* the DatabaseManager, but it doesn't define a special type of database management. Specialized database managers would be more like `MySQLDatabaseManager` or `PostgresDatabaseManager`. If anything, `FactManager` is more of a specialized `TableManager`.

Further, the way this is written right now, if we add another type of Manager that inherits DatabaseManager, now we have a whole separate pool of connections.

This argument will likely become obsolete when we move from `DatabaseManager` to `limpet`, but it's an important one to keep in mind regarding how we incorporate `limpet` into the project.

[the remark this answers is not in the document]
theunkn0wn1: >but it doesn't define a special type of database management

It does. It *extends* the DBM as to provide management for our facts table.

> Specialized database managers would be more like MySQLDatabaseManager or PostgresDatabaseManager. If anything, FactManager is more of a specialized TableManager.

if we need them from the upstream `limpet`, we would change our base.
Further, as it stands we have no use for other Databases, and it falls well outside our specifications,

>Further, the way this is written right now, if we add another type of Manager that inherits 

DatabaseManager, now we have a whole separate pool of connections. 
Examples?

Modules/fact_manager.py
beepbeat: I think leaving it there is fine, as it clarifies how the flow should work and will not break, if for whatever reason, the re-raise is removed.

Modules/fact_manager.py
```diff
@@ -0,0 +1,337 @@
...
+        Args:
+            name: name of fact to search, ie. 'prep'
+            lang: language ID for fact, defaults to 'en' (see Fact Class)
+
+        Returns: Fact()
+
+        """
+        # Build SQL Object for our query
+        query = sql.SQL(f"SELECT name, lang, message, aliases, author, edited, editedby, mfd from "
+                        f"{self._FACT_TABLE} where name=%s AND lang=%s")
```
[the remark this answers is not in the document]
shatteredbeam: Alias functionality has been intentionally left out, as it is a new feature and isn't critical for functionality.  However, I did leave the framework in place to hopefully get this working at a later date.

Modules/fact_manager.py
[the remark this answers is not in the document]
shatteredbeam: If we create another instance of DatabaseManager, a new SimplePool will be created for each instance, is what he means.

Modules/fact_manager.py
shatteredbeam: If you attempt to insert a fact that already exists, without checking first, the primary key on the table is violated, and you will throw a ``psycopg2.IntegrityError``.  This error is specifically captured and handled in the add method.,

Modules/fact_manager.py
shatteredbeam: each method is written around a different query.  Some, are reused (such as finding by name and language), but not often enough to justify storing them singularly.

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        >>> if self.exists("test", "en"):
+        ...     await self.add(NewFact)
+        ... else:
+        ...     # Abandon Ship!
+        ...     raise TypeError("Attempted commit on incomplete Fact.")
+
+        Aliases cannot be inserted this way, at this time.  None is forced for fact.aliases.
+        """
+        # Don't judge this SQL, its the most compact way to do it. (Thanks PEP8)
+        add_query = sql.SQL(f"INSERT INTO {self._FACT_TABLE} "
```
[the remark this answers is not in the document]
shatteredbeam: 0d99f5d9

Modules/fact_manager.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        del_query = sql.SQL(f"DELETE FROM {self._FACT_TABLE} WHERE name=%s AND lang=%s")
+
+        await self.query(del_query, (name, lang))
+
+    async def delete(self, name: str, lang: str):
+        """
+        Deletes a fact from the database.  This is an un-recoverable action.  This must be done
+        on a fact actually marked for deletion, using !factmfd.  Doing otherwise will result in
+        a thrown psycopg2.ProgrammingError.
```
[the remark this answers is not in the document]
shatteredbeam: 0d99f5d9

Modules/fact_manager.py
shatteredbeam: Fixed in 0d99f5d9

Modules/fact_manager.py
shatteredbeam: Every instance of a bare raise has been replaced with ``raise error``

Fixed in 0d99f5d9.

Modules/fact_manager.py
shatteredbeam: Fixed in 0d99f5d9.

Modules/fact_manager.py
shatteredbeam: log has been refactored to ``add_transaction``

Fixed in 0d99f5d9.

Modules/fact_manager.py
shatteredbeam: I agree with beepbeat.

Modules/fact_manager.py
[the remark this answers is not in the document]
shatteredbeam: I believe we've all agreed to table this issue and look at alternative arrangements when we excise the DBM as is, and move to the Limpet library.

shatteredbeam:
All SQL columns have been explicitly defined in any sql statement.

Fixed in 0d99f5d9.

Modules/fact.py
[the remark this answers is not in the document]
shatteredbeam: Fixed.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+            # Database is not available, or fact already exists and wasn't checked.
+            # Raise error, generating graceful cheese error and log the exception.
+            log.exception(f"Unable to add fact '{fact.name}!")
+            raise error
+
+    async def _destroy(self, name: str, lang: str):
+        """
+        Internal method to destroy a fact.  This is not to be invoked outside the delete
+        method.
+        Args:
```
theunkn0wn1: missing `\n`

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+        ... if await fm.mfd('test', 'en'): # Attempt to mark a fact for deletion
+        ...    await fm.delete('test', 'en')
+
+        Args:
+            name: name of fact to be deleted
+            lang: langID of fact to be deleted
+
+        Returns: Nothing
+
+        Raises: psycopg2.ProgrammingError if fact not marked for deletion.
```
theunkn0wn1:
```suggestion
        Raises: psycopg2.ProgrammingError if fact is not marked for deletion.
```

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+        """
+        fact = await self.find(name, lang)
+
+        # Check the fact's property to verify it is marked for deletion
+        if fact is not None and fact.mfd:
+            await self._destroy(name, lang)
+        else:
+            log.exception("Attempted deletion of fact not marked for delete or does not exist.")
+            raise psycopg2.ProgrammingError(f"{name}-{lang} is not marked for "
+                                            f"deletion or does not exist")
```
theunkn0wn1: if/else block can be simplified.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+        else:
+            await self.add_transaction(fact_name=name, fact_lang=lang, author=editor, msg='Edited',
+                                       new_field=new_message, old_field=current_fact.message)
+
+    async def exists(self, name: str, lang: str) -> bool:
+        """
+        Check if a fact exists, without having to substantiate a full Fact object.
+
+        >>> fm = FactManager()
+        ... fact_existence = await fm.exists('test', 'en')
```
theunkn0wn1: missing doctest assertion.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+
+    async def delete(self, name: str, lang: str):
+        """
+        Deletes a fact from the database.  This is an un-recoverable action.  This must be done
+        on a fact actually marked for deletion, using !factmfd.  Doing otherwise will result in
+        a thrown psycopg2.ProgrammingError.
+
+        >>> fm = FactManager()
+        ... if await fm.mfd('test', 'en'): # Attempt to mark a fact for deletion
+        ...    await fm.delete('test', 'en')
```
theunkn0wn1: missing doctest assertion.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+        Adds a new fact to the database.  This will result in a ProgrammingError being thrown
+        if a fact violates the (name, lang) Primary Key relationship.
+
+        VERIFY IF THE FACT EXISTS WITH self.EXISTS() FIRST.
+        >>> fm = FactManager()
+        ... if self.exists("test", "en"):
+        ...     await fm.add(NewFact)
+        ... else:
+        ...     # Abandon Ship!
+        ...     raise TypeError("Attempted commit on incomplete Fact.")
```
theunkn0wn1: missing doctest assertion.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+        # We are only getting a single integer as a response, so we can unpack it by index.
+        # it will always return a single integer.
+        return result[0][0]
+
+    async def fact_history(self, fact_name: str, fact_lang: str) -> list:
+        """
+        Pulls the last 5 transaction logs for a fact, as a tuple
+
+        >>> fm = FactManager()
+        ... history_list = fm.fact_history("test", "en")
```
theunkn0wn1: missing doctest assertion.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+
+    async def find(self, name: str, lang: str) -> Fact:
+        """
+        Queries the database for a fact, and returns a populated Fact object.
+
+        See Fact class for more information on Fact properties (Modules\fact.py)
+
+        >>> fm = FactManager()
+        ... example_query = sql.SQL("SELECT * some_table WHERE some_value=%s")
+        ... returned_fact = await fm.query(example_query, ('something',))
```
theunkn0wn1: missing doctest assertion.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+        * Edited
+        * Marked for delete
+        * Unmarked for delete
+
+        >>> fm = FactManager()
+        >>> await fm.add_transaction("test", "en", "Shatt", "Marked for delete")
+
+        An Enum for these values is silly, as we need to output the strings directly
+        elsewhere, and so we do not preclude the use of future or custom strings from outside
+        mecha.
```
theunkn0wn1: image

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+            raise error
+
+        return mfd_value
+
+    async def mfd_list(self, num_results=5) -> list:
+        """
+        Returns a list of facts marked for deletion.
+
+            >>> fm = FactManager()
+            ... mfd_list = await fm.mfd_list(3)
```
theunkn0wn1: missing doctest asertion

database/database_manager.py
```diff
@@ -88,7 +120,13 @@ def __init__(self,
...
             with connection.cursor() as cursor:
                 cursor.execute(query, values)
-                result = cursor.fetchall()
+                # Check if cursor.description is NONE - meaning no results returned.
+                if cursor.description:
+                    result = cursor.fetchall()
+                else:
+                    # Return a blank tuple if there are no results, since we are
+                    # forcing this to a list.
+                    result = tuple()
```
theunkn0wn1:
```suggestion
                    result = ()
```
Tuple literal is more performant.

main.py
theunkn0wn1: i would far prefer `fact_manager`.

Modules/fact.py
```diff
@@ -0,0 +1,254 @@
...
+        Checks that a fact has all the required elements to be added, and returns true.
+
+        We do not check fact.mfd or fact.edited here, as they are forced.  Aliases may be
+        omitted if there are none to add, and fact.lang defaults to 'en' if not explicitly set.
+
+        Returns: bool True/False
+        """
+
+        return True if self.name and self.lang and self.author \
+                       and self.message and self.editedby else False
```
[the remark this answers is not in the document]
shatteredbeam: Used in two places, and I find it bad form not to check for a complete fact before committing to the database.  This method has been left in place.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+import datetime
+import logging
+
+import psycopg2
+from psycopg2 import sql
+
+from Modules.fact import Fact
+from database import DatabaseManager
+
+log = logging.getLogger(f"mecha.{__name__}")
```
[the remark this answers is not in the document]
shatteredbeam: refactored in a5f33f5.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+class FactManager(DatabaseManager):
+
+    def __init__(self, fact_table="fact2", fact_log="fact_transaction"):
```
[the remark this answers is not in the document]
shatteredbeam: added in a5f33f5.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+
+        if not isinstance(fact_log, str):
+            raise TypeError("fact log must be a string.")
+
+        self._FACT_TABLE = fact_table
+        self._FACT_LOG = fact_log
+
+        # Proclaim loudly into the void that we are loaded.
+        log.info("Fact Manager Initialized.")
+        super().__init__()
```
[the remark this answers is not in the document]
shatteredbeam: it is now!

fix in a5f33f5.

Modules/fact_manager.py
```diff
@@ -0,0 +1,369 @@
...
+                # No tuple to unpack here, so assign manually:
+                add_values = (fact.name, fact.lang, fact.message, None,
+                              fact.author, fact.edited, fact.editedby, fact.mfd)
+
+                # run INSERT query
+                await self.query(add_query, add_values)
+            else:
+                # Lets handle this TypeError here, because if .complete is False, we didn't
+                # get enough information to create it without violating NOT NULL fields.
+                raise TypeError("Attempted commit on incomplete Fact.")
```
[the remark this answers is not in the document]
shatteredbeam: Fixed in a5f33f5.

Modules/fact_manager.py
shatteredbeam: Docstring test removed in a5f33f5.

Modules/fact_manager.py
shatteredbeam: Docstring example removed in a5f33f5.

Modules/fact_manager.py
shatteredbeam: Docstring example removed in a5f33f5.

database/database_manager.py
shatteredbeam: fixed in a5f33f5.

Modules/fact.py
theunkn0wn1: bump

Modules/fact_manager.py
shatteredbeam: fixed in a5f33f5.

Modules/fact_manager.py
shatteredbeam: fixed in a5f33f5.

Modules/fact_manager.py
shatteredbeam: Addressed in a5f33f5.

Modules/fact_manager.py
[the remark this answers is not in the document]
shatteredbeam: This was changed to None in a previous commit I can't be bothered to look up ;)

Modules/fact_manager.py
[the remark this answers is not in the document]
shatteredbeam: Column names have been explicitly added to all sql statements.

main.py
shatteredbeam: This has been fixed.

tests/test_fact_manager.py
shatteredbeam: Old habit, I guess.  I see no reason to change it though in a test environment.

Modules/fact.py
shatteredbeam: We don't need to validate if the string is empty anymore with the Fact class restructure.

tests/test_fact_manager.py
theunkn0wn1: `fm` is two characters, which is little better than one. It GREATLY reduces the readability of these tests. there is no reason to rename the fixture, please revert to using the fixture directly.

tests/test_fact.py
theunkn0wn1: No unicode tests?

tests/test_fact.py
theunkn0wn1: an empty string is Falsey, the routine asserts its not None nor empty

tests/test_fact_manager.py
```diff
@@ -0,0 +1,394 @@
...
+def test_module_prop_log(test_fm_fx):
+    """
+    Verify private properties are set when passed.
+    """
+    table_name = "some fact table"
+    log_table_name = "some log table"
+
+    test_fm = FactManager(table_name, log_table_name)
+
+    assert test_fm._FACT_LOG == log_table_name
```
theunkn0wn1: If this test fails we have much larger problems than this test failing.
From what i can gather there is no possible branch in which this could fail without a broken python install.

theunkn0wn1: The init method under test either raises, which isn't being tested here, or succeeds. 
The actual unit under test here is python builtins, specifically setattr and getattr. Neither of which are overloaded

this test is out of scope.

Eearslya: I would argue this method also tests the ordering and usage of `__init__`'s arguments.

theunkn0wn1: not that we need to test that, but tests are cheap.
/shrug

PR #87, opened 2019-01-09
Unpin pytest, as pytest-asyncio has updated

Eearslya:
`pytest-asyncio` has released version 0.10.0, which fixes the incompatibility introduced when `pytest` updated to 4.1.0. We can now unpin these dependencies.


Eearslya:
This PR now additionally removes the deprecation warning that will arise when updating to `pytest` 4.1.0.
See #88 for details.

PR #88, opened 2019-01-09
Fix Pytest deprecation warning

Eearslya:
Since updating to pytest 4.1.0, they have deprecated the `message` parameter of `pytest.raises`. And for good reason. Like a lot of developers, we were using the parameter incorrectly.

The `message` parameter was meant to take a string that Pytest would output to the console if the exception was not raised. It does **not** check the exception message against the string given. That is the responsibility of the `match` parameter, which I have changed our test to use.


Eearslya:
Merged into #87.

PR #94, opened 2019-01-16
Add find_nearest_landmark method to Galaxy

Eearslya:
For SPARK-95.

This adds a new method to `Galaxy`, `find_nearest_landmark`. Given an existing `StarSystem` object, it will find the nearest system out of a hardcoded list of "landmark" star systems. This will help to provide location context for things like rescue announcements (the `(158ly from Sol)` part, specifically.)

### DIFF
--- a/Modules/galaxy/galaxy.py
+++ b/Modules/galaxy/galaxy.py
@@ -12,7 +12,7 @@ See LICENSE.md
 
 from html import escape
 import json
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
 import aiohttp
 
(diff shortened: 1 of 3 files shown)


Modules/galaxy/galaxy.py
```diff
@@ -55,6 +90,36 @@ def __init__(self, url: str = None):
...
+        closest_distance = -1
+        for _, landmark in self.LANDMARK_SYSTEMS.items():
+            distance = system.distance(landmark)
+            if closest_distance == -1 or distance < closest_distance:
+                closest_distance = round(distance, 2)
+                closest_system = landmark
+
+        if closest_system is None:
+            # Something's gone terribly wrong, likely the LANDMARK_SYSTEMS dict is empty.
+            raise RuntimeError(f"Could not find a closest landmark match for '{system.name}'!")
```
theunkn0wn1: methinks `ValueError` is more accurate here.

Eearslya: I was debating on that fact for a while. The way I see it, the only way that this function could possibly fail, is if there are no `LANDMARK_SYSTEMS` defined within the class.

As I understand it, `ValueError` is meant to be used for when an argument is somehow invalid. But as long as it's a valid system, there should be no invalid values. Thus, having an empty landmark dict is a problem with the Galaxy class, not with the argument.

theunkn0wn1: Missed this reply, woops.

Your reasoning here is perfectly valid. 👍

PR #95, opened 2019-01-16
Add a system name auto-correction method

Eearslya:
This PR adds in a new method to Ratlib, `correct_system_name`. This method performs a similar task as Mecha2's autocorrection, where it will take a procedurally generated system name and ensure that it follows the proper format, substituting letters with numbers and vice-verse where appropriate.

### DIFF
--- a/utils/ratlib.py
+++ b/utils/ratlib.py
@@ -201,8 +201,8 @@ def correct_system_name(system: str) -> str:
     system = system.upper()
     match_regex = re.compile(r"(.*)\b([A-Z01258]{2}-[A-Z01258])\s+"
                              r"([A-Z01258])\s*([0-9OIZSB]+(-[0-9OIZSB]+)?)\b")
-    letter_replacements = dict(zip('01258', 'OIZSB'))
-    number_replacements = dict((v, k) for k, v in letter_replacements.items())
+    letter_replacements = {k: v for k, v in zip('01258', 'OIZSB')}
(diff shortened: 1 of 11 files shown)


utils/ratlib.py
```diff
@@ -188,6 +188,39 @@ def duration(time: datetime.timedelta) -> str:
     return humanfriendly.format_timespan(time, detailed=False, max_units=2)
 
 
+def correct_system_name(system: str) -> str:
```
MHajoha: Can we stop piling unrelated functions in `ratlib.py`? If @theunkn0wn1 agrees, I'd like you to move this into its own module under `utils/`.

theunkn0wn1: I would argue `ratlib.py` is horribly named if this function is unrelated.

According to the module docstring its for
```
ratlib.py - IRC/Fuelrat nickname utilities
```
yet its named as `ratlib`. 

Therefore, the docstring and module's name don't match, which leads to the confusion on what the file is actually for.
`ratlib` suggests is a generic library`, the docstring suggests its specifically for nickname utilities.

MHajoha: Regardless of the docstring, I see the problem in collecting unrelated functions and classes in a single module. It's not a crime to have modules containing only a single function or class, you know.

Eearslya: Resolved via its own file.

PR #96, opened 2019-01-18
RSignal parse / Spark 51

beepbeat:
it parses the RatMama announcement (including cmdr names with leading digits)
the selfissued parser accepts ',' ';' and '|' as seperators when deciphering. it will set the CR accordingly if provided. 
The Ratmama parser will also announce which details changed upon reconnect.


Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+
+    exist_rescue: Rescue = board.find_by_name(client_name)
+
+    if exist_rescue:
+        # we got a case already!
+        await ctx.reply(f"{client_name} has reconnected! #{exist_rescue.board_index}")
+        # now let's make it more visible if stuff changed
+        diff_response: str = ""
+        if system_name.casefold() != exist_rescue.system.casefold():
+            diff_response += f"System changed! Was {exist_rescue.system}"
```
MHajoha: Needs whitespace (also punctuation might serve to provide an atmosphere of professionalism to the client).

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if platform_name.casefold() != str(exist_rescue.platform).replace('Platforms.', '').casefold():
+            diff_response += "Platform changed! "
+
+        if o2_status != exist_rescue.code_red:
+            diff_response += "O2 Status changed" if o2_status else\
+                "O2 Status changed, it is now CODE RED"
+
+        if not diff_response == "":
+            await ctx.reply(diff_response)
```
MHajoha: Shouldn't `exist_rescue` get updated with the new information here?

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+        elif part.casefold() == "xb":
+            platform = Platforms.XB
+
+        elif "o2" in part.casefold():
+            cr = (part.casefold() != "o2 ok")
+
+        else:
+            system = part
+
+        # parts.remove(part)  # remove it since it does not generate any more value
```
MHajoha: Plz remove or uncomment this line. (Probably remove)

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+from Modules.rat_rescue import Platforms
+
+pytestmark = pytest.mark.ratmama
+
+
+@pytest.mark.asyncio
+async def test_ratmama_announcement(rat_board_fx: RatBoard, async_callable_fx: AsyncCallableMock,
+                                    context_channel_fx: Context, monkeypatch):
+    async def f(msg: str):
+        print(f"would have replied with: {msg}")
```
MHajoha: I don't believe tests should print anything.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+
+    monkeypatch.setattr(context_channel_fx._user, '_nickname', "RatMama[BOT]")
+    await RatMama.handle_ratmama_announcement(context_channel_fx)
+    index = RatMama.board.find_by_name("Ajdacho").board_index
+    # assert async_callable_fx.was_called_with(
+    #     f"RATSIGNAL - CMDR Ajdacho - "
+    #     f"Reported System: Alrai (distance to be implemented) - "
+    #     f"Platform: PC - "
+    #     f"O2: OK - "
+    #     f"Language: Polish (pl-PL) (Case #{index}) (PC_SIGNAL)")
```
MHajoha: What was wrong with this? It seems much safer than just asserting that it gets called with a string.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+
+    assert async_callable_fx.was_called_with(InstanceOf(str))
+
+    async_callable_fx.reset()
+
+    await RatMama.handle_ratmama_announcement(context_channel_fx)
```
MHajoha: You appear to be testing multiple cases in one function. Please split.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+    RatMama.board = rat_board_fx
+    monkeypatch.setattr(context_channel_fx, '_words_eol',
+                        ["ratsignal H, XB, O2 OK"
+                         ]
+                        )
+    monkeypatch.setattr(context_channel_fx._user, '_nickname', "Absolver")
+    await RatMama.handle_selfissued_ratsignal(context_channel_fx)
+    case = rat_board_fx.find_by_name("Absolver")
+    assert (
+        case is not None and
```
MHajoha: Splitting these into multiple assert statements might be clearer. I doubt that pytest will be able to print its helpful info with a statement that large.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+    await RatMama.handle_ratmama_announcement(context_channel_fx)
+    index = RatMama.board.find_by_name("Ajdacho").board_index
+    # assert async_callable_fx.was_called_with(
+    #     f"RATSIGNAL - CMDR Ajdacho - "
+    #     f"Reported System: Alrai (distance to be implemented) - "
+    #     f"Platform: PC - "
+    #     f"O2: OK - "
+    #     f"Language: Polish (pl-PL) (Case #{index}) (PC_SIGNAL)")
+
+    assert async_callable_fx.was_called_with(InstanceOf(str))
```
MHajoha: It seems you never assert the contents of the rescue. Would that not be prudent?

tests/test_ratmama.py
beepbeat: This was originally the plan, however it failed even if the strings matched exactly.

MHajoha: The string don't match exactly. Take a look at RatMama.py:115:
```python
                        f" (Case #{index}) ({platform_name.upper()}_SIGNAL) "
                        )
```

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+        index = board.find_by_name(client=client_name).board_index
+        await ctx.reply(f"RATSIGNAL - CMDR {client_name} - "
+                        f"Reported System: {system_name} (distance to be implemented) - "
+                        f"Platform: {platform_name} - "
+                        f"O2: {'OK' if o2_status else 'NOT OK'} - "
+                        f"Language: {result.group('full_language')}"
+                        f" (Case #{index}) ({platform_name.upper()}_SIGNAL) "
+                        )
+
+        print(f"RATSIGNAL - CMDR {client_name} - "
```
MHajoha: Please don't just print the ratsignal. Logging something is good, but please use a logger and please don't just log the whole message.

beepbeat: You are right that should not be there at all... One debug thing more to remove

tests/test_ratmama.py
beepbeat: ...
You are right, a trailing whitespace...

Modules/RatMama.py
beepbeat: That's a design choice that should be debated. I personally do not like mecha changing stuff up in the background. In my opinion, mecha is a tool for dispatch and therefore the more control it leaves to dispatch, the better. However I can also see it updating it and telling you what changed.

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+
+@rule(r"^\s*Incoming Client:", case_sensitive=False, full_message=True, prefixless=True,
+      pass_match=False)
+async def handle_ratmama_announcement(ctx: Context):
+    if ctx.user.nickname.casefold() not in ('ratmama[bot]',):
+        return
+
+    message: str = ctx.words_eol[0]
+    result = re.fullmatch(ratmama_regex, message)
+    client_name: str = result.group("cmdr")
```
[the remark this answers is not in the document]
beepbeat: fullmatch() does also only match the regex once, and should ratmama start changing up the format of the announcement, mecha should honestly explode.

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+    IRC\s+Nickname:\s*(?P<nick>[^\s]+)   # IRC nickname
+    )?                                   # ... emphasis on "Possibly"
+    )                                    # End of the main capture group
+    \s*                                  # Handle any possible trailing whitespace
+    $                                    # End of pattern
+""")
+
+
+@rule(r"^\s*Incoming Client:", case_sensitive=False, full_message=True, prefixless=True,
+      pass_match=False)
```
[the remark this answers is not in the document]
beepbeat: I know, but what's wrong with ensuring that the parameters are always as expected?

pytest.ini
```diff
@@ -20,6 +20,7 @@ markers =
     graceful_error: 'graceful_error' tests
     database_manager: database tests
     galaxy: marks a test as a galaxy test
+    ratmama: marks a test as a ratsignal parser test
```
shatteredbeam: I think this mark should be changed to something like 'parser or r**atsignal_parse*

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+    assert len(async_callable_fx.calls) == 2
+
+    async_callable_fx.reset()
+    monkeypatch.setattr(context_channel_fx._user, '_nickname', "MasterLoon")
+    await RatMama.handle_ratmama_announcement(context_channel_fx)
+    assert not async_callable_fx.was_called
+
+
+@pytest.mark.asyncio
+async def test_manual_rsig_handler(rat_board_fx: RatBoard, async_callable_fx: AsyncCallableMock,
```
shatteredbeam: Docstrings for test, please.  It doesn't need to be a full docstring, but it should explain the basis for what you are testing.

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+"""
+import re
+from Modules.context import Context
+from Modules.rules import rule
+from typing import Optional, List
+import Modules.rat_board
+from Modules.rat_rescue import Rescue, Platforms
+import logging
+
+log = logging.getLogger(f"mecha.{__name__}")
```
shatteredbeam: log should be full capital LOG.

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+    )?                                   # ... emphasis on "Possibly"
+    )                                    # End of the main capture group
+    \s*                                  # Handle any possible trailing whitespace
+    $                                    # End of pattern
+""")
+
+
+@rule(r"^\s*Incoming Client:", case_sensitive=False, full_message=True, prefixless=True,
+      pass_match=False)
+async def handle_ratmama_announcement(ctx: Context):
```
shatteredbeam: Add Docstring, please.

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+              f"Reported System: {system_name} (distance to be implemented) - "
+              f"Platform: {platform_name} - "
+              f"O2: {'OK' if o2_status else 'NOT OK'} - "
+              f"Language: {result.group('full_language')}"
+              f" (Case #{index}) ({platform_name.upper()}_SIGNAL) "
+              )
+
+
+@rule("ratsignal", case_sensitive=False, full_message=False, pass_match=False, prefixless=True)
+async def handle_selfissued_ratsignal(ctx: Context):
```
shatteredbeam: Add docstring to the def.

theunkn0wn1:
````
Modules/RatMama.py:29:104: E501 line too long (137 > 103 characters)
Exited with code 1
```

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+        # now let's make it more visible if stuff changed
+        diff_response: str = ""
+        if system_name.casefold() != exist_rescue.system.casefold():
+            diff_response += f"System changed! Was {exist_rescue.system}"
+
+        if platform_name.casefold() != str(exist_rescue.platform).replace('Platforms.', '').casefold():
+            diff_response += "Platform changed! "
+
+        if o2_status != exist_rescue.code_red:
+            diff_response += "O2 Status changed" if o2_status else\
```
theunkn0wn1: I believe i expressed it that string concatenation is not permissible. please change this to string interpolation either via a fstring or `str.format`.

In this case, i think `str.format` is appropriate

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+from tests.mock_callables import InstanceOf
+from Modules.rat_rescue import Platforms
+
+pytestmark = pytest.mark.ratmama
+
+
+@pytest.mark.asyncio
+async def test_ratmama_announcement(rat_board_fx: RatBoard, async_callable_fx: AsyncCallableMock,
+                                    context_channel_fx: Context, monkeypatch):
+    async def f(msg: str):
```
theunkn0wn1: single char function name.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+    async_callable_fx.reset()
+    monkeypatch.setattr(context_channel_fx._user, '_nickname', "MasterLoon")
+    await RatMama.handle_ratmama_announcement(context_channel_fx)
+    assert not async_callable_fx.was_called
+
+
+@pytest.mark.asyncio
+async def test_manual_rsig_handler(rat_board_fx: RatBoard, async_callable_fx: AsyncCallableMock,
+                                   context_channel_fx: Context, monkeypatch):
+    async def f(msg: str):
```
theunkn0wn1: single char function name.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+    await RatMama.handle_selfissued_ratsignal(context_channel_fx)
+    case = rat_board_fx.find_by_name("Absolver")
+    assert (
+            case is not None and
+            case.platform == Platforms.PS and
+            case.system.casefold() == "col 285" and
+            case.irc_nickname.casefold() == "absolver" and
+            case.code_red
+    )
+    rat_board_fx.clear_board()
```
theunkn0wn1: this call is not necessary, as the fixture itself is function scoped.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+async def test_ratmama_announcement(rat_board_fx: RatBoard, async_callable_fx: AsyncCallableMock,
+                                    context_channel_fx: Context, monkeypatch):
+    async def f(msg: str):
+        print(f"would have replied with: {msg}")
+        await async_callable_fx(msg)
+
+    monkeypatch.setattr(context_channel_fx, 'reply', f)
+    RatMama.board = rat_board_fx
+
+    monkeypatch.setattr(context_channel_fx, '_words_eol',
```
theunkn0wn1: if you are going to oveerwrite the content of the fixture, may as well not use the fixture.
I instead suggest using `Context.from_message`

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+
+@rule("ratsignal", case_sensitive=False, full_message=False, pass_match=False, prefixless=True)
+async def handle_selfissued_ratsignal(ctx: Context):
+    message: str = ctx.words_eol[0]
+    message = message.replace("ratsignal", "")  # the ratsignal is nothing we are interested in anymore
+    platform: Platforms
+    cr: bool
+    system: str = ""
+
+    for rescu in board.rescues.values():
```
theunkn0wn1: missing e.

Modules/RatMama.py
```diff
@@ -0,0 +1,178 @@
...
+    else:
+        # no case for that name, we have to make our own
+        rescue: Rescue = Rescue(client=client_name, system=system_name, irc_nickname=nickname,
+                                code_red=o2_status, lang_id=lang_code)
+        if platform_name.casefold() == "pc":
+            rescue.platform = Platforms.PC
+        elif platform_name.casefold() == "pc":
+            rescue.platform = Platforms.PS
+        elif platform_name.casefold() == "xb":
+            rescue.platform = Platforms.XB
```
theunkn0wn1: dictionary mapping may be more use here.

Modules/RatMama.py
beepbeat: that is intentional to not get a linter warning regarding shadowing

tests/test_ratmama.py
beepbeat: that has already been adressed...

Modules/RatMama.py
Eearslya: I think interpolation or formatting would prove to be horrendous-looking here. This isn't as simple as "O2: OK" and "O2: NOT OK". Interpolation would turn it into a monster of a one-liner with multiple conditions that simply does far too much.

tests/test_ratmama.py
beepbeat: see below

Modules/RatMama.py
beepbeat: dictionary mapping on an Enum?

tests/test_ratmama.py
theunkn0wn1: This appears not to have been addressed as it still exists in the source.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+@pytest.mark.asyncio
+async def test_manual_rsig_handler(rat_board_fx: RatBoard, async_callable_fx: AsyncCallableMock,
+                                   context_channel_fx: Context, monkeypatch):
+    async def f(msg: str):
+        print(f"would have replied with: {msg}")
+        await async_callable_fx(msg)
+
+    monkeypatch.setattr(context_channel_fx, 'reply', f)
+    RatMama.board = rat_board_fx
+    monkeypatch.setattr(context_channel_fx, '_words_eol',
```
theunkn0wn1: If you are just going to monkeypatch all the attributes of the Context fixture, its best to simply build a new `Context` object for the test.

tests/test_ratmama.py
theunkn0wn1: Too many assertions in one statement, please split them into individual lines. Pytest won't be able to give rich assertion error information like this, and in both implementations we benefit from short-circuit evaluation.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+                        )
+    monkeypatch.setattr(context_channel_fx._user, '_nickname', "Absolver")
+    await RatMama.handle_selfissued_ratsignal(context_channel_fx)
+    case = rat_board_fx.find_by_name("Absolver")
+    assert (
+            case is not None and
+            case.platform == Platforms.XB and
+            case.system.casefold() == "h" and
+            case.irc_nickname.casefold() == "absolver" and
+            not case.code_red
```
theunkn0wn1: too many assertions in one statement.

tests/test_ratmama.py
```diff
@@ -0,0 +1,131 @@
...
+                         ]
+                        )
+    monkeypatch.setattr(context_channel_fx._user, '_nickname', "Absolver")
+    await RatMama.handle_selfissued_ratsignal(context_channel_fx)
+    case = rat_board_fx.find_by_name("Absolver")
+    assert (
+            case is not None and
+            case.platform == Platforms.PS and
+            case.system.casefold() == "col 285" and
+            case.irc_nickname.casefold() == "absolver" and
```
theunkn0wn1: too many assertions in one statement.

pytest.ini
shatteredbeam: But I did say it?

beepbeat: @shatteredbeam nitpicked on the NAME, to which I agree, it is not perfect. @theunkn0wn1 "agreed", but nitpicked on the description string, which can be extended to provide way more information, but "a RSignal parser test" should be sufficient to know what the mark is for.

Modules/RatMama.py
```diff
@@ -0,0 +1,176 @@
...
+            platform = Platforms[platform_name.upper()]
+        else:
+            LOG.warning(f"Got unknown platform from RatMama: {platform_name}")
+
+        # no case for that name, we have to make our own
+        rescue: Rescue = Rescue(client=client_name, system=system_name, irc_nickname=nickname,
+                                code_red=not o2_status, lang_id=lang_code, platform=platform)
+
+        board.append(rescue, overwrite=False)
+        index = board.find_by_name(client=client_name).board_index
```
Eearslya: `board_index` is an attribute of the rescue, no? Assuming the rescue is added to the board by reference, shouldn't we be able to retrieve the index simply with `rescue.board_index` instead of doing a text search?

Modules/RatMama.py
```diff
@@ -0,0 +1,176 @@
...
+
+@rule("ratsignal", case_sensitive=False, full_message=False, pass_match=False, prefixless=True)
+async def handle_selfissued_ratsignal(ctx: Context):
+    message: str = ctx.words_eol[0]
+    # the ratsignal is nothing we are interested anymore
+    message = re.sub("ratsignal", "", message, flags=re.I)
+
+    for rescue in board.rescues.values():
+        if rescue.irc_nickname.casefold() == ctx.user.nickname.casefold():
+            ctx.reply("You already sent a signal, please be patient while a dispatch is underway.")
```
Eearslya: Coverage suggests there is no test for this case.

Modules/RatMama.py
```diff
@@ -0,0 +1,176 @@
...
+    elif ';' in message:
+        sep = ';'
+    elif '|' in message:
+        sep = '|'
+    elif '-' in message:
+        sep = '-'
+
+    if not sep:
+        board.append(Rescue(irc_nickname=ctx.user.nickname, client=ctx.user.nickname))
+        index = board.find_by_name(ctx.user.nickname)
```
Eearslya: Coverage suggests there is no test for this case.

Further, it appears there's a small error here, as you're setting `index` to be the `Rescue` object itself, rather than its board index. I'd also like to reiterate my previous comment about searching the board for a rescue we just created.

Modules/RatMama.py
```diff
@@ -0,0 +1,176 @@
...
+        # now let's make it more visible if stuff changed
+        diff_response: str = ""
+        if system_name.casefold() != exist_rescue.system.casefold():
+            diff_response += f"System changed! "
+
+        if platform_name.casefold() != exist_rescue.platform.name.casefold():
+            diff_response += "Platform changed! "
+
+        if not o2_status != exist_rescue.code_red:
+            diff_response += "O2 Status changed!" if o2_status else\
```
Eearslya: Coverage suggests there is no test for this case.

Modules/RatMama.py
```diff
@@ -0,0 +1,176 @@
...
+        await ctx.reply(f"Case #{index} created for {ctx.user.nickname}, please set details")
+        return
+    parts: List[str] = message.split(sep)
+    system: str = "unknown"
+    cr: bool = False
+    platform: Platforms
+    for part in parts:
+        part = part.strip()
+        if part.casefold() in ("pc",):
+            platform = Platforms["PC"]
```
Eearslya: Coverage suggests there is no test for this case.

Modules/RatMama.py
```diff
@@ -0,0 +1,176 @@
...
+        client=ctx.user.nickname,
+        system=system,
+        irc_nickname=ctx.user.nickname,
+        code_red=cr,
+        platform=platform
+    )
+    board.append(rescue)
+    await ctx.reply(f"Case created for {ctx.user.nickname}"
+                    f" on {platform.name} in {system}. "
+                    f"{'O2 status is okay' if not cr else 'This is a CR!'}")
```
Eearslya: Probably should report case number?

beepbeat: it should, yes

Modules/RatMama.py
beepbeat: yes, we should and we are now.

Modules/RatMama.py
MHajoha: Have we reached a consensus on this? (@theunkn0wn1) Perhaps other doers should chime in as well.

Modules/RatMama.py
MHajoha: I don't think we are. We're still searching the board again after adding the rescue.

Modules/RatMama.py
```diff
@@ -0,0 +1,195 @@
...
+        await ctx.reply(f"RATSIGNAL - CMDR {client_name} - "
+                        f"Reported System: {system_name} (distance to be implemented) - "
+                        f"Platform: {platform_name} - "
+                        f"O2: {'OK' if o2_status else 'NOT OK'} - "
+                        f"Language: {result.group('full_language')}"
+                        f" (Case #{index}) ({platform_name.upper()}_SIGNAL)"
+                        )
+
+
+@rule("ratsignal", case_sensitive=False, full_message=False, pass_match=False, prefixless=True)
```
MHajoha: Do we want a rotsagnal to also be triggered if it is later in the message than the first word? If so, `full_message` needs to be turned on.

Modules/RatMama.py
MHajoha: This is still the case. Why is the conversation resolved?

Modules/RatMama.py
```diff
@@ -0,0 +1,195 @@
...
+    # the ratsignal is nothing we are interested anymore
+    message = re.sub("ratsignal", "", message, flags=re.I)
+
+    for rescue in board.rescues.values():
+        if rescue.irc_nickname.casefold() == ctx.user.nickname.casefold():
+            await ctx.reply(
+                "You already sent a signal, please be patient while a dispatch is underway.")
+            return
+
+    sep: chr = None
```
MHajoha: `chr` is not a class representing a single character, Python doesn't have one. [`chr` is a function.](https://docs.python.org/3.7/library/functions.html#chr)

Modules/RatMama.py
```diff
@@ -0,0 +1,195 @@
...
+    elif '-' in message:
+        sep = '-'
+
+    if not sep:
+        board.append(Rescue(irc_nickname=ctx.user.nickname, client=ctx.user.nickname))
+        index = board.find_by_name(ctx.user.nickname)
+        await ctx.reply(f"Case #{index} created for {ctx.user.nickname}, please set details")
+        return
+    parts: List[str] = message.split(sep)
+    system: str = "unknown"
```
MHajoha: This will create the rescue with system `"unknown"` if no system is specified. The proper value would be `None`, however.

Modules/RatMama.py
```diff
@@ -0,0 +1,195 @@
...
+from Modules.context import Context
+from Modules.rules import rule
+from typing import Optional, List
+import Modules.rat_board
+from Modules.rat_rescue import Rescue, Platforms
+from Modules.galaxy.galaxy import Galaxy
+import logging
+
+LOG = logging.getLogger(f"mecha.{__name__}")
+board: Modules.rat_board.RatBoard = Modules.rat_board.RatBoard()
```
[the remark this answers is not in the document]
beepbeat: That was the plan, but there was something needed to make the tests run

Modules/RatMama.py
theunkn0wn1: System is nullable and should be null rather than unknown.

Modules/RatMama.py
theunkn0wn1: we want the rule to trigger if `r@signal` is present anywhere in the body of the message

Modules/RatMama.py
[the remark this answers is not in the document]
theunkn0wn1: please let notify me once upstream changes are applied.

theunkn0wn1:
Please notify me once upstream change have been applied.

also, please address the concerns raised by other reviewers, Their issues also need to be addressed prior to merging

Modules/RatMama.py
theunkn0wn1: We never actually reached a consensus on this for some reason, the conversation started in MECHA-254 has kinda died off for no apparent reason. 

Here is my verdict on this matter:
Mecha should be allowed to upgrade in severity, but not downgrade.
In the event of a downgrade as a result of a reconnect, it should warn the dispatch and take no further action. (allow the dispatch to validate the new data and issue the appropriate commands)

1.Mecha should **not** update the system name of a returning client.
 - instead it should warn the dispatch
2. Mecha **should** allow a non-cr case to be upgraded in severity to a CR.
 - it should warn the dispatch

theunkn0wn1:
```py
        if not diff_response == "":
```

this check can be simplified, and should be too.

an empty container is treated as falsy.

distributing in the negation we get
```py
if diff_response != "":
```
which means we are checking against a truthy string value.

Thus, the check can be simplified to 
```py
if diff_response:
```

Modules/RatMama.py
shatteredbeam: I have to agree with unknown here.  Mecha should not modify a case without user interaction.  Thus a case that comes in CR should remain CR until changed by dispatch, and vice versa.  Same with any other attribute of the case.  System, Client, etc.

theunkn0wn1: As per discussion in IRC, I agree. 

Mecha should warn on any inconsistency, but not act upon it.
After all, Mecha is a dispatch **assistant**, not an autopilot.

Formalized specification in this regard:
- [ ] mecha should not modify **existing** rescues in the event of a reconnection
- [ ] mecha must be able to detect inconsistencies in a reconnection event
- [ ] mecha should emit a warning to the dispatch in the event of an inconsistency

Modules/RatMama.py
```diff
@@ -0,0 +1,203 @@
...
+    platform: Platforms = None
+    for part in parts:
+        part = part.strip()
+        if part.casefold() in ("pc",):
+            platform = Platforms["PC"]
+
+        elif part.casefold() in ("ps", "ps4", "playstation", "playstation4", "playstation 4"):
+            platform = Platforms["PS"]
+
+        elif part.casefold() in ("xb", "xb1", "xbox", "xboxone", "xbox one"):
```
shatteredbeam: [not a nitpick] These disambiguation checks are great. I like this.

tests/test_ratmama.py
```diff
@@ -0,0 +1,244 @@
...
+        Tests, that the parser does not implode upon being given a wrong platform by RatMama
+        """
+        monkeypatch.setattr(context_channel_fx, '_words_eol',
+                            ["Incoming Client: Ajdacho - System: Alrai - "
+                             "Platform: Switch - O2: OK - Language: Polish (pl-PL)"
+                             ]
+                            )
+
+        # now we just set the nickname to the allowed on
+        monkeypatch.setattr(context_channel_fx._user, '_nickname', "RatMama[BOT]")
```
shatteredbeam: Another hard-coded nickname.  Let's discuss and see if there's a better way, like pulling the config value for the set nickname for continuity.

tests/test_ratmama.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # set the message to a valid announcement
+        monkeypatch.setattr(context_channel_fx, '_words_eol',
+                            ["Incoming Client: Ajdacho - System: Alrai - "
+                             "Platform: PC - O2: OK - Language: Polish (pl-PL)"
+                             ]
+                            )
+
+        # now we just set the nickname to the allowed on
+        monkeypatch.setattr(context_channel_fx._user, '_nickname', "RatMama[BOT]")
```
shatteredbeam: [Marker] hard-coded nickname.

tests/test_ratmama.py
```diff
@@ -0,0 +1,244 @@
...
+        assert case.platform == platform
+        assert case.system.casefold() == "h"
+        assert case.irc_nickname.casefold() == "absolver"
+        assert not case.code_red
+
+        # who needs flash when they can have cleanse?
+        rat_board.clear_board()
+
+        # now lets get a bit more mean with the message
+        monkeypatch.setattr(context_channel_fx, '_words_eol',
```
shatteredbeam: [Suggestion]  Can we use pytests' parametrize function to test more strings here?  I like the idea of pulling as much data as possible, but some TDD would be great here.

Modules/RatMama.py
```diff
@@ -0,0 +1,203 @@
...
+        if not diff_response == "":
+            await ctx.reply(diff_response)
+
+    else:
+        platform = None
+
+        if platform_name.casefold() in ("pc", "ps", "xb"):
+            platform = Platforms[platform_name.upper()]
+        else:
+            LOG.warning(f"Got unknown platform from RatMama: {platform_name}")
```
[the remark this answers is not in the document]
beepbeat: That logging call is referencing the Bot.

Modules/RatMama.py
```diff
@@ -0,0 +1,203 @@
...
+    An appropriate answer will be sent to IRC.
+
+    Args:
+        ctx: Context of the announcement
+
+    Returns: None
+
+    """
+
+    if ctx.user.nickname.casefold() not in ('ratmama[bot]',):
```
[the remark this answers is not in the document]
beepbeat: Well, we can make this a hardcoded String too, and I admit it's an idea I carried over from M2 but if for whatever reason (testing perhaps) you want to grant additional permissions to people you can just add their nicks there.

Modules/RatMama.py
```diff
@@ -0,0 +1,206 @@
...
+    message: str = ctx.words_eol[0]
+    result = re.fullmatch(ratmama_regex, message)
+    client_name: str = result.group("cmdr")
+    system_name: str = result.group("system")
+    platform_name: str = result.group("platform")
+    o2_status: bool = result.group("o2") == "OK"  # false is CR
+    lang_code: str = result.group("language_code")
+    nickname: Optional[str] = result.group("nick")
+
+    exist_rescue: Rescue = ctx.bot.board.find_by_name(client_name)
```
theunkn0wn1: value may be `None`, this type annotation is inaccurate. 
```suggestion
    exist_rescue: Optional[Rescue] = ctx.bot.board.find_by_name(client_name)
```

Modules/RatMama.py
```diff
@@ -0,0 +1,206 @@
...
+
+    if exist_rescue:
+        # we got a case already!
+        await ctx.reply(f"{client_name} has reconnected! Case #{exist_rescue.board_index}")
+        # now let's make it more visible if stuff changed
+        diff_response: str = ""
```
theunkn0wn1: redundant type annotation.
```suggestion
        diff_response = ""
```

Modules/RatMama.py
```diff
@@ -0,0 +1,206 @@
...
+                        f"Reported System: {system_name} (distance to be implemented) - "
+                        f"Platform: {platform_name} - "
+                        f"O2: {'OK' if o2_status else 'NOT OK'} - "
+                        f"Language: {result.group('full_language')}"
+                        f" (Case #{index}) ({platform_name.upper()}_SIGNAL)"
+                        )
+
+
+@rule(r"\bratsignal\b", case_sensitive=False, full_message=True, pass_match=False, prefixless=True)
+async def handle_selfissued_ratsignal(ctx: Context) -> None:
```
theunkn0wn1: This name is a little confusing. when i see `selfissued` in the context of mecha, it makes it sound like this method only handles when mecha itself raises one. 

Thus, i suggest refactor/renaming this method to be more descriptive.
something to the tune of `handle_ratsignal` should do the trick.

Modules/RatMama.py
```diff
@@ -0,0 +1,206 @@
...
+    # the ratsignal is nothing we are interested anymore
+    message = re.sub("ratsignal", "", message, flags=re.I)
+
+    for rescue in ctx.bot.board.rescues.values():
+        if rescue.irc_nickname.casefold() == ctx.user.nickname.casefold():
+            await ctx.reply(
+                "You already sent a signal, please be patient while a dispatch is underway.")
+            return
+
+    sep: str = None
```
theunkn0wn1: [nitpick] improper type annotation as you set it to `None`

Consider the type hint a contract that should be followed

Modules/RatMama.py
```diff
@@ -0,0 +1,206 @@
...
+    if not sep:
+        ctx.bot.board.append(Rescue(irc_nickname=ctx.user.nickname, client=ctx.user.nickname))
+        index = ctx.bot.board.find_by_name(ctx.user.nickname)
+        await ctx.reply(f"Case #{index} created for {ctx.user.nickname}, please set details")
+        return
+    parts: List[str] = message.split(sep)
+    system: str = None
+    cr: bool = False
+    platform: Platforms = None
+    for part in parts:
```
theunkn0wn1: [minor] seems `parts` is not used elsewhere, may i suggest simplifying this iterator?
```py
for word in message.split(sep):
    ...
```

config/config.template.json
```diff
@@ -72,5 +72,10 @@
       ],
       "level": 4
     }
+  },
+  "ratsignal_parser": {
+    "announcer_nicks" : [
+      "ratmama[bot]"
+    ]
   }
 }
```
theunkn0wn1: missing `\n` at EOL

tests/test_ratmama.py
theunkn0wn1: 👍

tests/test_ratmama.py
```diff
@@ -0,0 +1,255 @@
...
+                                           monkeypatch):
+        """
+        Tests, that a wrong nickname has no associated action when handed to the RatMama handler.
+        """
+        # use a valid announcement
+        monkeypatch.setattr(context_channel_fx, '_words_eol',
+                            ["Incoming Client: Ajdacho - System: H - Platform: XB - "
+                             "O2: OK - Language: Polish (pl-PL)"
+                             ]
+                            )
```
theunkn0wn1: This monkeypatch sets `words_eol` to a strictly invalid value. 

`words_eol` is essentially generated by using `str.split(maxsplit=1)` repeatidly as to decompose the string into a list of strings. Consequently, a input that yields `words_eol[0]` with spaces in it necessitates other elements in this list.


That said, since its a test and you are up against the deadline I will differ this to a cleanup task.

PR #98, opened 2019-01-28
Add platform to Rescue's constructor

theunkn0wn1:
Adds `platform` to Rescue's constructor with a default value of `None`.


theunkn0wn1:
@beepbeat I believe you requested this a while ago

PR #100, opened 2019-01-30
Workaround to enable proper ctrl+c handling.

shatteredbeam:
Small PR to re-enable ctrl+c task ending functionality, since it was driving me bat**** crazy.

### DIFF
--- a/main.py
+++ b/main.py
@@ -14,6 +14,7 @@ This module is built on top of the Pydle system.
 """
 import asyncio
 import logging
+import signal
 from asyncio import AbstractEventLoop
 from uuid import uuid4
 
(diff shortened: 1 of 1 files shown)


theunkn0wn1:
I am concerned the Os is simply going to Kill the python process ungracefully, which means teardown tasks won't be called. I am concerned this may introduce issues, such as oprhaned file / socket handles.

Have we tested to ensure that Python's teardowns are called or if it simply terminates the program?

theunkn0wn1:
So, putting concerns about ungraceful termination aside,
Any fix relating to ctrl+c handling (KeyboardInterrupt) should probably be introduced upstream in Pydle. The root issue has to do with how the Pydle framework handles connections, and therefore any fix should be applied there.

PR #104, opened 2019-02-03
adding handler for Vhost. Added missing 'await'

shatteredbeam:
Added missing await and Vhost handler to mechaclient.py.


nathanctech:
Wait...why is this PR upgrading the Pydle version as well? Should it be, or should we have a different PR?

PR #105, opened 2019-02-12
[SPARK-103] Add board reference to MechaClient

shatteredbeam:
Title says it all.

### DIFF
--- a/mechaclient.py
+++ b/mechaclient.py
@@ -16,6 +16,7 @@ from logging import getLogger
 from uuid import uuid4
 from pydle import Client
 from Modules import rat_command, graceful_errors
+from Modules.rat_board import RatBoard
 from Modules.context import Context
 from Modules.fact_manager import FactManager
 from config import config
@@ -45,6 +46,7 @@ class MechaClient(Client):
         self._api_handler = None  # TODO: replace with handler init once it exists
         self._fact_manager = None  # Instantiate Global Fact Manager
         self._rat_cache = None  # TODO: replace with ratcache once it exists
+        self._rat_board = None  # Instantiate Rat Board
         super().__init__(*args, **kwargs)
 
     async def on_connect(self):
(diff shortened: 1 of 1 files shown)


mechaclient.py
```diff
@@ -139,3 +141,32 @@ def api_handler(self) -> object:
...
+        log.warning("Board Setter invoked!")
+        self._rat_board = value
+
+    @board.deleter
+    def board(self):
+        """
+        Mecha's Ratboard Deleter
+        """
+        log.warning("Board Deleted!")
+        del self._rat_board
```
theunkn0wn1: once the underlying is freed, methings we need to assign `None` to the attribute to prevent an attribute error in the getter. Have you tested this use-case yet?

mechaclient.py
```diff
@@ -139,3 +141,32 @@ def api_handler(self) -> object:
...
+        if not isinstance(value, RatBoard):
+            raise TypeError("board property must be of type RatBoard.")
+
+        log.warning("Board Setter invoked!")
+        self._rat_board = value
+
+    @board.deleter
+    def board(self):
+        """
+        Mecha's Ratboard Deleter
```
theunkn0wn1: [nitpick] drop `Mecha's` in the property docstrings, IMHO its redundant since it is an instance method of the `MechaClient` as it were...

mechaclient.py
shatteredbeam: This is almost a direct copy paste from Fact Manager's attribute.  Do you want to change that as well, or is this a different case?

theunkn0wn1: yeah, I think both cases have this particular bug.  before we make a code change, lets run a test
```py
def test_rat_board_deleter(bot_fx):
   """
   verifies deleting the rat board property doesn't raise an attribute error on next access
   """
   old_board = bot_fx.board
   del bot_fx.board
   new_board = bot_fx.board   # this line i expect to raise an AttributeError
   assert new_board is not old_board  
```

theunkn0wn1:
the deleter basically allows for easy destruction of the Board object, which is useful e.g. for the `!fbr` command.

The main reason this deleter is required is because this attribute is lazy-initialized, therefore we need a custom deleter behavior to reset the instance-attribute in the Mechaclient.

PR #106, opened 2019-02-22
Offline awareness Base

theunkn0wn1:
this PR introduces the basis for an global offline awareness system.

Through this system, Instances of classes can be notified in the event Mecha moves to either offline or online mode. It does so without specific registration steps, and requires no decorators.

This Pr doesn't serve as a general purpose signalling platform, rather  as a specific purpose implementation.

# The interface
`Modules.offline_awareness.OfflineAwareABC`

- Classes that wish to be made aware of mecha's online state need to inherit from this base class. 
- this ABC will automatically maintain a **weak** reference to instances of its children
   - this weak reference allows for event propagation without keeping the child unnecessarily alive 
   - a hard reference would keep the child alive where it would otherwise fall out of scope and be GC'ed
- this ABC exposes the `go_online` and `go_offline` coroutines, which serve to notify children of a state change.

## Abstract methods
These methods must be implemented by child classes, and serve as the system's public API
1. `on_online` called when the system moves into online mode
2. `on_offline` called when the system moves into offline mode

## Private methods
1. `cls.__gc` called when one of the weak references dies, used to prune the weakref list automatically
This method is not intended to be directly called.

## private members
1. `cls._storage` Stores the list of weak references, for use in internal machinery.
this object should not be tinkered with manually.

## public members and methods
1. `cls.online` boolean representing the current state of the system. True = Online
2. `cls.go_online()` coroutine to move the system to online mode. 
3. `cls.go_offline()` coroutine to move the system to offline mode.


theunkn0wn1:
```
************* Module Modules.offline_awareness
Modules/offline_awareness/__init__.py:32: [W0622(redefined-builtin), ] Redefining built-in '__name__'
Modules/offline_awareness/__init__.py:19: [C0103(invalid-name), ] Constant name "online" doesn't conform to UPPER_CASE naming style
Modules/offline_awareness/__init__.py:25: [C0103(invalid-name), ] Constant name "offline" doesn't conform to UPPER_CASE naming style
```

Modules/offline_awareness/_offline_aware_abc.py
```diff
@@ -0,0 +1,118 @@
...
+            # check if its still alive (possible to expire during iteration)
+            if reference.alive:
+                # resolve the reference
+                strong_reference = reference.peek()[0]
+
+                # and invoke its callback
+                await strong_reference.on_online()
+
+    @classmethod
+    async def go_offline(cls):
```
shatteredbeam: Please log when this function is invoked.

Modules/offline_awareness/_offline_aware_abc.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+
+    @abstractmethod
+    async def on_offline(self) -> NoReturn:
+        """
+        on_offline event callback, invoked when the system moves to "offline" mode
+        """
+
+    @classmethod
+    async def go_online(cls) -> NoReturn:
```
shatteredbeam: Please log when this function is invoked.

tests/offline_mode/test_offline_abc.py
```diff
@@ -0,0 +1,120 @@
...
+
+    assert async_callable_fx.was_called_once, "event got called multiple times!"
+
+    assert async_callable_fx.was_called_with(mode), "wrong event called! (check on_online," \
+                                                    " on_offline)"
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("mode", ['online', 'offline'])
+async def test_idempotency(offline_aware_fx, async_callable_fx, mode: str):
```
shatteredbeam: I feel like this test should win an award for 'best named method'

theunkn0wn1: 😅

Modules/offline_awareness/_offline_aware_abc.py
theunkn0wn1: will do. What level do we want? debug? info? warn?

Modules/offline_awareness/_offline_aware_abc.py
theunkn0wn1: see above

Modules/offline_awareness/_offline_aware_abc.py
shatteredbeam: Debug is fine for both.

theunkn0wn1: resolved in 

` [0ab407a](/FuelRats/pipsqueak3/pull/106/commits/0ab407a527c5660aec22fb6d776cee9f3708938a)`

Modules/offline_awareness/_offline_aware_abc.py
theunkn0wn1: resolved in 

0ab407a

Modules/offline_awareness/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+All rights reserved.
+
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+"""
+
+from ._offline_aware_abc import OfflineAwareABC
+
+# convenience aliases
```
Eearslya: Nitpick: This comment seems a bit too high up from the code it relates to below, it almost looks like a blank section.

Modules/offline_awareness/_offline_aware_abc.py
```diff
@@ -0,0 +1,123 @@
...
+    @classmethod
+    async def go_offline(cls):
+        """
+        Moves the system to "Offline" mode.
+
+        Notes:
+            This function is Idempotent.
+        """
+        if not cls.online:
+            LOG.debug("already in offline mode, ignoring call to go_offline ...")
```
Eearslya: Shouldn't there be a matching log call for `go_online` when already online?

pytest.ini
```diff
@@ -21,6 +21,7 @@ markers =
     database_manager: database tests
     galaxy: marks a test as a galaxy test
     autocorrect: tests for the autocorrect util
+    offline
```
Eearslya: No description for the test marker

Modules/offline_awareness/__init__.py
theunkn0wn1: looks more like a stray comment, will purge

Modules/offline_awareness/_offline_aware_abc.py
theunkn0wn1: yes,for the sake of consistency. will fix

pytest.ini
theunkn0wn1: >.<

theunkn0wn1: resolved in 

e531764

Modules/offline_awareness/_offline_aware_abc.py
theunkn0wn1: resolved in e531764

Modules/offline_awareness/__init__.py
theunkn0wn1: resolved in e531764

PR #111, opened 2019-03-13
Import and packaging fixes

theunkn0wn1:
This PR does two main things:

1. It restructures the packages to expose their methods on the interface, rather than have the interface forward the internal module

before
```py

from src.packages.context.context import Context
```

after
```py
# from some other src.packages package

from src.packages.context import Context
# note: `Context` remains defined as src.packages.context.context.Context, its just forwarded via the package.
```

2. Applies relative imports within the packages directory
before
```py
# from some other src.packages package

from src.packages.context import Context
```

after
```py
# from some other src.packages package

from ..context import Context
```

3. As a side effect because i was there... I fixed a couple redundancy and style issues while i was there 😅


shatteredbeam (approved):
Looks good to me.

src/packages/utils/ratlib.py
```diff
@@ -11,15 +11,15 @@
...
+import re
+from dataclasses import dataclass
 from enum import Enum
 from math import isclose, sqrt
-from uuid import UUID
 from typing import Optional
-from dataclasses import dataclass
+from uuid import UUID
+
+import humanfriendly
```
Eearslya, line 22: We should fix this while we're at it as well.
```
src/packages/utils/ratlib.py:17: [C0411(wrong-import-order), ] standard import "from enum import Enum" should be placed before "from dataclasses import dataclass"
src/packages/utils/ratlib.py:18: [C0411(wrong-import-order), ] standard import "from math import isclose, sqrt" should be placed before "from dataclasses import dataclass"
src/packages/utils/ratlib.py:19: [C0411(wrong-import-order), ] standard import "from typing import Optional" should be placed before "from dataclasses import dataclass"
src/packages/utils/ratlib.py:20: [C0411(wrong-import-order), ] standard import "from uuid import UUID" should be placed before "from dataclasses import dataclass"
src/packages/utils/ratlib.py:22: [C0411(wrong-import-order), ] third party import "import humanfriendly" should be placed before "from dataclasses import dataclass"
```

theunkn0wn1: I have a fix for that, but I felt it was out of scope. the import order here is actually correct, but pylint doesn't recognize `dataclasses` as a standard import. the fix is a 1 liner in the `pylint.rc` which i have queued for another PR.

Eearslya: Fair enough, I see it now. Rescinded!

Eearslya (approved):

PR #115, opened 2019-03-14
Lint/adjust pylint

theunkn0wn1:
Adjust the pylint configuration file, hopefully its a tad more sane now.

Supressed warnings for number of arguments and number of instance attributes, I would rather judge these ones by hand as we have some large objects int he codebase...

Also tried to fix permissions up a bit, fixing a wording error and adding a `__all__` as a wildcard import is the cleanest solution here...

### DIFF
--- a/pylint.ini
+++ b/pylint.ini
@@ -138,7 +138,9 @@ disable=print-statement,
         deprecated-sys-function,
         exception-escape,
         comprehension-escape,
-        logging-fstring-interpolation
+        logging-fstring-interpolation,
+        too-many-arguments,
+        too-many-instance-attributes
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/pylint.ini
+++ b/pylint.ini
@@ -209,7 +209,7 @@ bad-names=foo,
           tata
 
 # Naming style matching correct class attribute names.
-class-attribute-naming-style=snake_case
+class-attribute-naming-style=any
 
 # Regular expression matching correct class attribute names. Overrides class-
 # attribute-naming-style.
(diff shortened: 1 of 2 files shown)


Eearslya (approved):
Seems like nice rules to me. Do we still intend to add [pylint-quotes](https://pypi.org/project/pylint-quotes/)?

shatteredbeam (approved):
LGTM

Eearslya (commented):
On second thought, taking a look at pylint's latest output...

```
src/packages/ratmama/ratmamma_parser.py:135: [R0912(too-many-branches), handle_ratsignal] Too many branches (13/12)
```
Is this maybe another one we want to decide internally?

Do we want to keep TODOs and FIXMEs appearing in pylint output as well?

I also notice there are a handful of cyclic import warnings from `src.commands.__init__`. Are these false positives, or something to investigate?

theunkn0wn1:
too many branches is a valid warning, and a quick glance over the code in question suggests there is room for simplification and refactoring that will reduce the number of branches present.

The todos and fixmes will remain in the output, though at this point we may be better off converting them to jira tickets and removing them from source.


I am 99.9% certain the circular import warnings are false positives, but have not been able to isolate a minimum use case scenario in which to reproduce this bug (as to report it as a pylint issue).

Eearslya:
Understood. Concerns withdrawn, approval re-granted! (Even though it was never actually removed. GitHub is weird.)

theunkn0wn1:
yeah, you need to file a "changes required" review in order to withdraw approval formally.

That said, you voicing concerns via a comment was sufficient to prevent be from hitting the big green button. for future reference file a change request to voice disapproval ❤️

PR #116, opened 2019-03-14
Pylint Spring Cleaning

Eearslya:
This PR addresses the majority of pylint's complaints since the recent configuration change.

There are still a few warnings that I left in, mainly the `too-many-branches` warnings, as I feel that these are candidates for refactoring and should not be suppressed.

There is also an issue of `rule.py` redefining the name `rule`, though I'm not sure if this should be fixed or ignored.


shatteredbeam (dismissed):
LGTM

src/packages/commands/rat_command.py
```diff
@@ -15,39 +15,35 @@
...
 
 class NameCollisionException(CommandException):
     """
     Someone attempted to register a command already registered.
     """
-    pass
 
 
-_registered_commands = {}
+_REGISTERED_COMMANDS = {}
```
theunkn0wn1: registered commands is by no means a constant, as multiple things modify its contents

src/packages/commands/rat_command.py
```diff
@@ -108,25 +104,26 @@ def _register(func, names: list or str) -> bool:
...
     """
     Flushes registered commands
     Probably useless outside testing...
     """
-    global _registered_commands  # again this feels ugly but they are module-level now...
-    _registered_commands = {}
+
+    # again this feels ugly but they are module-level now...
+    global _REGISTERED_COMMANDS  # pylint: disable=global-statement
+    _REGISTERED_COMMANDS = {}
```
theunkn0wn1: Since we are modifying this statement anyways in this context we can actually get rid of the global statement.

we can use `_REGISTERED_COMMANDS.clear()` which will get rid of any reason to use the global statement


```py
_REGISTERED_COMMANDS.clear()
```

theunkn0wn1 (changes requested):
Looking good for the most part, just a couple small things that should be addressed.

tests/test_rat_command.py
```diff
@@ -108,7 +108,7 @@ def test_register_non_callable(self, garbage):
             # print(f"bot={bot}\tchannel={channel}\tsender={sender}")
             return bot, channel, sender
 
-        assert alias.lower() in Commands._REGISTERED_COMMANDS.keys()
+        assert alias.lower() in Commands._registered_commands.keys()
```
theunkn0wn1: [suggestion] access to `.keys` is redundant. the contains check for dictionary checks against the keys
```suggestion
        assert alias.lower() in Commands._registered_commands
```

theunkn0wn1 (commented):
nitpick, with a simple resolution (hopefully)

theunkn0wn1 (approved):

shatteredbeam (approved):

PR #117, opened 2019-03-15
[SPARK-109] Add pylint-quotes to linting

Eearslya:
This package adds the ability to lint our project against our string quotations. Given that python can use either single or double quotes interchangeably, it's up to us to standardize it.

Per @theunkn0wn1's suggestion, I've made it enforce double quotes for everything. This does not, however, apply to string interpolation. Within interpolated values, single quotes are allowed.

e.g. `"this is a {'fancy' if fancy else 'boring'} string"`

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -28,7 +28,7 @@ limit-inference-results=100
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=
+load-plugins=pylint_quotes
 
 # Pickle collected data for later comparisons.
 persistent=yes
@@ -562,3 +562,11 @@ known-third-party=enchant
 # Exceptions that will emit a warning when being caught. Defaults to
 # "Exception".
 overgeneral-exceptions=Exception
+
+
+[QUOTES]
+
+# Settings for pylint-quotes to enforce the usage of double quotes for strings.
+string-quote=double
+triple-quote=double
+docstring-quote=double
--- a/src/packages/cache/rat_cache.py
+++ b/src/packages/cache/rat_cache.py
@@ -31,8 +31,8 @@ class RatCache(Singleton):
         """
         if not hasattr(self, "_initalized"):
             self._initalized = True
-            self._cache_by_id: Dict[UUID, 'Rat'] = {}
-            self._cache_by_name: Dict[str, 'Rat'] = {}
+            self._cache_by_id: Dict[UUID, "Rat"] = {}
+            self._cache_by_name: Dict[str, "Rat"] = {}
             self._api_handler = api_handler
 
     @property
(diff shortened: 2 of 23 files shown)


src/packages/cache/rat_cache.py
```diff
@@ -142,14 +142,14 @@ def by_name(self, value: Dict[str, 'Rat']) -> None:
...
-                found = await self.api_handler.get_rat_by_id(id=uuid)
 
-            return found
+        found = None
+        if uuid in self.by_uuid:
+            found = self.by_uuid[uuid]
+        elif self.api_handler is not None:  # pragma: no cover
+            found = await self.api_handler.get_rat_by_id(id=uuid)
+
+        return found
```
theunkn0wn1: [scope] out of scope change.

src/packages/commands/rat_command.py
```diff
@@ -108,9 +108,9 @@ def _register(func, names: list or str) -> bool:
         if alias in _registered_commands:
             # command already registered
             raise NameCollisionException(f"attempted to re-register command(s) {alias}")
-        else:
-            formed_dict = {alias: func}
-            _registered_commands.update(formed_dict)
+
+        formed_dict = {alias: func}
+        _registered_commands.update(formed_dict)
```
theunkn0wn1: [scope]

src/packages/rescue/rat_rescue.py
```diff
@@ -555,10 +555,11 @@ def updated_at(self, value):
...
         if not isinstance(value, datetime):
             raise TypeError(f"Expected datetime, got {type(value)}")
-        elif value < self.created_at:
+
+        if value < self.created_at:
             raise ValueError(f"{value} is older than the cases creation date!")
-        else:
-            self._updated_at = value
+
+        self._updated_at = value
```
theunkn0wn1: [scope]

theunkn0wn1 (changes requested):
couple out of scope items, otherwise looks good

theunkn0wn1 (approved):
LGTM

shatteredbeam (changes requested):
[Discussion Needed] Benefit analysis

Python documentation itself allows for multiple string types, and there is no PEP standard forcing compliance with any method (outside of docstrings, which requires double quotes in docstrings (PEP257 or 258, I dont remember).

.pylintrc
```diff
@@ -562,3 +562,11 @@ known-third-party=enchant
 # Exceptions that will emit a warning when being caught. Defaults to
 # "Exception".
 overgeneral-exceptions=Exception
+
+
+[QUOTES]
+
+# Settings for pylint-quotes to enforce the usage of double quotes for strings.
+string-quote=double
```
Holladiewal, line 570: [Consensus] After a quick glance at the documentation, I believe this should be 
```suggestion
string-quote=double-avoid-escape
```
to have it enforce single quotes if it would reduce escaping

Holladiewal (changes requested):
Generally, I would like to continue using single quoted strings for single-char-strings as this is the closest representation we have of a char data type.

PR #118, opened 2019-03-17
[SPARK-48]  Unit Tests for mechaclient.py

shatteredbeam:
Unit tests for mechaclient.py


tests/conftest.py
```diff
@@ -315,4 +316,4 @@ def test_fact_fx() -> Fact:
                 editedby='Shatt',
                 mfd=False,
                 edited=None
-                )
+                )
```
theunkn0wn1: missing \n at EOL

tests/test_mechaclient.py
```diff
@@ -0,0 +1,168 @@
...
+
+pytestmark = pytest.mark.mechaclient
+
+
+def test_verify_version(bot_fx):
+    """
+    Asserts the version is set, and returning string.
+    """
+    assert bot_fx.__version__
+    assert isinstance(bot_fx.__version__, str)
```
theunkn0wn1: [scope] testing builtin.
[error] testing behavior of hardcoded attribute.

tests/conftest.py
```diff
@@ -37,6 +37,7 @@
 
 setup_logging("logs/unit_tests.log")
 
+from src.mechaclient import MechaClient
```
Eearslya: [Style] (unused-import) What does adding this import provide? I can't see any additional code that uses this import.

tests/test_mechaclient.py
```diff
@@ -0,0 +1,168 @@
...
+    Asserts the Fact Manager can be set, requires a FactManager object, and returns properly.
+    """
+    # Helper Class
+    class FakeFactManager(FactManager):
+        # Overriding the parent class init to prevent it from creating a DB connection.
+        def __init__(self):
+            ...
+
+    bot_fx.fact_manager = FakeFactManager()
+    assert isinstance(bot_fx._fact_manager, FakeFactManager)
```
Eearslya, line 78: [Discussion] Should these tests be separated? Given the parametrize, it's testing both a good path and a bad path with each invocation.

tests/test_mechaclient.py
```diff
@@ -0,0 +1,168 @@
...
+
+@pytest.mark.parametrize("rat_board_bad_types", ["Something that isn't a Rat Board", (1, 3), 3.14])
+def test_verify_rat_board(bot_fx, rat_board_bad_types):
+    """
+    Asserts the Rat Board can be set, requires a Rat Board object, and returns properly.
+    """
+    bot_fx.board = RatBoard()
+
+    assert isinstance(bot_fx._rat_board, RatBoard)
+    assert bot_fx.board is bot_fx._rat_board
```
Eearslya, line 115: [Discussion] Should these tests be separated? Given the parametrize, it's testing both a good path and a bad path with each invocation.

pytest.ini
```diff
@@ -23,6 +23,7 @@ markers =
     ratsignal_parse: marks a test as a ratsignal parser test
     autocorrect: tests for the autocorrect util
     offline: offline-aware tests
+    mechaclient: marks a test as a mechaclient test.
```
theunkn0wn1:
```suggestion
    mechaclient: mechaclient tests
```

tests/conftest.py
theunkn0wn1: Id argue this is convention / refactor more than it is style, but yes, this appears to be a needless import.

tests/test_mechaclient.py
```diff
@@ -0,0 +1,168 @@
...
+def test_verify_rat_board_private(bot_fx):
+    """
+    Asserts the private reference for the Rat Board is none, after instantiation.
+    """
+    assert bot_fx._rat_board is None
+
+
+def test_verify_api_handler(bot_fx):
+    """
+    Asserts the api_handler property returns an object, and cannot be set (no setter)
```
theunkn0wn1: [error] this is not what the test is doing.

tests/test_mechaclient.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    # Helper Class
+    class FakeAPIHandler:
+        ...
+
+    assert bot_fx.api_handler is None
```
theunkn0wn1, line 59: [error] this is already tested in a different test case.

theunkn0wn1 (changes requested):

tests/conftest.py
shatteredbeam: If its an invalid import, I'll remove it.  I'll double check.

tests/test_mechaclient.py
shatteredbeam: Not specifically this one, removing the test removes coverage for the api_handler getter.

tests/test_mechaclient.py
shatteredbeam: Docstring fixed in 96dac7b.

pytest.ini
shatteredbeam: Changed to 'mechaclient tests' in 96dac7b.

tests/test_mechaclient.py
shatteredbeam: Superfluous test removed in 96dac7b.

tests/conftest.py
shatteredbeam: unused import removed in 96dac7b.

tests/test_mechaclient.py
shatteredbeam: They could be separated, but what would be gained from splitting it?  The tests do pass with known good or bad data.

Eearslya: It's more of a convention thing IMO, making it so one test tests one thing. No real tangible benefit other than organization. (Which is still nice to have.)

tests/conftest.py
Eearslya: I believe you removed the wrong import; the one I was referring to is in `conftest.py`, which is still there.

tests/test_mechaclient.py
theunkn0wn1: Arguably, tests should only test one scenario

Either test the good path, or the bad path, but not both within a test definition.

tests/conftest.py
theunkn0wn1: This resolved?

theunkn0wn1 (approved):
LGTM

Eearslya (approved):

PR #119, opened 2019-03-17
[SPARK-110] contributing update.

shatteredbeam:
Please note this adds a tagging system for the peer review process, and is open for discussion for ALL team members.  DO contribute to the discussion, and add your input/suggest changes.

Significant changes have been added to the CONTRIBUTING.md file.  Please read it over carefully, as once adopted it will become policy for the project.

You can see the rendered file here: https://github.com/shatteredbeam/pipsqueak3/blob/doc/contributing_update_tagging/CONTRIBUTING.md

@nathanctech @beepbeat

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,13 +2,13 @@
 
 >Welcome!  Contributors are welcome, under the following guidelines.  You don't have to be a member of our GitHub Org to contribute!
 
-This project is a complete rewrite and replacement for mechas2's codebase, written in Python 3.6.5.  This project is open source, released under [BSD-3](LICENSE).
+This project is a complete rewrite and replacement for mechas2's codebase, written in Python 3.7.  This project is open source, released under [BSD-3](LICENSE).
 
-Please note, that while some organization members have direct write access, all code additions are REQUIRED to be peer reviewed.  No one contributor may review their own code for the purposes of PR approval.
+Please note, that while some organization members have direct write access, all code additions are REQUIRED to be peer reviewed.
 
 ## Getting Started
 
-1. Fork the [pipsqueak3](https://github.com/fuelrats/pipsqueak3) repo.
+1. Fork the [pipsqueak3](https://github.com/fuelrats/pipsqueak3) repository.
 2. Checkout `develop` branch
 3. Create a feature branch (See 'Feature Branches')
 4. Develop your feature/doc/fix/insanity.
@@ -18,22 +18,20 @@ Please note, that while some organization members have direct write access, all 
 8. Make requested changes, if any.
 9. Obtain cold beverage and celebrate when your PR is merged.
 
-(If you are working off of a registered issue, please include the issue number in your PR title:
-`Spark 99 / Update to Documentation`)
-
 ## Feature Branches
 
 Make a feature branch off of master using `git checkout -b feature/my-awesome-feature`.
-For other types of PRs please use one of the following:
+For other types of pull requests please use one of the following:
 
 * doc - Documentation, or Documentation Update
 * feature - New Features/Functionality
 * fix - Bug fixes
+* testing - New or updated tests.
 
 ## PR Requirements
 * **ALL** Pull Requests must undergo a peer review phase before being committed.
-* PRs for new features without **meaningful** testing will _**NOT**_ be accepted.
-* PRs that are out of scope or modify unnecessary files will _**NOT**_ be accepted.
+* Pull Requests for new features without **meaningful** testing will _**NOT**_ be accepted.
+* Pull Requests that are out of scope or modify unnecessary files will _**NOT**_ be accepted.
 
 Use docstrings and comments to document how your features function and why. Docstrings should be formatted to the flavour of [Google](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments).
 
@@ -45,11 +43,11 @@ def example_function(param1: int, param2: str) -> bool:
     it may not break the 100 characters per line PEP8 standard.  It is a
     good standard to include expected types, explain what your function
     attempts to accomplish, and what result is returned, if any.
-    
+
     Args:
         param1 (int): First Parameter.
         param2 (str): Second Parameter.
-    
+
     Returns:
         bool: True if successful, False otherwise.
     """
@@ -67,20 +65,50 @@ All pull requests must pass standard checks:
     * Circleci unit test run (Pytest)
 * Circleci deployment testing (if applicable)
 
-Everything should be runnable in your dev environment, all necessary configuration files are provided
-in VCS. for pylint you may need to manually point it at `pylint.ini`
+Everything should be executable in your development environment, all necessary configuration files are provided
+in VCS. for pylint you may need to manually point at `pylint.ini`
 ## Hygiene
 
 **Please do not squash your commits prior to your PR.**  If this is necessary, it will be brought up during the review process.
 
-Your PR should have sensible commits and messages.
+Your PR should have sensible commits and messages.  Please see the **Pull Requests Naming & Commenting** section below.
 
-All new features are required to contain tests sufficient for newly added code to be evaluated without reducing coverage.
+New features are required to contain tests sufficient for newly added code to be evaluated without reducing coverage.
 
 After branching, you should immediately use `git push -u origin feature/my-awesome-feature` to make that the default upstream ref.
 
 Before opening a PR, rebase onto develop so your PR can be merged fast-forward only, without merge commits.
-The easiest way to do that is to make the main repo a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+The easiest way to do that is to make the main repository a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+
+## Naming & Commenting Standards
+### Pull Requests Naming
+If working from a registered issue, Include the issue name enclosed in brackets in the title of your Pull Request, ie `[SPARK-99] Update to CONTRIBUTING.md`
+
+Otherwise, use the type of request.
+``[Doc] Update TOS to revoke snickers delivery``
+``[Fix] Mechaclient.py should return None after valid on_message event``
+
+You may be asked to rename your Pull Request before peer review can begin if the name is not specific enough or not within standard.
+
+### Peer Review Tagging System
+During a peer review, reviewers will add a tag to the comment to help the author understand what type of action is required to receive an approving review.
+|Comment Tag|Explanation  |
+|--|--|
+|[Blocking]|This issue must be fixed before approval.  |
+|[Consensus]|Discussion is required.  Blocked until consensus has been reached by the author and reviewer.
+|[Discussion]| Discussion is required, but this is not blocking an approval.
+|[Incomplete]| Incomplete code and/or Pull Request.
+|[Kudos]| The reviewer liked this, specifically. Go you.
+|[Nitpick]| Same as suggestion.  Not blocking, and at authors discretion to change.
+|[Style]| Breaking style guidelines, or other issue specifically with code style.
+|[Scope]| Out of scope for your Pull Request.
+|[Suggestion]| Suggested change.  Not blocking, and at authors discretion to change.
+|[Testing]| Reviewer requests a usage case, or testing to validate. Not blocking.
+|[Testing Needed]| Testing or usage case needed before approval. Blocking.
+
+Adding multiple tags to a comment is fine.  Do not combine tags.
+Good: ``[Suggestion][Kudos] This is great, but...``
+Bad:  ``[Suggestion|Kudos] This is great, but...``
 
 ## Testing
 
@@ -93,4 +121,3 @@ Please see our [Testing](./TESTING.md) file for in-depth requirements on this to
 ## Reporting Issues
 
 Please refrain from using GitHub to report Issues, as we have since moved our main Issue Tracking to JIRA. Feel free to report any issues [there](http://t.fuelr.at/help).
-
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,7 +114,8 @@ Commit Messages:
 ### Pull Requests Naming
 If working from a registered issue, Include the issue name enclosed in brackets in the title of your Pull Request, ie `[SPARK-99] Update to CONTRIBUTING.md`
 
-Otherwise, use the type of request.
+Otherwise, use the type of request:
+
 ``[Doc] Update TOS to revoke snickers delivery``
 
 ``[Fix] Mechaclient.py should return None after valid on_message event``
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
+During a peer review, reviewers will add a tag to the comment to help the author understand what type of action is required to receive an approving review.
+
+| Tag | Explanation |
+| --- | --- |
+[Blocking]|This issue must be fixed before approval.
+[Consensus]|Discussion is required.  Blocked until consensus has been reached by the author and reviewer.
+[Discussion]|Discussion is required, but this is not blocking an approval.
+[Incomplete]|Incomplete code and/or Pull Request.
+[Kudos]|The reviewer liked this, specifically. Go you.
+[Nitpick]|Same as suggestion.  Not blocking, and at authors discretion to change.
```
314numberpi: [Discussion]
```
[Nitpick]|Same as suggestion.  Not blocking, and at authors discretion to change.
```
In what way is this different to [Suggestion] and if it isn't, why have it at all?

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
 
 Before opening a PR, rebase onto develop so your PR can be merged fast-forward only, without merge commits.
-The easiest way to do that is to make the main repo a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+The easiest way to do that is to make the main repository a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+
+## Naming & Commenting Standards
+### Pull Requests Naming
+If working from a registered issue, Include the issue name enclosed in brackets in the title of your Pull Request, ie `[SPARK-99] Update to CONTRIBUTING.md`
+
+Otherwise, use the type of request.
```
314numberpi: [Suggestion]
```
Otherwise, use the type of request.
```
Perhaps in addition to the issue name?

314numberpi (commented):
Generally, looks good.

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
+Your PR should have sensible commits and messages.  Please see the **Pull Requests Naming & Commenting** section below.
 
-All new features are required to contain tests sufficient for newly added code to be evaluated without reducing coverage.
+New features are required to contain tests sufficient for newly added code to be evaluated without reducing coverage.
 
 After branching, you should immediately use `git push -u origin feature/my-awesome-feature` to make that the default upstream ref.
 
 Before opening a PR, rebase onto develop so your PR can be merged fast-forward only, without merge commits.
-The easiest way to do that is to make the main repo a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+The easiest way to do that is to make the main repository a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
```
Eearslya: [Blocking] This command points to mecha2's repo, `pipsqueak`. Mecha3's repo is `pipsqueak3`.

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
     * Circleci unit test run (Pytest)
 * Circleci deployment testing (if applicable)
 
-Everything should be runnable in your dev environment, all necessary configuration files are provided
-in VCS. for pylint you may need to manually point it at `pylint.ini`
+Everything should be executable in your development environment, all necessary configuration files are provided
+in VCS. for pylint you may need to manually point at `pylint.ini`
```
Eearslya: [Blocking] This is now false, as the pylint configuration has moved to the default `.pylintrc`, no "manual pointing" is required.

CONTRIBUTING.md
Eearslya: [Nitpick] Grammatical issue: `authors` -> `author's`

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
+| --- | --- |
+[Blocking]|This issue must be fixed before approval.
+[Consensus]|Discussion is required.  Blocked until consensus has been reached by the author and reviewer.
+[Discussion]|Discussion is required, but this is not blocking an approval.
+[Incomplete]|Incomplete code and/or Pull Request.
+[Kudos]|The reviewer liked this, specifically. Go you.
+[Nitpick]|Same as suggestion.  Not blocking, and at authors discretion to change.
+[Style]|Breaking style guidelines, or other issue specifically with code style.
+[Scope]|Out of scope for your Pull Request.
+[Suggestion]|Suggested change.  Not blocking, and at authors discretion to change.
```
Eearslya: [Nitpick] Grammatical issue: `authors` -> `author's`

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
-All new features are required to contain tests sufficient for newly added code to be evaluated without reducing coverage.
+New features are required to contain tests sufficient for newly added code to be evaluated without reducing coverage.
 
 After branching, you should immediately use `git push -u origin feature/my-awesome-feature` to make that the default upstream ref.
 
 Before opening a PR, rebase onto develop so your PR can be merged fast-forward only, without merge commits.
-The easiest way to do that is to make the main repo a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+The easiest way to do that is to make the main repository a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+
+## Naming & Commenting Standards
```
Eearslya, line 85: [Discussion] Additional standards for commit messages as they pertain to JIRA issues.

Eearslya (changes requested):

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
+[Discussion]|Discussion is required, but this is not blocking an approval.
+[Incomplete]|Incomplete code and/or Pull Request.
+[Kudos]|The reviewer liked this, specifically. Go you.
+[Nitpick]|Same as suggestion.  Not blocking, and at authors discretion to change.
+[Style]|Breaking style guidelines, or other issue specifically with code style.
+[Scope]|Out of scope for your Pull Request.
+[Suggestion]|Suggested change.  Not blocking, and at authors discretion to change.
+[Testing]|Reviewer requests a usage case, or testing to validate. Not blocking.
+[Testing Needed]|Testing or usage case needed before approval. Blocking.
+
```
theunkn0wn1, line 151: add a `[error]` block for things that are clearly errors. (whether in methodology or implementation)

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
+
+Adding multiple tags to a comment is fine.  Do not combine tags.
+Good: ``[Suggestion][Kudos] This is great, but...``
+Bad:  ``[Suggestion|Kudos] This is great, but...``
```
theunkn0wn1, line 157: [discussion] why not? the latter is easier to read.

theunkn0wn1 (changes requested):

CONTRIBUTING.md
```diff
@@ -67,20 +65,52 @@ All pull requests must pass standard checks:
...
-The easiest way to do that is to make the main repo a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+The easiest way to do that is to make the main repository a remote using `git remote add upstream git@github.com:fuelrats/pipsqueak` and then running `git pull --rebase upstream develop`.
+
+## Naming & Commenting Standards
+### Pull Requests Naming
+If working from a registered issue, Include the issue name enclosed in brackets in the title of your Pull Request, ie `[SPARK-99] Update to CONTRIBUTING.md`
+
+Otherwise, use the type of request.
+``[Doc] Update TOS to revoke snickers delivery``
+``[Fix] Mechaclient.py should return None after valid on_message event``
```
Eearslya, line 124: [Suggestion] Markdown syntax requires two newlines in order to actually break onto a new line. With just one newline, all three of these lines run together on one.

CONTRIBUTING.md
Eearslya, line 157: [Suggestion] Markdown syntax requires two newlines in order to actually break onto a new line. With just one newline, all three of these lines run together on one.

CONTRIBUTING.md
Holladiewal, line 151: [Suggestion] Add a [Questions] tag for things that don't look wrong, but may need a bit more clarification, or are just questions by that person specifically.

Holladiewal (dismissed):

CONTRIBUTING.md
shatteredbeam: Will fix the typo. Thanks.

CONTRIBUTING.md
shatteredbeam: Will fix the typo :)

CONTRIBUTING.md
shatteredbeam: It may just be me, but I think this is well covered by the [Discussion] tag.

314numberpi: Discussion and questions could cover the same topics. I guess one leans towards an individual response and the other a collective response?

CONTRIBUTING.md
shatteredbeam: "This is just a nitpick, but..." has been a really common occurrence in recent peer reviews.  Having this as a tag is just an easy way to say "This _may_ be overly picky, but ..."

TLDR, its an easy way to open the subject regarding something that may be considered pedantic, but still merits attention without going into a full blown [discussion].  Overkill, perhaps, but using it is entirely optional in lieu of a more befitting tag.

CONTRIBUTING.md
shatteredbeam: I will fix this to something more descriptive. Perhaps "Otherwise, please use the type of request along with the issue name, such as ``[Doc] Updating CONTRIBUTING.md``"

314numberpi: I realise I wasn't clear with my original comment, I mean something along the lines of this, when it is applicable.
```
[Doc] [SPARK-99] Update to CONTRIBUTING.md
```

shatteredbeam: Well, if you're working a 'registered' issue you do have a ticket number.  Otherwise you will not, and have no SPARK identifier to use...unless I still misunderstand?

314numberpi: Your original wording had me thinking of either [SPARK-99] _**or**_ [Doc], dependant on whether a Jira issue exists. Whereas, I believe the [Doc] part should exist regardless of a linked Jira issue.

shatteredbeam: Ahh, I see.  I still think this is overkill, because the type of request is also in the branch name.

ie. ``[SPARK-10298301] Update to CONTRIBUTING.md`` is expected to be in a branch with a ``doc/`` prefix, such as ``doc/update_to_contributing``


Anyone else want to weigh in here?  This is a group decision :)  @Eearslya @theunkn0wn1 @beepbeat

Eearslya: I actually feel like "bugfix/feature/doc" is a task better suited for labels, rather than a tag within the title.

Eearslya (dismissed):

nathanctech:
I'm liking this as-is! Following it will be a bit of a challenge at first but I'll add it to my cheat sheet for git stuff.

CONTRIBUTING.md
Holladiewal: For me, a Discussion is about how it **should** work, whereas a Question is about how it **does** work. They can indeed overlap.

theunkn0wn1: For me, I think there is little distinction between what necessitates a discussion and what necessitates a question. As it were, Questions often lead to a discussion as a response.

Thus, while not Discussions are results of questions, I do believe all questions result in discussions.
I think these two terms can be combined without loss in meaning.

theunkn0wn1 (dismissed):
I fail to see why `[testing]` is non-blocking. As far as i am concerned, if an extra test case is requested it should either be provided or satifactorally addressed.


as for `[Nitpick]`, i believe it is a superfluous item and should not be in the formal specification, its unnecessary

-- Superficial review, haven't fully read proposal yet

MHajoha (dismissed):

CONTRIBUTING.md
shatteredbeam: I agree with @Eearslya here.

CONTRIBUTING.md
shatteredbeam: If we move to using labels, having them separate better facilitates this.  Also,  There are many ways one could adjoin tags, and rather than have to create a separate spec for that, we can just say use them individually.

CONTRIBUTING.md
shatteredbeam: Is this blocking?  I dont see why we can't have two distinct separate meanings.  One indicates a discussion between the author and the reviewer, the other is a 'Full stop, this needs to be a group decision' tag.

nathanctech: I wonder why we're attempting to define these as blocking/non-blocking. Instead of requiring separate tags, why not make a [Blocking] tag? This way the reviewer can block a merge regardless of other tags on the review.

shatteredbeam: The idea is that some issues are blocking by the very nature of the issue, such as Error, but technically any issue could be blocking, dependent on the reviewer.

I'd like to keep the flexibility in review comments, because we need to make tags available for all reviewers without putting them in a box.

theunkn0wn1: My concept is any comment is blocking until addressed. If someone raises a question the Pr is blocked until it gets a response or is otherwise resolved. 

The only thing i can see that does not inherently block isa  kudos.

shatteredbeam: That is certainly an way to review things, but it is not the intent of the proposal as discussed in IRC, nor is it the way we have historically reviewed pull requests.  Many times suggestions and nitpicks have been brought up that have explicitly been stated as not blocking.

theunkn0wn1: Historically it *has* been how pull requests have been treated, unless explicitly enumerated otherwise.

theunkn0wn1: At the very least, IMHO, suggestions should be acknowledged. at the very least, questions deserve a response (unless marked as resolved by the issuer due to PEBCAK etc)

shatteredbeam: in light of that, can you detail the changes you'd make?  Are you suggesting we invert the system, and all tags are blocking unless we add a [Not Blocking] tag?

nathanctech (dismissed):

Eearslya (dismissed):

CONTRIBUTING.md
theunkn0wn1: as per conversations in IRC, i believe this conversation is resolved.

theunkn0wn1 (approved):
May take some adjustment, but im willing to give the current proposal a go.

If needs be, we can adjust this later, but for now it should work.

Approved with reservations

Eearslya (approved):

314numberpi (approved):
Seems to be a marked improvement that can fit most if not all reviewers.

PR #121, opened 2019-03-23
[SPARK-110] Pull request labels

theunkn0wn1:
This PR proposes a set of labels that can be applied to pull requests.

The function of the labels is to complement our PR review tags, providing a high-level status of a given pull request.

For instance, a reviewer may append the `awaiting changes` label to a pull request as to denote, without looking at comments, that the pull request is in the submitter's hand.

This pull request also proposes labels denoting the *kind* of pull request it is. For instance, this pull request would fall under `documentation`, as it changes documentation;
`policy`, as it changes policy;
 and `consensus required`, as policy changes require a consensus.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -159,7 +159,7 @@ Bad:  ``[Suggestion|Kudos] This is great, but...``
 Labels may be applied to your pull requests, for the purposes of organization and / or overall 
 status.
 
-| label | Explanation |
+|       label       |                         Explanation                          |
 | ----------------- | ------------------------------------------------------------ |
 Awaiting changes    | is on hold until requested changes have been effected
 Bug Fix             | fixes some bug
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -162,15 +162,15 @@ status.
 | label | Explanation |
 | ----------------- | ------------------------------------------------------------ |
 Awaiting changes    | is on hold until requested changes have been effected
-Feature             | adds a new feature
 Bug Fix             | fixes some bug
-Refactor            | refactors something
-Documentation       | only effects documentation
-Linting             | adjusts the linting configurations
 CI change           | reconfigures the CI environment
-Dependency change   | modifies the dependencies
-Policy change       | modifies the contributing policies
 Consensus required  | requires the consensus of all core contributors
+Dependency change   | modifies the dependencies
+Documentation       | only effects documentation
+Feature             | adds a new feature
+Refactor            | refactors something
+Linting             | adjusts the linting configurations
+Policy change       | modifies the contributing policies
 
 ## Testing
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,6 +148,13 @@ During a peer review, reviewers will add a tag to the comment to help the author
 [Kudos]|The reviewer liked this, specifically. Go you.
 [Suggestion]| An optional, alternative approach.
 
+Adding multiple tags to a comment is fine.  Do not combine tags.
+
+Good: ``[Suggestion][Kudos] This is great, but...``
+
+Bad:  ``[Suggestion|Kudos] This is great, but...``
+
+
 ## Pull request labels
 Labels may be applied to your pull requests, for the purposes of organization and / or overall 
 status.
@@ -164,14 +171,6 @@ CI change           | This pull request reconfigures the CI environment
 Dependency change   | this pull request modifies our dependencies
 Policy              | this pull request modifies our contributing policies
 Consensus required  | this pull request requires the consensus of all core contributors
-
-
-
-Adding multiple tags to a comment is fine.  Do not combine tags.
-
-Good: ``[Suggestion][Kudos] This is great, but...``
-
-Bad:  ``[Suggestion|Kudos] This is great, but...``
 
 ## Testing
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,17 +160,17 @@ Labels may be applied to your pull requests, for the purposes of organization an
 status.
 
 | label | Explanation |
-| --- | --- |
-Awaiting changes    | this pull request is on hold until requested changes have been effected
-Feature             | this pull request adds a new feature
-Bug Fix             | this pull request fixes some bug
-Refactor            | this pull request refactors something
-Documentation       | this pull request only effects documentation
-Linting             | this pull request adjusts our linting configurations
-CI change           | This pull request reconfigures the CI environment
-Dependency change   | this pull request modifies our dependencies
-Policy              | this pull request modifies our contributing policies
-Consensus required  | this pull request requires the consensus of all core contributors
+| ----------------- | ------------------------------------------------------------ |
+Awaiting changes    | is on hold until requested changes have been effected
+Feature             | adds a new feature
+Bug Fix             | fixes some bug
+Refactor            | refactors something
+Documentation       | only effects documentation
+Linting             | adjusts the linting configurations
+CI change           | reconfigures the CI environment
+Dependency change   | modifies the dependencies
+Policy change       | modifies the contributing policies
+Consensus required  | requires the consensus of all core contributors
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
-
 Adding multiple tags to a comment is fine.  Do not combine tags.
 
 Good: ``[Suggestion][Kudos] This is great, but...``
 
 Bad:  ``[Suggestion|Kudos] This is great, but...``
 
+
+## Pull request labels
+Labels may be applied to your pull requests, for the purposes of organization and / or overall
```
314numberpi, line 159: [Question] Is this supposed to be 'may', or 'should'? Another way to phrase it: is there a circumstance in which not applying a label would be considered fine?

theunkn0wn1: hmm. good question. *may* in this context means one of many. I don't see a circumstance where /all/ of the available labels could be applied concurrently. (or, rather, such a circumstance would violate scoping issues...)

Further, by using *may* that grants permission for reviewers to append tags without explicitly requiring them to do so.

It makes more sense for me to use `may` rather than `should`. `Should` denotes the submittor **must** apply labels, and also compels reviewers to do the same; which i feel is too forceful in this context.

314numberpi: The latter paragraph answers my question, thank you.

CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
+
+| label | Explanation |
+| --- | --- |
+Awaiting changes    | this pull request is on hold until requested changes have been effected
+Feature             | this pull request adds a new feature
+Bug Fix             | this pull request fixes some bug
+Refactor            | this pull request refactors something
+Documentation       | this pull request only effects documentation
+Linting             | this pull request adjusts our linting configurations
+CI change           | This pull request reconfigures the CI environment
```
Eearslya: Capitalization should be consistent throughout this table.
Either every cell has the first letter capitalized, or none do. (This applies to the `label` header, too.)

CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
 Good: ``[Suggestion][Kudos] This is great, but...``
 
 Bad:  ``[Suggestion|Kudos] This is great, but...``
 
+
+## Pull request labels
+Labels may be applied to your pull requests, for the purposes of organization and / or overall 
+status.
+
+| label | Explanation |
```
Eearslya: [Suggestion] Alphabetize the labels? Easier for people to look up the exact meaning of a label they've seen.
Also just a bit neater.

CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
+| --- | --- |
+Awaiting changes    | this pull request is on hold until requested changes have been effected
```
Eearslya: [Discussion] Should we maybe have an 'Awaiting response' label as well?
There are several ways that a PR can be stalled on the submitter that doesn't necessarily involve code changes.
Alternatively, 'Awaiting response' could replace 'Awaiting changes' entirely, and simplify the number of labels..but it's a little less clear what action is required if we only use one label for both.

CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
-
-
 Adding multiple tags to a comment is fine.  Do not combine tags.
 
 Good: ``[Suggestion][Kudos] This is great, but...``
 
 Bad:  ``[Suggestion|Kudos] This is great, but...``
 
+
+## Pull request labels
```
shatteredbeam: [Style] Please match the header indentation with the other subject markers.  this should be one more deep, so use ``###`` instead of just two.

CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
+Labels may be applied to your pull requests, for the purposes of organization and / or overall 
+status.
+
+| label | Explanation |
+| --- | --- |
+Awaiting changes    | this pull request is on hold until requested changes have been effected
+Feature             | this pull request adds a new feature
```
shatteredbeam: [Style]  Please remove the repeated 'this pull request' phrasing and just use the subject.

CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
+Bug Fix             | this pull request fixes some bug
+Refactor            | this pull request refactors something
+Documentation       | this pull request only effects documentation
+Linting             | this pull request adjusts our linting configurations
+CI change           | This pull request reconfigures the CI environment
+Dependency change   | this pull request modifies our dependencies
```
shatteredbeam: [Error] Please keep perspective. As much as it is 'our' project, properly written it should be *the* project :)

``Dependency Change | Modifies project dependencies``
``Policy | Change to contributing expectations``

These are examples, but I think you the idea.

shatteredbeam (changes requested):
I like this, but I'd like to have a conversation about some of the tags when the requested changes are completed.

CONTRIBUTING.md
theunkn0wn1: I had organized these in the fashion of what i saw was most likely to be used to least, but if you would prefer alphabetized that can be arranged.

CONTRIBUTING.md
theunkn0wn1: I argue this is not an error but style. 

At any rate, will effect

CONTRIBUTING.md
theunkn0wn1: I feel `awaiting response` would be added and removed too frequently and is therefore too broad.

CONTRIBUTING.md
theunkn0wn1: resolved in 
5aa1fd6

CONTRIBUTING.md
theunkn0wn1: resolved in 0c964a9

CONTRIBUTING.md
theunkn0wn1: resolved in 0c964a9

CONTRIBUTING.md
theunkn0wn1: This does not fall under the `Naming & Commenting Standards` key, which was the previous `##` header. Thus, it belongs under its own `##` key and not tacked onto the end.

shatteredbeam (approved):
LGTM

CONTRIBUTING.md
theunkn0wn1: resolved in 5aa1fd6

314numberpi (approved):
Looks good to me

CONTRIBUTING.md
```diff
@@ -148,14 +148,30 @@ During a peer review, reviewers will add a tag to the comment to help the author
...
+Bug Fix             | Fixes a bug
+CI change           | Modifies the configuration of Continuous Integration (CI)
+Dependency change   | Changes the project's dependencies
+Documentation       | Modifies project documentation
+Feature             | Adds a new feature
+Refactor            | Refactors an existing system
+Linting             | Adjusts the linting rules
+Policy change       | Modifies contributing or development policies
+Team decision       | Requires the consensus of all core contributors
+
```
Holladiewal, line 174: [Suggestion] Maybe add an "Awaiting review"/"Awaiting CO review" for when the only thing blocking at this point in time is a review. Several PRs have been stalled in the past because a review that blocked hasn't been looked by the reviewing party.

Holladiewal (changes requested):

CONTRIBUTING.md
theunkn0wn1: i do believe the PRs you are referring to were not clearly communicated as being ready for review. More accurately, PRs  have sat without response to previous reviews / comments which is why they have stalled. 

The addition of "awaiting review" is superfluous, as Github already has a mechanism for requesting reviews. 
image

Eearslya (approved):

PR #122, opened 2019-03-25
[SPARK-111] Resolve import error caused by wildcard import

theunkn0wn1:
- i should have known better than to trust wildcards -_-

Resolves an import error that originates from the wildcard import in `src.packages.permissions`


shatteredbeam (approved):
I heart 1 liners.  LGTM

Eearslya (approved):

PR #123, opened 2019-03-25
[SPARK-51 / SPARK-107] Additions to the RatSig parser that didn't make the cut

Holladiewal:


theunkn0wn1 (changes requested):
Moving the galaxy fixture is out of scope for this pull request.

Additions to the parser has nothing to do with the location of a test fixture.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -19,8 +19,10 @@
 from ..rescue import Rescue
 from ..rules import rule
 from ..utils import Platforms
+from ..galaxy.galaxy import Galaxy
```
Eearslya, line 22: [Nitpick] Couldn't this be simplified to `from ..galaxy import Galaxy`?

src/packages/ratmama/ratmama_parser.py
```diff
@@ -19,8 +19,10 @@
...
 
 LOG = logging.getLogger(f"mecha.{__name__}")
+Galaxy = Galaxy()
```
Eearslya, line 25: [Style] Constants should follow `UPPER_CASE` notation.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -122,9 +124,12 @@
 
         ctx.bot.board.append(rescue, overwrite=False)
         index = ctx.bot.board.find_by_name(client=client_name).board_index
+        landmark = await Galaxy.find_nearest_landmark((await Galaxy.find_system_by_name(system_name)))
```
Eearslya, line 127: [Error] This method call is risky, because `find_system_by_name` can return `None`, which will cause `find_nearest_landmark` to break.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -122,9 +124,12 @@
...
+        distance_str = f"{landmark[1]} LY from {landmark[0].name}" if landmark[1] != 0 else ""
+
         await ctx.reply(f"RATSIGNAL - CMDR {client_name} - "
-                        f"Reported System: {system_name} (distance to be implemented) - "
-                        f"Platform: {platform_name} - "
+                        f"Reported System: {system_name} ({distance_str})"
```
Eearslya, line 131: [Error] If `find_nearest_landmark` does not return a distance greater than `0`, this string will interpolate to, for instance, `"Reported System: Sol ()"`

src/packages/ratmama/ratmama_parser.py
```diff
@@ -167,39 +172,48 @@
...
     if not sep:
-        ctx.bot.board.append(Rescue(irc_nickname=ctx.user.nickname, client=ctx.user.nickname))
+        ctx.bot.board.append(Rescue(irc_nickname=ctx.user.nickname.split("[")[0],
+                                    client=ctx.user.nickname))
         index = ctx.bot.board.find_by_name(ctx.user.nickname)
         await ctx.reply(f"Case #{index} created for {ctx.user.nickname}, please set details")
         return
     system: str = None
-    code_red: bool = False
+    cr: bool = False
```
Eearslya, line 181: [Suggestion] I think `code_red` was a better name for this variable.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -167,39 +172,48 @@
...
     platform: Platforms = None
     for part in message.split(sep):
         part = part.strip()
         if part.casefold() in ("pc",):
             platform = Platforms["PC"]
+            message = re.sub(part, "", message, flags=re.IGNORECASE)
```
Eearslya, line 187: [Discussion] [Not Blocking] Is there a case in which this `re.sub` shouldn't be called? It appears to be called in every branch of this `if`. Would it not be better placed outside of the `if`s to remove the duplication?

src/packages/ratmama/ratmama_parser.py
```diff
@@ -167,39 +172,48 @@
...
-        client=ctx.user.nickname,
+        client=ctx.user.nickname.split("[")[0],
         system=system,
         irc_nickname=ctx.user.nickname,
-        code_red=code_red,
+        code_red=cr,
         platform=platform
     )
+    landmark = await Galaxy.find_nearest_landmark(
+        (await Galaxy.find_system_by_name(system))
```
Eearslya, line 211: [Error] Same concern as above. If the system does not resolve into a `StarSystem` object, `find_nearest_landmark` will break.

tests/test_ratmama.py
```diff
@@ -61,46 +70,52 @@ class TestRSignal(object):
...
         monkeypatch.setattr(context_channel_fx._user, '_nickname', f"{nick}")
         # and fire away!
-        await ratmama.handle_ratmama_announcement(context_channel_fx)
+        await RatMama.handle_ratmama_announcement(context_channel_fx)
         # lets grab the result!
         rescue: Rescue = rat_board.find_by_name("Ajdacho")
         # remember the index
         index = rescue.board_index
         # and assert, the right announcement was to be send
+        print(async_callable_fx.calls)
```
Eearslya, line 91: [Error] Debug print should be removed from production tests.

Eearslya (changes requested):
Somewhat shallow review, haven't yet thought about the methodology used, but these are the issues on the surface that stood out to me.

src/packages/ratmama/ratmama_parser.py
theunkn0wn1: [Error]

in this case, this import is erroneous as it imports from the internal module, rather than the interface.

it should be simplified to `from ..galaxy import Galaxy`

src/packages/ratmama/ratmama_parser.py
theunkn0wn1: [Error][style]
this was renamed to `code_red` intentionally, as per naming conventions and rules

config/configuration.md
```diff
@@ -75,3 +75,8 @@ API configuration elements
 |online_mode|should mecha start up in online mode?|
 |url|base url for the API|
 |tokenfile|name of the API token file, relative to `certs/`|
+
+# ratsignal_parser
+| Element| description|
+|--------|------------|
+|announcer_nicks| list of case-insensitive nicks, that are allowed to announce incoming clients|
```
theunkn0wn1, line 82: [error] empty table intentionally empty?

src/packages/ratmama/ratmama_parser.py
theunkn0wn1, line 25: [style][error] instance of an object should be lower case, further this shadows name from containing scope (`Galaxy` already defined from  import)

src/packages/ratmama/ratmama_parser.py
Holladiewal: it is not called in the event none of the conditions are met.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -167,39 +172,48 @@
...
+            cr = "o2 ok" not in part.casefold()
+            message = re.sub(part, "", message, flags=re.IGNORECASE)
 
-        else:
-            system = part
+    system = message.strip(f"{sep} ")
 
     rescue = Rescue(
-        client=ctx.user.nickname,
+        client=ctx.user.nickname.split("[")[0],
```
theunkn0wn1, line 204: [suggestion] this expression, `.split('[')[0]`, while simple, is repeated at least twice within this module. Further, i can forsee usages of this expression in other modules/

Therefore, it may be wise to move this routine into our utilities package
```py
def strip_tags(nickname:str) -> str
    return nickname.strip('[')[0]
```

tests/test_ratmama.py
```diff
@@ -26,24 +26,33 @@
...
+        Tests, that the parser does not implode upon being given a wrong platform by RatMama
         """
+        words = "Incoming Client: Ajdacho - System: ANGRBONII - "\
+                "Platform: Switch - O2: OK - Language: Polish (pl-PL)"
+        split_words = words.split(" ")
         monkeypatch.setattr(context_channel_fx, '_words_eol',
-                            ["Incoming Client: Ajdacho - System: Alrai - "
-                             "Platform: Switch - O2: OK - Language: Polish (pl-PL)"
+                            [" ".join(v) for v in
+                             [split_words[k:] for k, v in enumerate(split_words)]
```
theunkn0wn1, line 47: [error] [style] usage of single character variables

tests/test_ratmama.py
```diff
@@ -61,46 +70,52 @@ class TestRSignal(object):
...
 
         # set the message to a valid announcement
+        words = "Incoming Client: Ajdacho - System: EORLD PRI QI-Z D1-4302 - "\
+                "Platform: PC - O2: OK - Language: Polish (pl-PL)"
+        split_words = words.split(" ")
         monkeypatch.setattr(context_channel_fx, '_words_eol',
-                            ["Incoming Client: Ajdacho - System: Alrai - "
-                             "Platform: PC - O2: OK - Language: Polish (pl-PL)"
+                            [" ".join(v) for v in
+                             [split_words[k:] for k, v in enumerate(split_words)]
```
theunkn0wn1, line 78: [error][style] usage of single character variables

tests/test_ratmama.py
```diff
@@ -111,29 +126,35 @@ class TestRSignal(object):
...
 
         # set our first message
+        words = "Incoming Client: Ajdacho - System: EORLD PRI QI-Z D1-4302 - Platform: PC - O2: OK - " \
+                "Language: Polish (pl-PL)"
+        split_words = words.split(" ")
         monkeypatch.setattr(context_channel_fx, '_words_eol',
-                            ["Incoming Client: Ajdacho - System: Alrai - "
-                             "Platform: PC - O2: OK - Language: Polish (pl-PL)"
+                            [" ".join(v) for v in
+                             [split_words[k:] for k, v in enumerate(split_words)]
```
theunkn0wn1, line 134: [error][style] usage of single character variables

tests/test_ratmama.py
```diff
@@ -111,29 +126,35 @@ class TestRSignal(object):
...
         # prepare second announcement, this one has different details,
         # but is from the same commander
+        words = "Incoming Client: Ajdacho - System: ANGRBONII - Platform: XB - O2: NOT OK - " \
+                "Language: Polish (pl-PL)"
+        split_words = words.split(" ")
         monkeypatch.setattr(context_channel_fx, '_words_eol',
-                            ["Incoming Client: Ajdacho - System: H - Platform: XB - "
-                             "O2: NOT OK - Language: Polish (pl-PL)"
+                            [" ".join(v) for v in
+                             [split_words[k:] for k, v in enumerate(split_words)]
```
theunkn0wn1, line 152: [error][style] usage of single character variables

tests/test_ratmama.py
```diff
@@ -146,21 +167,25 @@ class TestRSignal(object):
...
         """
         # use a valid announcement
+        words = "Incoming Client: Ajdacho - System: H - Platform: XB - "\
+                "O2: OK - Language: Polish (pl-PL)"
+        split_words = words.split(" ")
         monkeypatch.setattr(context_channel_fx, '_words_eol',
-                            ["Incoming Client: Ajdacho - System: H - Platform: XB - "
-                             "O2: OK - Language: Polish (pl-PL)"
+                            [" ".join(v) for v in
+                             [split_words[k:] for k, v in enumerate(split_words)]
```
theunkn0wn1, line 181: [error][style] usage of single character variables

theunkn0wn1 (changes requested):
you make frequent identical monkeypatches
```py
       monkeypatch.setattr(context_channel_fx, '_words_eol',
                            [" ".join(v) for v in
                             [split_words[k:] for k, v in enumerate(split_words)]
```

I suggest making a fixture to this effect rather than monkeypatching (greatly reduces duplicity)

config/configuration.md
Holladiewal: 404 empty table not found

theunkn0wn1: ``md

# ratsignal_parser
| Element| description|
|--------|------------|
|announcer_nicks| list of case-insensitive nicks, that are allowed to announce incoming clients|
```

Comment (theunkn0wn1) [reply config/configuration.md:82]: nope, i misread here. 🤦‍♂️

File: src/packages/ratmama/ratmama_parser.py (hunk 10 above)
Comment (Holladiewal) [reply src/packages/ratmama/ratmama_parser.py:204]: looking into ratlib, this function actually exists already

Comment (theunkn0wn1) [reply src/packages/ratmama/ratmama_parser.py:204]: I could have sworn it had existed, but evidently i didn't look hard enough when i searched for it prior to rendering this suggestion 😅 

```py
def strip_name(nickname: str) -> str:
    """
    This function accepts one input `nickname` and returns the input string
    minus any tags.

    An IRC tag is anything starting with "`[`". Further, anything following a
    `[` is truncated.

    Args:
        nickname (str): raw nickname to strip tags

    Returns:
        str: nickname stripped of tags
    """
    split_string = nickname.split("[")
    return split_string[0]

``` 

so it does

In which case, best not re-invent the wheel, i suggest using this method rather than duplicating it

.github/CODEOWNERS
```diff
@@ -22,6 +22,9 @@
 # rules module
 /src/packages/rules/ @MHajoha
 
+# RatSig related module
+/src/packages/ratmama @beepbeat
```
shatteredbeam, line 26: [Scope] Modifying codeowners is out of scope for your pull request, and currently irrelevant for those without write access to the repository.  Please remove this.

src/packages/ratmama/ratmama_parser.py
shatteredbeam: Agreed. Please fix.

shatteredbeam (changes requested):
I think Earsy and Unknown have covered almost all the same issues I would have pointed out.

src/packages/ratmama/ratmama_parser.py
Eearslya: As this is a module-level constant, don't you mean upper case?

.github/CODEOWNERS
Holladiewal: Modifying Codeownership is something that should have been in the original PR, therefore it's well within scope of this one.

 Even if it's useless since I have no write access, I still have the right to put my thumbprint on modules that have been created by me.

src/packages/ratmama/ratmama_parser.py
Holladiewal: I agree it should be uppercase since it's a module level constant, even if it's a temporary one.

.github/CODEOWNERS
shatteredbeam: I'm not going to get drawn into a debate with you about 'rights' with your code.  Once it is integrated into the project its no longer 'your code.'  That ideology is incompatible with an open-source project.

That point aside, anyone can see who contributed to the project and who authored specific commits.   What you are proposing is not the intended usage of the codeowners file.  If you had submitted a modification to the codeowners file as part of your original PR, It also would have been considered out of scope.

theunkn0wn1:
the function and purpose of this file is not to assign who owns what
files(the Fuel Rats own everything) but rather toassign who is
responsiblefor Reviewing changes to the specified files.

it's clear *who* wrote a bit of code by looking at who committed the lines
in question, as provided

For its intended use, this file's settings are only enforced for members
with write access. GH will not permit a code owner that does not have write
access

putting all that aside, it is beyond the scope of a feature pull request to
effect policy change.By modifying this file, you are inherently modifying
policy.

src/packages/ratmama/ratmama_parser.py
theunkn0wn1: yes. the original naming made it confusing to tell if it was constant or not. if its a constant then yes, UPPER_CASE is correct

PR #126, opened 2019-03-26
[SPARK-122] Adjust Galaxy's search parameters to use API indexes

Eearslya:
Galaxy was performing searches for system main stars by using the `systemId64` and `isMainStar` properties, which are not indexed. This caused slow and inefficient searches for the API.

This changes the search to use the `systemId` property, which is properly indexed. It also adjusts the code to find the Main star by itself, which should be relatively lightweight.

### DIFF
--- a/src/packages/galaxy/galaxy.py
+++ b/src/packages/galaxy/galaxy.py
@@ -80,15 +80,34 @@ class Galaxy:
         data = await self._call("api/systems", {"filter[name:eq]": name.upper()})
         result_count = data['meta']['results']['available']
         if result_count > 0:
+            system_id = data['data'][0]['id']
             sys = data['data'][0]['attributes']
-            main_star = await self._call("api/stars",
-                                         {"filter[systemId64:eq]": sys['id64'],
-                                          "filter[isMainStar:eq]": 1})
-            if main_star['meta']['results']['available'] > 0:
-                sys['spectral_class'] = main_star['data'][0]['attributes']['subType'][0]
+            main_star = await self._find_main_star(system_id)
+            sys['spectral_class'] = main_star['spectral_class'] if main_star is not None else None
             return StarSystem(position=Vector(**sys['coords']),
                               name=sys['name'],
-                              spectral_class=sys.get('spectral_class'))
+                              spectral_class=sys['spectral_class'])
+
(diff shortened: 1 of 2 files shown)


Holladiewal:
I think the BugFix label is more appropriate here, as refactoring is changing the structure of the code without changing it's (external) behaviour. Since this changes the requests that are send to the API, it is explicitly changing behaviour which is out of scope for a refactor.

shatteredbeam:
I'm Okay with refactor.  This is more of an optimization pass than a direct bugfix.

src/packages/galaxy/galaxy.py
```diff
@@ -80,15 +80,34 @@ def __init__(self, url: str = None):
...
-            if main_star['meta']['results']['available'] > 0:
-                sys['spectral_class'] = main_star['data'][0]['attributes']['subType'][0]
+            main_star = await self._find_main_star(system_id)
+            sys['spectral_class'] = main_star['spectral_class'] if main_star is not None else None
             return StarSystem(position=Vector(**sys['coords']),
                               name=sys['name'],
-                              spectral_class=sys.get('spectral_class'))
+                              spectral_class=sys['spectral_class'])
+
+    async def _find_main_star(self, system_id: int) -> Optional[Dict]:
```
shatteredbeam, line 91: [Nitpick] Does this *actually* return None if the system isn't found?  Where does that happen?

shatteredbeam (changes requested):
I think you may want to make sure that your method explicitly returns None to prevent an error case, but other than that looks good.  I'm okay with this not being explicitly set *if* you can demonstrate that the method can ONLY return the intended type or None.

src/packages/galaxy/galaxy.py
Eearslya: The `return None` happens implicitly with Python if the function reaches the end without an explicit `return`.

This case is also tested with `test_find_system_by_invalid_name`, asserting the result is `None`.

shatteredbeam: Good enough for me.  Thanks.

shatteredbeam (dismissed):

theunkn0wn1 (approved):
LGTM

PR #128, opened 2019-03-26
[Documentation] Establish a project Code of Conduct

shatteredbeam:
This is a verbatim copy of the Contributor Covenant available here (https://www.contributor-covenant.org/version/1/4/code-of-conduct.html) and released via CC license.  The only modification made was to insert a proper email address.

I propose that we add this to the project, as we are currently lacking in a code of conduct.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -55,7 +55,7 @@ further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting the project lead. All
+reported by contacting the project lead at unknown@fuelrats.com. All
 complaints will be reviewed and investigated and will result in a response that
 is deemed necessary and appropriate to the circumstances. The project team is
 obligated to maintain confidentiality with regard to the reporter of an incident.
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+  advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project lead. All
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. The project team is
+obligated to maintain confidentiality with regard to the reporter of an incident.
+Further details of specific enforcement policies may be posted separately.
+
+Project maintainers who do not follow or enforce the Code of Conduct in good
+faith may face temporary or permanent repercussions as determined by other
+members of the project's leadership.
+
+## Attribution
+
+This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
+available at https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
+
+[homepage]: https://www.contributor-covenant.org
+


Eearslya (approved):

theunkn0wn1 (approved):
LGTM

314numberpi:
[Question] Would the correct email not be ops?

theunkn0wn1:
For issues arising from contribution guidelines & behavior on the repository,Is it notwithin the scope of the Project Manager (me) to handle such events? If the issue is sufficiently severe, I could escalate to the ops team. 

@kenneaal @UncleClapton Whats the ops team stance on this?

kenneaal:
We actually discussed this in the ops time a while ago. When it comes to internal issues in the Tech Rats, we're keeping arbitration of such matters to ops who are also tech rats, since technical context is important in such matters.
By and large, project leads should seek to resolve any conflicts relating to their specific project, but may at any time request help or arbitration from ops. If project members feel that they have issues relating to a project lead that they don't want to bring up with them directly, they're free to contact us as well.

By and large, I hope that the need for a repo CoC is related to any external actors that may have interest in the project rather than internally in the tech team.

314numberpi (approved):
Looks good, with the response to my question

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
```
theunkn0wn1, line 53: Just for the record, yeah this is too board for our purposes. i think @UncleClapton is doing something about it?

theunkn0wn1 (changes requested):

CODE_OF_CONDUCT.md
UncleClapton: If I had to pick out what I disagree with **MOST**, it would be this line:

    Representation of a project may be further defined and clarified by project maintainers.

This opens things up for broad interpretation where broad interpretation **holds no place**. If we must adjust scope to fit something we're missing we can, but allowing individual maintainers to redefine where this CoC applies at-will is improper.

Since this CoC is intended for developers of Fuel Rats applications and services, a scope this wide is really overkill. While I certainly expect Crunchy and Xlexious to follow the idea of this CoC when tweeting (squeaking?) via our official Twitter, I don't think this specific CoC should bind their actions. The way it is currently worded, any activity from @fuelrats would be included, not just when we're speaking as techs. This is just one example of how the current scope is way too broad.

If I were to reword this to better fit our needs.. it might say something like... 

```
This Code of Conduct applies to everyone within the organization's project spaces,
and any individual officially representing the organization or project in public spaces.
```

However this alone is not enough as we need to further define the limit of what _"representing the organization"_ means. Otherwise it may be seen as _"as a maintainer you always represent the org and project."_ instead of _"as a maintainer you represent the org when speaking to users of the project, about the project"_.

Making that definition is something I don't feel comfortable fully defining on my own, and if we cannot get a stable definition then this CoC should be scoped back to just communication within projects. Personally I think it should be JUST the org and projects, but I'm willing to give applying a good expanded definition a shot if we can get good suggestions.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+  advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
```
UncleClapton, line 31: This is highly up to personal interpretation IMO. Whats acceptable in my workplace may not be acceptable in yours. 

I realize I might be a bit pedantic about this one. I completely understand what the writers of the Contributor Covenant CoC are going for here, but it's the "walking the fine line" situations that may make some question rather this would apply to what they're saying, and I don't want to stifle conversation as a result.

In my experience.. if people question if what they're saying is acceptable (and I mean actually question it), they're more apt to just not say anything at all instead of asking for clarification. I'm not saying all people are like this.. but there are quite a few.

For that reason I'm very opposed to overly-broad and easily misinterpreted wording such as this.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
```
UncleClapton, line 42: Unfortunately this text does not work well with the split jurisdiction we're holding on org repos. This will probably have to get re-worded eventually to specify what maintainers are responsible for and what tech-mods are responsible for.

Lets hold that off until we're sure how we're moving forward with the adoption of this CoC, though.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project lead at unknown@fuelrats.com. All
```
UncleClapton, line 58: This is something we're gonna have to re-address as well.

I think actual reports of abusive behavior coming from outside collaborators should go to the "tech rat moderators" (really don't have a good name for those who exist in the overlap), as their complains are more likely to be of someone from within the org.

We also need to make a final decision on how bans of those outside of the tech rats will work... I believe disagreements and conflicts should be settled with conversation whenever possible (I feel like CMDR Rhea advocating against discord's partnership CoC :P), but if a project lead thinks that a ban is **absolutely necessary** then the case should be handed over to the "techs who are also mods" group.

I think this step is necessary because anyone who has broken the CoC that badly here, should also be considered for an IRC ban or FR account suspension. This cannot happen if us mods can't investigate and discuss the right course of action on our own.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
```
UncleClapton, line 8: 💢 This list doesn't include political views. Why does it not include political views?

UncleClapton (changes requested):
So the idea I proposed way back when this started was if we all (as in us tech rats) found it suitable, we should apply this to the whole org instead of just this project. If we go down that route, then this PR should be closed, and the text transferred to a separate repo. Probably [tech-rat-code-standards](https://github.com/FuelRats/tech-rat-code-standards), though we will rename that repo before this is put into effect.

In this specific document's place, A new file should be made containing a link to the org-wide CoC, and quick contact links for complaints and questions.

I think some additional discussion on that point should be made before we close this PR, however.

Finally, I realize a lot of my complaints in this review stem from my general feelings of this particular CoC template. Let me be clear when I say I ***HATE*** the Contributor Covenant. Up front it looks very solid and takes hard stances on whats acceptable, but when read into it's clear that as-is it leaves gaping holes for interpretation to reject contributions for anything the maintainers might disagree with, while giving the appearance that they're pro acceptance and pro free speech. What sucks is that it's easily fixable, but often just left as-is with the glaring holes in tact...

shatteredbeam:
Closed.  This discussion is being moved to an org-wide level to be discussed with that scope in mind.

PR #129, opened 2019-03-26
[SPARK-124] Add a retry handler to Galaxy's _call

Eearslya:
The Internet is a weird and mysterious place. Things fail. Servers fall over. Sometimes these failures are brief. Sometimes they only last a second or two.

This PR gives Galaxy the ability to retry a failed API call (timeout, 404, 500, etc.) up to three times before it will consider it a permanent failure and raise the red flag.

### DIFF
--- a/src/packages/galaxy/galaxy.py
+++ b/src/packages/galaxy/galaxy.py
@@ -13,7 +13,7 @@ See LICENSE.md
 import asyncio
 from html import escape
 import json
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Union
 
 import aiohttp
 
@@ -66,9 +66,6 @@ class Galaxy:
 
     TIMEOUT = aiohttp.ClientTimeout(total=10)
     "A ClientTimeout object representing the total time an HTTP request can take before failing."
-
-    RETRY_SLEEP = 1
-    "The amount of time, in seconds, to wait between HTTP retries."
 
     def __init__(self, url: str = None):
         self.url = url or config['api']['url']
@@ -281,7 +278,16 @@ class Galaxy:
         if nearest['data']:
             return [neighbor['name'] for neighbor in nearest['data']]
 
-    async def _call(self, endpoint: str, params: Dict[str, str] = None) -> object:
+    async def _retry_delay(self, current_retry: int) -> None:
+        """
+        Uses asyncio.sleep to pause execution for a number of seconds equal to
+        `current_retry` squared.
+        """
+        await asyncio.sleep(current_retry ** 2)
+
+    async def _call(self,
+                    endpoint: str,
+                    params: Optional[Dict[str, str]] = None) -> Union[dict, list]:
         """
         Perform an API call on the Fuel Rats Systems API.
 
@@ -290,7 +296,7 @@ class Galaxy:
             params (Dict): A dictionary of key-value pairs that will make up the query string.
 
         Returns:
-            An object representing the parsed JSON data returned from the API endpoint.
+            A dict or list object representing the parsed JSON data returned from the API endpoint.
         """
 
         base_url = self.url
@@ -312,4 +318,4 @@ class Galaxy:
                     raise
 
                 # Introduce a short pause between retries
-                await asyncio.sleep(self.RETRY_SLEEP)
+                await self._retry_delay(retry)
--- a/src/packages/galaxy/galaxy.py
+++ b/src/packages/galaxy/galaxy.py
@@ -14,7 +14,7 @@ from html import escape
 import json
 from typing import Dict, List, Optional, Tuple
 
-import aiohttp
+from aiohttp import ClientError, ClientSession, ClientTimeout
 
 from config import config
 from ..utils import Vector
@@ -59,6 +59,8 @@ class Galaxy:
         )
     }
     MAX_PLOT_DISTANCE = 20000
+    MAX_RETRIES = 3
+    TIMEOUT = ClientTimeout(total=10)
 
     def __init__(self, url: str = None):
         self.url = url or config['api']['url']
@@ -271,7 +273,7 @@ class Galaxy:
         if nearest['data']:
             return [neighbor['name'] for neighbor in nearest['data']]
 
-    async def _call(self, endpoint: str, params: Dict[str, str]) -> object:
+    async def _call(self, endpoint: str, params: Dict[str, str] = None) -> object:
         """
         Perform an API call on the Fuel Rats Systems API.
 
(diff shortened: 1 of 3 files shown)


src/packages/galaxy/galaxy.py
```diff
@@ -271,6 +273,14 @@ def __init__(self, url: str = None):
                 [f"{key}={escape(str(value))}" for key, value in params.items()]
             )
         url = f"{base_url}{endpoint}?{param_string}"
-        async with aiohttp.ClientSession() as session:
-            async with session.get(url) as response:
-                return json.loads(await response.text())
+        retry_count = 0
+        while True:
```
theunkn0wn1: [style] Dangerous usage of a while block. i strongly advise using a for loop in this circumstance, or add `retry_count` as the LCV for this loop. 

```py
for retry in range(MAX_ATTEMPTS)
```

src/packages/galaxy/galaxy.py
```diff
@@ -271,6 +273,14 @@ def __init__(self, url: str = None):
...
+            try:
+                async with ClientSession(raise_for_status=True,
+                                         timeout=self.TIMEOUT) as session:
+                    async with session.get(url) as response:
+                        return json.loads(await response.text())
+            except ClientError:
+                retry_count += 1
+                if retry_count > self.MAX_RETRIES:
+                    raise
```
theunkn0wn1, line 318: [suggestion] following on above comment, this logic can probably be simplified, consider refactoring.

tests/galaxy/galaxy_fx.py
```diff
@@ -95,6 +95,12 @@ def mock_system_api_server_fx():
             """{"data":[]}"""
         )
 
+        # Tests that Galaxy will retry failed requests
+        httpserver.expect_oneshot_request("/badendpoint").respond_with_data(status = 400)
+        httpserver.expect_oneshot_request("/badendpoint").respond_with_data(status = 500)
+        httpserver.expect_oneshot_request("/badendpoint").respond_with_data("""{"success":true}""")
```
theunkn0wn1: [question] how does this work, having a single endpoint that has multiple returns?

Eearslya: Each of those are defined as "oneshot" requests. Meaning that once `pytest_httpserver` has responded to it, it will remove that response from its list.

Since the test lib searches from the front to the back of the array, it will find the first oneshot request, use it to respond, then remove it. The next call will use the second response, and remove it. And so on and so forth.

theunkn0wn1: wierd but ok. thanks for explaining that.

src/packages/galaxy/galaxy.py
Eearslya: Unfortunately I couldn't simplify this very well while also keeping the functionality I want. Which is to say, re-raising the exception from `aiohttp`.

theunkn0wn1 (dismissed):
Test approval is a test. this is not an official approval

# DO NOT MERGE

src/packages/galaxy/galaxy.py
```diff
@@ -10,11 +10,12 @@
 See LICENSE.md
 """
 
+import asyncio
 from html import escape
 import json
 from typing import Dict, List, Optional, Tuple
 
-import aiohttp
+from aiohttp import ClientError, ClientSession, ClientTimeout
```
theunkn0wn1: [style] [non-blocking] For the sake of clarity on where these exceptions come from at a glance, i would prefer you to just import `aiohttp` here

src/packages/galaxy/galaxy.py
```diff
@@ -60,6 +61,15 @@ class Galaxy:
     }
     MAX_PLOT_DISTANCE = 20000
 
+    MAX_RETRIES = 3
+    "The maximum number of times to retry a failed HTTP request before failing permanently."
+
+    TIMEOUT = ClientTimeout(total=10)
+    "A ClientTimeout object representing the total time an HTTP request can take before failing."
+
+    RETRY_SLEEP = 1
```
theunkn0wn1: [suggestion] implement exponential back off? 1 second between re-attempts may not be long enough...

src/packages/galaxy/galaxy.py
```diff
@@ -60,6 +61,15 @@ class Galaxy:
     }
     MAX_PLOT_DISTANCE = 20000
 
+    MAX_RETRIES = 3
+    "The maximum number of times to retry a failed HTTP request before failing permanently."
+
+    TIMEOUT = ClientTimeout(total=10)
```
theunkn0wn1: [style] _blinks_ yeah id prefer `aiohttp.ClientTimeout`

I don't know what it is about these names, but i am getting very easily confused between these and our clients...

src/packages/galaxy/galaxy.py
theunkn0wn1: [style] please rename this to `RETRY_DELAY`, as it is a tad more descriptive of what this constant does, and is less likely to be confused with `time.sleep`

src/packages/galaxy/galaxy.py
```diff
@@ -271,7 +281,7 @@ def __init__(self, url: str = None):
         if nearest['data']:
             return [neighbor['name'] for neighbor in nearest['data']]
 
-    async def _call(self, endpoint: str, params: Dict[str, str]) -> object:
+    async def _call(self, endpoint: str, params: Dict[str, str] = None) -> object:
```
theunkn0wn1: [error] As `None` is now a valid value, the type hint must reflect this too.

```suggestion
    async def _call(self, endpoint: str, params: Optional[Dict[str, str]] = None) -> object:
```

theunkn0wn1: [error] looks like this returns a dictionary, not `object`
```suggestion
    async def _call(self, endpoint: str, params: Dict[str, str] = None) -> Dict:
```

theunkn0wn1 (changes requested):

src/packages/galaxy/galaxy.py
Eearslya: 704dd81

src/packages/galaxy/galaxy.py
Eearslya: 704dd81

src/packages/galaxy/galaxy.py
Eearslya: b4713e1

Eearslya: Variable was removed entirely in lieu of a helper method.

src/packages/galaxy/galaxy.py
Eearslya: b4713e1

Eearslya: b4713e1

shatteredbeam:
Is this ready for review now or whats the story?

theunkn0wn1:
based off the fact i was requested for review, @shatteredbeam , i think its safe for review.

Eearslya:
It's been ready since the feature freeze ended 😅

theunkn0wn1 (approved):
LGTM

PR #130, opened 2019-03-28
[SPARK-126] Add a Mecha3 smoke test

Eearslya:
This adds a simple test to the CircleCI suite that will spin up a full IRC/Services setup, give Mecha the chance to connect, then make sure that it did.

This ensures that the entire project as a whole is able to run. It does not yet ensure that the bot will respond to commands, but this will set the ground work for adding in such tests.


theunkn0wn1 (approved):
we will have to sort out the environment duplication issue later, as it doesn't block this PR.

LGTM

PR #132, opened 2019-04-02
First pass at using a constant executor definition

theunkn0wn1:


theunkn0wn1:
necessary tech on Circle's end is very unstable. We should revisit this in a few months after CCI v2.1 calms down a bit.

PR #133, opened 2019-04-02
[SPARK-133] Restructure test folder

Eearslya:


tests/unit/test_autocorrect.py
```diff
@@ -13,7 +13,7 @@
 
 from src.packages.utils import correct_system_name
 
-pytestmark = pytest.mark.autocorrect
+pytestmark = [pytest.mark.unit, pytest.mark.autocorrect]
```
theunkn0wn1, line 16: [question] can i get a documentation link for this syntax?

theunkn0wn1 (approved):
LGTM

tests/unit/test_autocorrect.py
Eearslya: https://docs.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules

>or if you need to use multiple markers you can use a list:
>```python
>import pytest
>
>class TestClass(object):
>    pytestmark = [pytest.mark.webtest, pytest.mark.slowtest]
>```

theunkn0wn1: awesome, thanks! wasn't aware test marks could be combined like this!

PR #135, opened 2019-04-10
SPARK-136 Part 1.

theunkn0wn1:
TL;DR The Ratmama test suite makes a bad practice of mutating read-only objects and has made any refactoring of unrelated systems extremely difficult.

Unfortunately it requires a rewrite, in the mean time I propose we simply skip the suite until @Eearslya finishes the rewrite.


Eearslya (approved):

shatteredbeam (approved):
LGTM

PR #136, opened 2019-04-10
User Dataclass

theunkn0wn1:
The `User` object happens to be a prime target for dataclass conversion

1. Its read-only, once an object is created it doesn't get modified at runtime.
 - its event scoped, disappearing as its containing context falls out of scope
2. As it is read only, its hashable
 - dataclass implements this for us, simplifying logic
3. Dataclass implements equality operations for us, simplifying logic
4. Doesn't have any advanced getter/setter logic

Bottom line, by converting `User` to a dataclass a bunch of magics get implemented for us, while greatly reducing testing requirements.


Eearslya (approved):

shatteredbeam (approved):
No concerns. LGTM.

PR #138, opened 2019-04-11
CodeCov reconfiguration

theunkn0wn1:
This will hopefully alleviate the repeat-comment issue we are having with CodeCov. ...

This also improves the message CodeCov gives us, with more information and pretty graphs!


theunkn0wn1:
Should be good to go now, I suppose we can't really tell until the next PR but i *think* i fixed the 4 message issue now.
I changed the message behavior from its default to message exactly once.

Its allowed to update its comment, but its not allowed to make more than one. (it also won't create a new comment if its comment was deleted).
hopefully this will be sufficient

Eearslya (approved):

shatteredbeam (approved):

PR #140, opened 2019-04-16
Spark 130 -- Cofiguration overhaul.

theunkn0wn1:
This pull request completley redoes how Mecha handles configurable settings.

# High level changes
This re-implementation uses [`pluggy`](https://github.com/pytest-dev/pluggy) as its core infrastructure.

1. This package is already in our development dependencies ( `Pytest` depends on it amongst other modules ), meaning its already present in our development environments.
2. `Pluggy` prevents us from re-inventing the wheel several times over, il get tot his more later.
3. `Pluggy` lays the foundation for **plugins**
4. `Pluggy` is extensible

Pluggy is designed to enable plugins, making this approach possibly the most extensible solution to the problem of event propagation.

## Existing dependency
- By using an existing dependency, We know it doesn't break anything when introduced.
- It is a well tested, well developed package made by the same people as our `pytest` test runner.

## Re-inventing the wheel and alternative implementations
I have considered other implementations for the configuration overhaul including but not limited to abstract / Base classes and decorators.

## objectives
 - [x] the new system must be extensible
    - Seems simple enough to extend with other event types
- [x] the new system should support plugins 
   - pluggy is *designed* for plugins
- [x] the new system must support reloading of the configuration file, aka `rehashing`
    - Implemented with literal file hashing to boot
- [x] the new system must support validating the configuration file
   - done through plugin events.

### subclasses
The issue with the usage of subclasses, a la Offline Awareness, is that  i did not find this approach to be sufficiently extensible. 

OA's implementation depends on a Base class that keeps shallow references to its children.
It does this by requiring children to invoke the base class's constructor via a super call.
```py
class Aware(OfflineAwareABC):
    async def on_online(self):
        ...

    async def on_offline(self):
         ...
    def __init__(self):
        ...
        super().__init__()
```

On its own, this works fine. However, once we consider the underlying interface, the problems become obvious in the context of extensibility.

1. it uses class-bound attributes
- By using class-bound attributes, out of necessity any subclass of this object will have these attributes too.
-  specialized subclasses would need their own unique storage object
2. machinery can't be reused between specialized event ABCs, **leading to duplication
 - each specialized event ABC would need to essentially copy/paste the machinery, due to the cross-ABC state.
 - reinvention of wheel.
3. cross-ABC state would lead to unnecessary, but also unavoidable cross-ABC coupling.

### Decorator based registration
This approach could work one of two ways.

1. acts as a class decorator, similar to `dataclasses.dataclass`, and does some magic as to mutate the class being registered.
 - this would be similar to the ABC approach, and suffer from similar issues
2.  Acts as a pass-through class decorator, similar to `@command`
 - Relatively simple to implement 
 - functions are hashable, so we can use a set for storage
 - instances  of objects cannot be stored this way
 - no ABC enforcement, would need to depend on introspection for sanity
 - could be built with native async support
 - potentially reinventing the wheel.

# `Pluggy` based approach
This approach is somewhat similar to the decorator based approach insofar as it also uses decorators, however *how* it uses them is for a completely different approach.

pluggy's interface is rather streight forward, and is made up of three parts.
1. the **plugin manager** 
2. the **specification**, 
3.  the **implementations** / **plugins**

## The Plugin manager
the plugin manager serves as a centralized instance of an object used to manage a plugin-enabled service.
Specifications and implementations are registered against this management object.

This Configuration implementation holds such an object as a static, module-bound object in `/src/config/__init__.py`

## Specifications
Specification functions are used to designate /what/ a plugin event looks like, this can be considered a prototype similar to those in c++ and similar languages.

a sample hook specification pulled from this feature `/src/config/_spec.py`
```py
rehash_spec = pluggy.HookspecMarker(_PLUGIN_NAME)

@rehash_spec
def rehash_handler(data: typing.Dict):
    """
    Apply new configuration data

    Args:
        data (typing.Dict): registered subsection of new configuration data to apply.

    """
    ...
```
 
Once hooks are defined, they can be registered against the plugin manager. 

This implementation does this directly after initializing the plugin manager in `/src/config/__init__.py`

## The Implementations AKA the good stuff AKA Usage
Plugins can take one of two forms, one can either bind a *namespace* such as a Module or an instance of an object. 

**Note** plugins need not be objects, namespaces work too! this is an improvement over the ABC approach discussed above

To register as a plugin, the object / namespace needs to be registered against the plugin manager.
this implementation exposes both the plugin manager and the necessary plugin marker for external use.

to register a plugin one simply needs to decorate the functions / methods with the exported `config_marker` decorator, then register the object itself against the manager.
```py
from src.config import config_marker, plugin_manager
class MyConfigAwareClass:
    @config_marker
    def validate_config(self, data: Dict):
        print(f"in {self}.validate_config")

    @config_marker
    def rehash_handler(self, data: Dict):
        print(f"in {self}.rehash_handler")

# make an instance 
instance = MyConfigAwareClass()

# register the plugin
plugin_manager.register(instance, "OptionalPluginName")
```

a module can be used in place of `instance`, when registering modules as plugins.
```py
from . import rat_command
...
from src.config import plugin_manager

plugin_manager.register(rat_command, "commands")
```

## limitations of this approach
1. Currently pluggy only supports synchronous plugins, and does not support an async mode of operation.

2. While pluggy works with both, there is a word of caution to be had, Pluggy's plugin-detection mechanism tends to call `getattr` on every attribute of the object, meaning something like the `Mechaclient`'s lazy loaded attributes don't play well with it.

3. we canot have multiple plugins with the same name (or hash)

# DL; DR (but please do) 
This PR adds the capability to validate and reload the configuration file without bouncing the Mechaclient (for the most part).

### Validator
This event may be used by plugins to receive the candidate configuration object for validation. 
Each plugin gets the same, full, configuration dictionary.

Plugins are responsible for validating any portion of the configuration object they directly depend on, though they are allowed to validate other portions as they see fit. 

Plugins may raise either a `ValueError` or a `KeyError`, though any exception will do. (but please keep to these two, so i don't need to introduce a second wildcard exception handler 😷 )

```py
@validator_spec
def validate_config(data: typing.Dict):
    """
    Validate new configuration data.

    Args:
        data (typing.Dict): new configuration data  to validate

    Raises:
        ValueError: config section failed to validate.
    """
    ...
```

### Applicator
This event may be used by plugins to receive the new configuration object for application internally.

Each plugin gets the same, full, configuration dictionary.

Each plugin is responsible for applying any changes to its underlying data structures.

```py
@rehash_spec
def rehash_handler(data: typing.Dict):
    """
    Apply new configuration data

    Args:
        data (typing.Dict): new configuration data to apply.

    """
    ...
```

## Flow of control
the project will **no longer** load the configuration file upon import, instead differing this to the program's entry point code.

This means the configuration file is **unavailable without using the plugin hooks**.
Further, this means the config file is **unavailable until post-init** so should not be depended upon as a module / plugin loads.


Dockerfile
```diff
@@ -13,7 +13,7 @@
 
 
 # Use an official Python runtime as a parent image
-FROM python:3.6.6-alpine
+FROM python:3.7-alpine
```
Eearslya: [Scope] Isn't the whole Docker config on the chopping block?

config.py
```diff
@@ -99,30 +101,55 @@ def setup_logging(logfile: str):
...
-    """
-    Sets up the module by loading the specified configuration file from disk
+_hash: str = ''
+
 
+def setup(filename: str) -> Dict:
+    """
+    Validate & apply configuration from disk
     Args:
         filename (str): path and filename to load.
```
Eearslya: [Style] No return documentation.

config.py
```diff
@@ -99,30 +101,55 @@ def setup_logging(logfile: str):
...
     """
-    global config
+    # do the loading part
+    config_dict, file_hash = load_config(filename)
+    global _hash
+
+    if file_hash == _hash:
+        raise FileExistsError(f"new file's hash matches what we already have!"
```
Eearslya: [Question] Why is this a fatal condition?
If we're attempting to reload an unchanged config file, shouldn't it simply return without doing anything?

config.py
```diff
@@ -99,30 +101,55 @@ def setup_logging(logfile: str):
...
+                              f"\told = {_hash}\tnew={file_hash}")
```
Eearslya: [Error][Not Blocking] If the condition is that these two are equal, isn't it a bit redundant to log both of them?

config.py
```diff
@@ -99,30 +101,55 @@ def setup_logging(logfile: str):
...
+    plugin_manager.hook.validate_config(
+        data=config_dict)  # FIXME: this does nothing as it runs before plugins are loaded
+    logging.info("done verifying. config loaded without error.")
+
+    logging.info(f"emitting new configuration to plugins...")
+    plugin_manager.hook.rehash_handler(data=config_dict)
+    return config_dict
+
+
+def load_config(filename) -> Tuple[Dict, str]:
```
Eearslya: [Style] Missing docstring.

config.py
```diff
@@ -99,30 +101,55 @@ def setup_logging(logfile: str):
...
+    logging.debug(f"Found a file/directory at {path.resolve(strict=True)}'! attempting to load...")
+    with path.open('r', encoding="UTF8") as infile:
+        infile.flush()
+        config_dict = toml.load(infile)
+        logging.info("Successfully loaded from file specified!")
+
+    # update hasher with the raw bytes of the file
+    hasher.update(path.read_bytes())
+
+    del infile
```
Eearslya: [Question] Is this not handled automatically by the `with` block?

src/__main__.py
```diff
@@ -15,9 +15,11 @@
 import asyncio
 import logging
 
-from config import config
+from config import setup
 # noinspection PyUnresolvedReferences
+from src import commands
```
Eearslya: [Style] `[W0611(unused-import), ] Unused commands imported from src`

src/commands/debug.py
```diff
@@ -55,3 +58,33 @@
     """
 
     await context.reply("pong!")
+
+
+@command("getPlugins")
+@require_channel
+@require_permission(TECHRAT)
+async def cmd_get_plugins(context: Context):
```
Eearslya, line 64: [Style] Missing docstring.

src/commands/debug.py
```diff
@@ -55,3 +58,33 @@
...
+    LOG.warning(f"config rehashing invoked by user {context.user.nickname}")
+    try:
+        path = cli_manager.args().config_file
+        await context.reply(f"using config file {path}...")
+        setup(path)
+    except (KeyError, ValueError) as exc:
+        await context.reply(f"unable to rehash configuration file.")
+        raise ValueError("failed to rehash configuration") from exc
+    except FileExistsError:
+        await context.reply(f"unable to rehash configuration, you sure it changed?")
```
Eearslya: [Suggestion] Following my comment above, I'm not sure this should be an error case.

src/config/__init__.py
```diff
@@ -0,0 +1,36 @@
+"""
+__init__.py -   # FIXME teplates
```
Eearslya: [Style] FIXME
(I know you probably already know this already, but I'm just documenting it so it isn't forgotten.)

src/config/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+import pluggy
+
+# load our hook specification
+from ._constants import _PLUGIN_NAME
+from . import _spec
+
+# create the plugin manager
+plugin_manager = pluggy.PluginManager(_PLUGIN_NAME)
```
Eearslya: [Style] `[C0103(invalid-name), ] Constant name "plugin_manager" doesn't conform to UPPER_CASE naming style`
(Either rename or add an ignore?)

src/config/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+Configuration plugin manager
+"""
+
+# register our specifications
+plugin_manager.add_hookspecs(_spec)
+
+# create implementation markers
+
+config_marker = pluggy.HookimplMarker(_PLUGIN_NAME)
```
Eearslya: [Style] `[C0103(invalid-name), ] Constant name "config_marker" doesn't conform to UPPER_CASE naming style`
(Either rename or add an ignore?)

src/config/_constants.py
```diff
@@ -0,0 +1,13 @@
+"""
+_constants -   # FIXME teplates
```
Eearslya: [Style] FIXME

src/config/_spec.py
```diff
@@ -0,0 +1,46 @@
+"""
+_validator -   # FIXME teplates
```
Eearslya: [Style] FIXME

src/config/_constants.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+Copyright (c) 2019 The Fuel Rats Mischief,
+All rights reserved.
+
+Licensed under the BSD 3-Clause License.
+
+See LICENSE.md
+"""
+_PLUGIN_NAME = "mecha_config"
```
Eearslya, line 11: [Question] Is there a reason this is labeled a private variable?
It seems to be the exact kind of thing this module *wants* to export.

src/config/_spec.py
```diff
@@ -0,0 +1,46 @@
...
+
+See LICENSE.md
+"""
+import typing
+
+import pluggy
+
+from ._constants import _PLUGIN_NAME
+
+validator_spec = pluggy.HookspecMarker(_PLUGIN_NAME)
```
Eearslya: [Style] `[C0103(invalid-name), ] Constant name "validator_spec" doesn't conform to UPPER_CASE naming style`

src/config/_spec.py
```diff
@@ -0,0 +1,46 @@
...
+rehash_spec = pluggy.HookspecMarker(_PLUGIN_NAME)
```
Eearslya: [Style] `[C0103(invalid-name), ] Constant name "rehash_spec" doesn't conform to UPPER_CASE naming style`

src/mechaclient.py
```diff
@@ -12,11 +12,12 @@
 
 """
 from logging import getLogger
+from typing import Dict, ClassVar
```
Eearslya: [Style] `[W0611(unused-import), ] Unused Dict imported from typing`
`[W0611(unused-import), ] Unused ClassVar imported from typing`

src/mechaclient.py
```diff
@@ -12,11 +12,12 @@
...
 from uuid import uuid4
 
 from pydle import Client
 
-from config import config
+from src.config import config_marker
```
Eearslya: [Style] `[W0611(unused-import), ] Unused config_marker imported from src.config`

src/packages/commands/rat_command.py
```diff
@@ -43,7 +43,31 @@ class NameCollisionException(CommandException):
 _registered_commands = {}  # pylint: disable=invalid-name
 
 # character/s that must prefix a message for it to be parsed as a command.
-PREFIX = config['commands']['prefix']
+PREFIX: str
+
+
+@config_marker
+def rehash_handler(data: Dict):
```
Eearslya: [Style] Missing docstring.

src/packages/commands/rat_command.py
```diff
@@ -43,7 +43,31 @@ class NameCollisionException(CommandException):
...
+        data (Dict): new config object
+    """
+    global PREFIX
+
+    PREFIX = data['commands']['prefix']
+    LOG.debug(f"in rehash, applying new configuration data...")
+
+
+@config_marker
+def validate_config(data: Dict):
```
Eearslya: [Style] Missing docstring.

src/packages/commands/rat_command.py
```diff
@@ -43,7 +43,31 @@ class NameCollisionException(CommandException):
...
+    """
+    handle rehash events
+
+    Args:
+        data (Dict): new config object
+    """
+    global PREFIX
+
+    PREFIX = data['commands']['prefix']
+    LOG.debug(f"in rehash, applying new configuration data...")
```
Eearslya: [Error] Debug log seems a little unhelpful without any context as to WHICH rehash it's in.

src/packages/context/context.py
```diff
@@ -10,22 +10,38 @@
...
 """
-from typing import Optional, Tuple, List, TYPE_CHECKING
+import typing
+from typing import Optional, Tuple, List, TYPE_CHECKING, Dict
 
-from config import config
+from src.config import config_marker
 from ..user import User
 
+import logging
```
Eearslya: [Style] `[C0411(wrong-import-order), ] standard import "import logging" should be placed before "from src.config import config_marker"`

src/packages/context/context.py
```diff
@@ -10,22 +10,38 @@
 
 See LICENSE.md
 """
-from typing import Optional, Tuple, List, TYPE_CHECKING
+import typing
```
Eearslya, line 16: [Question] Why is this imported wholesale instead of simply adding `ClassVar` to the imports below?

src/packages/context/context.py
```diff
@@ -10,22 +10,38 @@
...
+import logging
+
+LOG = logging.getLogger(f"mecha.{__name__}")
 if TYPE_CHECKING:
     from src.mechaclient import MechaClient
 
-PREFIX = config['commands']['prefix']
+
+@config_marker
+def validate_config(data: Dict):
```
Eearslya: [Style] Missing docstring.

src/packages/context/context.py
```diff
@@ -10,22 +10,38 @@
...
+    if not isinstance(data['commands']['prefix'], str):
+        raise ValueError
+
+
+@config_marker
+def rehash_handler(data: Dict):
```
Eearslya: [Style] Missing docstring.

src/packages/database/database_manager.py
```diff
@@ -52,6 +52,53 @@ class property, and the DatabaseManage.query() method used to perform a query.
...
+            data (typing.Dict): new configuration data to apply.
+
+        """
+        cls._config = data
+
+    _config: ClassVar[Dict] = {}
+
+    @classmethod
+    @config_marker
+    def validate_config(cls, data: Dict):
```
Eearslya: [Style] Missing docstring.

src/packages/database/database_manager.py
```diff
@@ -135,3 +183,6 @@ def __init__(self,
         self._dbpool.putconn(connection)
 
         return list(result)
+
+
+plugin_manager.register(DatabaseManager, "Database")
```
Eearslya: [Question] For consistency's sake, most modules seem to be registering through their `__init__.py` files. Should we stick to that standard?

src/packages/galaxy/galaxy.py
```diff
@@ -319,3 +335,6 @@ def __init__(self, url: str = None):
 
                 # Introduce a short pause between retries
                 await self._retry_delay(retry)
+
+
+plugin_manager.register(Galaxy, "galaxy")
```
Eearslya: [Question] Again, the question of consistency and where modules should register.

src/packages/permissions/permissions.py
```diff
@@ -16,18 +16,34 @@
...
 
-from config import config
+from src.config import config_marker
 from ..context import Context
 
 LOG = logging.getLogger(f"mecha.{__name__}")
 
 
+@config_marker
+def validate_config(data: Dict):
```
Eearslya, line 26: [Style] Missing docstring.

src/packages/permissions/permissions.py
```diff
@@ -16,18 +16,34 @@
...
+    ...
+
+
+@config_marker
+def rehash_handler(data: Dict):
```
Eearslya, line 61: [Style] Missing docstring.

src/packages/permissions/permissions.py
```diff
@@ -16,18 +16,34 @@
...
-from config import config
+from src.config import config_marker
 from ..context import Context
 
 LOG = logging.getLogger(f"mecha.{__name__}")
 
 
+@config_marker
+def validate_config(data: Dict):
+    ...
```
Eearslya: [Question] Is this a to-do, or is there really no validation to be done?

src/packages/ratmama/ratmama_parser.py
```diff
@@ -13,16 +13,32 @@
 
 import logging
 import re
-from typing import Optional
+from typing import Optional, Dict
 
 import config
```
Eearslya: [Style] `[W0611(unused-import), ] Unused import config`

tests/conftest.py
```diff
@@ -319,3 +320,32 @@ def test_fact_fx() -> Fact:
                 mfd=False,
                 edited=None
                 )
+
+
+@pytest.fixture(scope="session", autouse=True)
+def global_init_fx():
```
Eearslya: [Style] Missing docstring.

tests/unit/test_permissions.py
```diff
@@ -19,16 +19,17 @@
 import pytest
 
 import src.packages.commands.rat_command as Commands
+from config import setup
+from src.packages.cli_manager import cli_manager
```
Eearslya: [Style] Both of these appear to be unused imports?

Eearslya (commented):
I like the implementation, mostly a lot of style violations and needed documentation.

Of course there's plenty of testing to add as well, but you've already acknowledged that.
I'd also like to generally take a look at the grammar being used within documentation and logging. Should we care about proper capitalization and structuring sentences?

Dockerfile
theunkn0wn1: that shouldn't have made it in, il need to hunt down whatever commit touched that.

config.py
theunkn0wn1: I found it very useful to have both, even if logically speaking logging both is redundant.

config.py
theunkn0wn1: refering to the `del`? no.

the `with` block simply ensures the file is closed, but it does not remove the name nor the file handle.

src/commands/debug.py
theunkn0wn1: in debug module, OOS to review :P

src/config/__init__.py
theunkn0wn1: It hasn't been forgotten, its one of the reasons this PR is in a draft state 😉

src/config/__init__.py
theunkn0wn1: hmm.... this one may require discussion as it is not constant.

yes, the name is constant insofar as it always points to the same object, but the object to which it points to is not constant (due to registration activities)

src/config/__init__.py
theunkn0wn1: yeah, this should be named constant. will fix.

src/config/_constants.py
theunkn0wn1: no, external plugins shouldn't need to access the plugins name, further it is intended to only be used internally by the package's modules.

src/config/_spec.py
theunkn0wn1: *grumbles* will fix

src/packages/commands/rat_command.py
theunkn0wn1: context is added by the logger itself. reviewing the log file its quite clear which module emitted this message.

src/packages/permissions/permissions.py
theunkn0wn1: TODO, missing relevant `#TODO` comment

tests/unit/test_permissions.py
theunkn0wn1: quite possibly. this Pr is still a draft so imports have yet to be optimized.

Dockerfile
Eearslya: cff1264

`@theunkn0wn1 SPARK-130 begun adding Commands plugin stuff`

src/commands/debug.py
Eearslya: To be fair, so is the rehash command, and that's definitely in scope. :smile:

src/config/__init__.py
Eearslya: I feel like as long as the value of this particular variable (i.e. the *pointer* to what object this variable references) doesn't change, that it's still technically a constant.
It doesn't necessarily mean that absolutely nothing about it can ever change.

If we were to destroy and re-create it, changing the object ID, *then* it would be a problem.

src/config/_constants.py
Eearslya: Still feels odd to have an entire file dedicated to one constant value, and importing private vars is always a bit of a code smell imo.

Dockerfile
theunkn0wn1: resolved via an interactive rebase. thanks @Eearslya

config.py
theunkn0wn1: This is the only effective way to assert that we actually loaded a *new* configuration from disk, and it is a error state to reload the same configuration state. 

My reasoning here is the code handling the rehash itself would be expecting this exception to be raised and catch it, returning a well formatted error to the end users.

Eearslya: I still find it a bit odd to use an exception for this use case. In my mind, if a function wanted to check if a value had changed, it would save a copy before and verify afterwards. Exceptions usually mean something has gone wrong. Arguably, reloading the same config should not be an error state *for this specific method*.

theunkn0wn1: given the hastles i had when initially testing this implementation, it *was* arguably an error in this method to load the same file twice.

src/config/__init__.py
theunkn0wn1: renamed.

src/__main__.py
```diff
@@ -15,9 +15,11 @@
 import asyncio
 import logging
 
-from config import config
 # noinspection PyUnresolvedReferences
+from src import commands  # pylint: disable=unused-import
```
Eearslya, line 19: [Question] If this import truly is unused, why suppress the linting error?

src/config/_manager.py
```diff
@@ -0,0 +1,32 @@
+"""
+$file.fileName - $shortDescripion  # FIXME template
```
Eearslya: [Style] FIXME

src/config/_manager.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+
+#
+#  Copyright (c) $DateInfo.year The Fuel Rats Mischief,
+#  All rights reserved.
+#
+#  Licensed under the BSD 3-Clause License.
+#
+#  See LICENSE.md
+#
```
Eearslya: [Error] Two styles of file docstring.

src/config/_parser.py
```diff
@@ -99,30 +107,76 @@ def setup_logging(logfile: str):
...
+    global _hash  # pylint: disable=invalid-name, global-statement
+
+    if file_hash == _hash:
+        raise FileExistsError(f"new file's hash matches what we already have!"
+                              f"\told = {_hash}\tnew={file_hash}")
+
+    _hash = file_hash
+
+    setup_logging(config_dict['logging']['log_file'])
+    logging.info(f"new config hash is {file_hash}")
```
Eearslya, line 159: [Suggestion] Demote to debug level logging?

src/packages/cli_manager/cli_manager.py
```diff
@@ -2,7 +2,7 @@
 cli_manager.py - Manage CLI arguments
 
 This module will automatically parse arguments provided from the command line when this module is
-**first imported** and provide it for importing by other modules via the `args` attribute
+**first imported** and provide it for importing by other modules via the `GET_ARGUMENTS` attribute
```
Eearslya: [Error] `GET_ARGUMENTS` is not an attribute, it is a callable.

src/packages/context/context.py
```diff
@@ -10,22 +10,38 @@
...
+
+LOG = logging.getLogger(f"mecha.{__name__}")
 if TYPE_CHECKING:
     from src.mechaclient import MechaClient
 
-PREFIX = config['commands']['prefix']
+
+@CONFIG_MARKER
+def validate_config(data: Dict):
+    if not isinstance(data['commands']['prefix'], str):
```
Eearslya, line 34: [Question] This value is validated twice, in Context and Rat_Command. Is that intentional?

src/packages/commands/rat_command.py
```diff
@@ -43,7 +43,37 @@ class NameCollisionException(CommandException):
 _registered_commands = {}  # pylint: disable=invalid-name
 
 # character/s that must prefix a message for it to be parsed as a command.
-PREFIX = config['commands']['prefix']
+PREFIX: str
```
Eearslya: [Question] If prefix handling is done in Context, why does RatCommand need its own?

src/packages/database/database_manager.py
```diff
@@ -52,6 +52,53 @@ class property, and the DatabaseManage.query() method used to perform a query.
...
+        """
+        Apply new configuration data
+
+        Args:
+            data (typing.Dict): new configuration data to apply.
+
+        """
+        cls._config = data
+
+    _config: ClassVar[Dict] = {}
```
Eearslya: [Style] [Non-Blocking] Wouldn't this be better placed at the top of the class?

src/config/_manager.py
theunkn0wn1: I argue that a `fixme` is error. will fix

src/__main__.py
theunkn0wn1: its not unused. 

What happens here is `src.commands` package simply *needs to be loaded*, but its contents are not *directly* used.

This package contains our user-facing Interface, the commands system. Commands are registered at load time, but that **requires them to be loaded**

This seems like the best place to do this, is at the top of main as part of pre-initialization. 

The import needs to exist, even if its not used by this module directly.

src/config/_parser.py
theunkn0wn1: Im thinking it would be best to emit the fact a rehash occured, as well as the destination hash, as part of normal operation.

Perhaps the more verbose elements (such as the individual components handling things) would be debug level details.

src/packages/cli_manager/cli_manager.py
theunkn0wn1: not strictly true. it is an attribute as it is a member of a (module) object.

That said, the more apt description of this object is now `method` and will be adjusted

src/packages/commands/rat_command.py
theunkn0wn1: curiously enough, its used (outside configuration validation) in exactly one place within the module, and it doesn't even need to be there.

```py
                LOG.debug(f"Could not find command or rule for {PREFIX}{ctx.words[0]}.")
```

Il consider removing it.

src/config/_parser.py
Eearslya: I agree that rehashes should be logged, but I just think that outputting the actual SHA256 is a bit overkill for normal operation.

src/packages/context/context.py
theunkn0wn1: There is no restriction on what plugin may validate what portion of a configuration file.

src/packages/cli_manager/cli_manager.py
theunkn0wn1: Resolved in last push.

src/packages/context/context.py
Eearslya: I'm not saying there should be, I just find it odd that it's validated **twice**. Seems like a good place for a "changed-in-one-place-but-not-the-other" error to appear.

src/config/_parser.py
theunkn0wn1: No kill quite like the overkill.

I would rather know the exact SHA256 of what it applied without bouncing mecha into verbose mode to find out.

src/packages/context/context.py
theunkn0wn1: Ahh, I misunderstood.

Fair point, il look at it.

src/config/_constants.py
theunkn0wn1: this name is required internally by other modules, to prevent duplicity i needed to place it somewhere the other internals could access. The alternative here is to duplicate the hardcoded string across multiple modules.

I don't see a reason in which something external to this plugin system needs to access this name.

src/packages/context/context.py
theunkn0wn1: In this instance, we ended up importing an excessive number of names from this module.

Stylistically it is better after about 5 names to just import the module wholesale.

Further, I argue its more readable to have `typing.ClassVar` than `ClassVar` etc.

theunkn0wn1: I will fix the other type hints in this module to conform to the change as well, while im looking at the file.

src/packages/database/database_manager.py
theunkn0wn1: yes. will move.

src/packages/galaxy/galaxy.py
theunkn0wn1: Again, fair point will move.

theunkn0wn1:
Feature complete.

Just missing some tests and whatever you guys come up with in review

src/commands/debug.py
```diff
@@ -55,3 +57,34 @@
...
+    """ rehash the hash browns. (reloads config file)"""
+    LOG.warning(f"config rehashing invoked by user {context.user.nickname}")
+    try:
+        path = cli_manager.GET_ARGUMENTS().config_file
+        await context.reply(f"using config file {path}...")
+        setup(path)
+    except (KeyError, ValueError) as exc:
+        await context.reply(f"unable to rehash configuration file.")
+        raise ValueError("failed to rehash configuration") from exc
+    except FileExistsError:
```
shatteredbeam: [Suggestion]  Let's not raise during a rehash if there are no value changes.

src/config/_constants.py
shatteredbeam: I agree with Eearsy here, but I can't offer a better alternative.

src/mechaclient.py
```diff
@@ -49,16 +49,24 @@ def __init__(self, *args, **kwargs):
         self._fact_manager = None  # Instantiate Global Fact Manager
         self._rat_cache = None  # TODO: replace with ratcache once it exists
         self._rat_board = None  # Instantiate Rat Board
+        self._config = mecha_config if mecha_config else {}
         super().__init__(*args, **kwargs)
 
+    @property
```
shatteredbeam: [Question]  Do we need a public property here?  I would think config should remain private across the module, since each module has its own config object now.

src/packages/galaxy/galaxy.py
```diff
@@ -68,7 +84,7 @@ class Galaxy:
     "A ClientTimeout object representing the total time an HTTP request can take before failing."
 
     def __init__(self, url: str = None):
-        self.url = url or config['api']['url']
+        self.url = url or self.configuration['api']['url']
```
shatteredbeam: [Style] property name across modules should be consistent.

config.py
theunkn0wn1: Seems neither you nor @shatteredbeam like the exception, so il remove it.
Rehashing will now be state-agnostic.

shatteredbeam (changes requested):
Left in RC status due to missing tests.  Yes, I know they're forthcoming.

src/mechaclient.py
theunkn0wn1: No we do not, each module gets its own private property in this regard, and shouldn't be getting it from the mechaclient. will fix.

src/config/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+    .. hint::
+        For module-scoped plugins it is recommended to use the containing package's `__init__.py`
+        file to have this done automatically at load time
+
+    See Also:
+        `src.config <src >`_
+
+hello world!
```
Eearslya: [Error] A bit of test code left in? :smile:

src/__main__.py
```diff
@@ -41,6 +43,8 @@
     """
     Initializes and connects the client, then passes it to rat_command.
     """
+
+    config = setup(cli_manager.GET_ARGUMENTS().config_file)
```
theunkn0wn1: [question]  should i move `setup()` to something a tad more descriptive, like `reload_configuration` ?

src/commands/debug.py
theunkn0wn1: [error] this no longer raises, this except block should now be removed

src/config/_constants.py
theunkn0wn1: If you guys can come up with a better alternative, im all ears.

src/mechaclient.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, *args, **kwargs):
         """
         LOG.debug(f"Connecting to channels...")
         # join a channel
-        for channel in config["irc"]["channels"]:
+        for channel in self.config["irc"]["channels"]:
```
theunkn0wn1: [error] this should now point to `self._config` as the public `.config` property was removed.

src/packages/database/database_manager.py
theunkn0wn1: possibly, though it makes no difference functionally. 
Il see to moving it with other pending changes.

src/packages/galaxy/galaxy.py
```diff
@@ -11,21 +11,37 @@
 """
 
 import asyncio
-from html import escape
 import json
-from typing import Dict, List, Optional, Tuple, Union
+import logging
+from html import escape
+from typing import Dict, List, Optional, Tuple, Union, ClassVar
```
theunkn0wn1: [style] this should be moved to a wholesale typing import due to number of imports.

src/packages/galaxy/galaxy.py
```diff
@@ -11,21 +11,37 @@
...
+    @CONFIG_MARKER
+    def rehash_handler(cls, data: Dict):
+        """
+        Apply new configuration data
+
+        Args:
+            data (Dict): new configuration data to apply.
+
+        """
+        cls.config = data
```
theunkn0wn1: [error][style] this should be a private member

theunkn0wn1 (commented):
Did a cursory review over my own work looking for anything obvious that can be improved, found a few minor things. 

changes pending!

src/packages/database/database_manager.py
```diff
@@ -53,6 +53,59 @@ class property, and the DatabaseManage.query() method used to perform a query.
...
+            raise ValueError("No Database configuration in file.") from ex
+
+        # Require all values to be set
+        for setting in module_config.items():
+            if setting is None:
+                raise ValueError(f"{setting} is required for instantiation")
+
+        # Host
+        if not isinstance(module_config['host'], str):
+            raise ValueError("host must be a string.")
```
Eearslya: [Question] Perhaps these error messages should be a tad more specific? Unlike logging, we don't have any context here.
We could possibly have several config options by the name of `host` under different categories.

src/packages/database/database_manager.py
```diff
@@ -53,6 +53,59 @@ class property, and the DatabaseManage.query() method used to perform a query.
...
+
+        # Port
+        if not isinstance(module_config['port'], int):
+            raise ValueError("port must be an integer.")
```
Eearslya: [Question] Might it be better to keep a running list of errors and return them at the end? The fast-fail solution we have now means that users are only made aware of one config issue at a time.

src/packages/permissions/permissions.py
```diff
@@ -16,18 +16,51 @@
...
+
+
 class Permission:
     """
     A permission level
     """
+    __slots__ = ["level", "_vhosts", "_denied_message"]
 
-    def __init__(self, level: int, vhosts: Set[str],
+    def __init__(self, level: int = -1, vhosts: Set[str] = None,
```
Eearslya, line 83: [Style] I've always had it drilled into my head that magic numbers are bad. If `-1` is meant to mean something special, it should be assigned to a constant variable that can then be documented.

tests/unit/test_configuration.py
```diff
@@ -0,0 +1,10 @@
+"""
+
+"""
+from src.config import PLUGIN_MANAGER
+
+
+def test_rehash(configuration_fx):
+    PLUGIN_MANAGER.hook.rehash_handler(data=configuration_fx)
+def test_validate(configuration_fx):
```
Eearslya: [Style] Newline between method definitions.

src/config/__init__.py
theunkn0wn1: yep, thats a bit of test code. Sphinx & rst can be a bit finicky on occasions.

will fix!

src/packages/permissions/permissions.py
theunkn0wn1: Its not really a magic number, and i feel it requires no documentation.

Without specific initialization, all permission objects have a default permission level of -1, which is lower than any assigned value. 

I suppose this is a "fail safe" default, rather than a "fail secure".
Optimally, this value is immediately overwritten during post-init & during subsequent configuration file reloads.

My question here is should we "fail secure" and set this to be a very large integer value rather than a out-of-range small integer.

tests/unit/test_configuration.py
theunkn0wn1: this file wasn't supposed to be commited yet... sigh.

src/packages/database/database_manager.py
theunkn0wn1: As per [pluggy's documentation](https://pluggy.readthedocs.io/en/latest/#exception-handling), 
> If any hookimpl errors with an exception no further callbacks are invoked and the exception is packaged up and delivered to any Wrappers before being re-raised at the hook invocation point:

So holding for several errors in validation is not possible.

Or are you referring to within each module have it gather exceptions and raise one at the end?

src/packages/database/database_manager.py
theunkn0wn1: the erroring module will be in the stacktrace, though the most direct thing we can do here is have `__name__` embedded into the error strings or something.

Thoughts?

src/packages/context/context.py
theunkn0wn1: validations in rat command have been removed. issue resolved

src/commands/debug.py
```diff
@@ -55,3 +57,33 @@
...
+
+    plugins = PLUGIN_MANAGER.list_name_plugin()
+    names = [plugin[0] for plugin in plugins]
+    await context.reply(",".join(names))
+
+
+@command("rehash")
+@require_channel(message="please do this where everyone can see 😒")
+@require_permission(TECHRAT, override_message="no.")
+async def cmd_rehash(context: Context):
```
Eearslya: [Question] Should this really be among the debug commands?

As I understand it, the debug commands are the ones not meant to ever be used in production, but this one certainly would be, as it's half of the point of the PR.
Am I mis-understanding the purpose of this file?

src/config/_constants.py
Eearslya: Since this is indeed used in multiple modules, I can't come up with anything much better. I had thought it was used once and only once.

Eearslya (dismissed):
There are one or two aspects about this I'm not fully confident in yet, and documentation cleanup will need to be done in another pass.

But at the same time, if we sit here and polish out absolutely everything, this will never be merged. The code presented is functional and all of my concerns can easily be addressed without severe refactors later on.

For now, I'm giving the green light.

src/commands/debug.py
theunkn0wn1: you are misunderstanding the purpose of this file.

I  use it as a playground to experiment with commands-based functionality, not for final-product implementations.

That said, this command doesn't require much rework to be usable in production, so it proably should be moved out of this module

theunkn0wn1: moved the command to a new commands module for administration commands.
I foresee other similar commands being implemented in the future for administrative tasks so i am gunna just make a module for them now.

src/config/_constants.py
shatteredbeam: That being said, I'm not going to consider it blocking.  Perhaps we can improve this at a later date, but for now I don't see a reason to halt the PR (over this).

tests/unit/commands/test_administration.py
```diff
@@ -0,0 +1,65 @@
...
+
+    await administration.cmd_rehash(context_pm_fx)
+
+    assert "everyone can see" in bot_fx.sent_messages[0]['message']
+    assert not async_callable_fx.was_called
+
+
+@pytest.mark.asyncio
+async def test_rehash_not_authorized(bot_fx, async_callable_fx, monkeypatch):
+    context = await Context.from_message(bot_fx, "#unittest", "some_recruit", "!rehash 👿")
```
shatteredbeam, line 38: Hardcoded channel name?

tests/unit/commands/test_administration.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.mark.parametrize("exc_class", (ValueError, KeyError))
+@pytest.mark.asyncio
+async def test_rehash_application_failure(bot_fx, async_callable_fx, monkeypatch, exc_class,
+                                          context_channel_fx, caplog):
+    """
+    verifies the commands behavior when the setup() call fails (predictably).
+    """
+    context = await Context.from_message(bot_fx, "#unittest", "some_admin", "!rehash")
```
shatteredbeam, line 55: Another hard coded channel name.

shatteredbeam (changes requested):
Other than pointing out hardcoded channels in the test file, Most of this has already been discussed.

tests/unit/commands/test_administration.py
theunkn0wn1: This is within a test file, im not sure i see the issue with using a hardcoded channel name?

The channel is ultimately irrelevant, it simply needs to be a valid channel name.

Further, in many test files we make use of similar hardcoded strings, why are you calling these out in specific?

shatteredbeam (approved):

Eearslya (approved):

PR #141, opened 2019-04-16
Refactored `..context._split_message` to be less hacky

theunkn0wn1:
While our existing solution works, it depends on using exceptions to control program flow.
This, coupled with a `while True` loop, is a code smell and leaves much to be desired.

turns out that entire control structure can be implemented with a well written list comprehension, greatly saving in both complexity and code smell.

**Bonus feature!!** added a dedicated test to this routine

This PR does not cause any change in end-behavior, and does not modify any existing tests.


Eearslya (approved):
Voodoo magic. I feel like the parametrize is a bit aggressively indented, but it's not enough for me to deny. :+1:

shatteredbeam:
I second the phrase 'aggressively indented' but follow under the same guidelines.

shatteredbeam (approved):
See Comment.

PR #144, opened 2019-05-17
[Spark-47][SPARK-52] RatBoard rewrite

theunkn0wn1:
This pull request completely redoes the Rat Board.

# Rational for rewriting 
The old rat-board implementation is one of the oldest pieces of the codebase I committed, and the implementation simply did not last the test of time compared to newer approaches.

## contains complexity
For instance, the old implementation used a single keyed dictionary by client-name. The trouble with this stems from what has to be done against this container type: access by multiple indexes.

As a design requirement, we must be able to access a rescue contained in this collection by three key attributes:
1. client name (mutable)
2. board index (mutable)
3. API uuid (immutable)

The old implementation actually checks against 4 key attributes, adding the creation timestamp as a key attribute. Further, its `__contains__` implementation checks the **equality** of these attributes, which can be a potentially costly comparison. Its worst-case `__contains__` check would have a complexity of `O(4*(len(board))`, with many equality checks thrown in for good measure.
This contains routine can be very expensive, and inefficient.

The new implementation gets around this by using multiple parallel dictionaries, keyed by the above three key attributes.
the worst-case complexity of `__contains__` is reduced to `O(len(board[KEY_TYPE]))`, with no equality checks.
As the `client` and `board_index` fields may be nullable, their respective dictionaries may be smaller than the length of the collection as a whole, hence the distinction of `KEY_TYPE` in this complexity measurement.

Bottom line being the proposed implementation should be faster and more reliable in the regards of `__contain__`

### the old
The old implementation does not implement the `Mapping` interface, and does not support dictionary-style lookup.

The result of this being we ended up with **N identical methods** for accessing a contained rescue for each key type.
 - `board.find_by_uuid` for API uuids
 - `board.find_by_index` for board indexies
 - `board.find_by_name` for client names
 - `board.search` for any type
This introduces significant code duplicity, which reduces maintainability and increases likelihood of regressions. 

### the new
we need not these different, identical functions for lookup, especially when we have the different types already indexed. We can simply ask python to use its builtin dictionary access logic to do the heavy-lifting for us.

By implementing `__getitem__`, as required by the `Mapping` interface, we enable dictionary-style look-ups.
```py

In [1]: from src.packages.board import RatBoard

In [2]: board = RatBoard()

In [3]: async with board.create_rescue(client="fred") as rescue:
   ...:     ...

In [4]: board['fred']
Out[4]: <src.packages.rescue.rat_rescue.Rescue at 0x152d0d1c1d0>

In [5]: rescue
Out[5]: <src.packages.rescue.rat_rescue.Rescue at 0x152d0d1c1d0>

In [6]: board[rescue.board_index]
Out[6]: <src.packages.rescue.rat_rescue.Rescue at 0x152d0d1c1d0>

In [7]: board[rescue.api_id]
Out[7]: <src.packages.rescue.rat_rescue.Rescue at 0x152d0d1c1d0>
```
`__getitem__` will perform introspection to detect the key type, and return the value from the corresponding internal storage dictionary. 
- `__contains__` is implemented in a similar fashion.


theunkn0wn1:
Turns out i mixed up two related tickets here, this PR resolves both SPARK-52 and SPARK-47.

During my work towards SPARK-52 i hit a roadblock that required implementing SPARK-47. 😅

src/packages/board/board.py
```diff
@@ -0,0 +1,206 @@
...
+import itertools
+import logging
+import typing
+from collections import abc
+from contextlib import asynccontextmanager
+from uuid import UUID
+
+from src.packages.rescue import Rescue
+
+CYCLE_AT = 15
```
shatteredbeam: This should be a configurable value, not hard coded, especially if it isn't private.

src/packages/board/board.py
```diff
@@ -0,0 +1,206 @@
...
+    @asynccontextmanager
+    async def modify_rescue(self, key: _KEY_TYPE) -> Rescue:
+        """
+        Context manager to modify a Rescue
+
+        Args:
+            key ():
+
+        Returns:
+
```
shatteredbeam: Incomplete doc string.

shatteredbeam (changes requested):
two minor changes.  otherwise looks very good.

src/packages/board/board.py
theunkn0wn1: Hm, I don't really foresee a need to increase this limit in a configurable fashion, after all its meerly a hint for when mecha should attempt to start over indexing.

theunkn0wn1: made it configurable. in ebdf85a

PR #145, opened 2019-06-02
Point Circleci at the new announcer webhook

theunkn0wn1:
The Back end webhook for CI output has changed, this PR accommodates this.


kenneaal (approved):
Breaks ALL THE THINGS!

PR #146, opened 2019-06-15
[SPARK-141] Decoupled Rat from RatCache

theunkn0wn1:
This PR decouples Rat from RatCache
 this resolves a circular import, and is a precursor step to future refactoring/removal of the RatCache.


src/packages/cache/rat_cache.py
```diff
@@ -160,3 +161,10 @@ def flush(self) -> None:
         """
         self.by_name.clear()
         self.by_uuid.clear()
+
+    def append(self, rat: Rat):
+        if not rat.uuid or not rat.name:
+            raise ValueError(rat)
```
Eearslya, line 167: [Testing] Needs test for bad case

Eearslya (changes requested):

Eearslya (approved):

PR #147, opened 2019-06-28
[SPARK-144] Give MechaClient a persistent Galaxy object

Eearslya:


tests/unit/test_mechaclient.py
```diff
@@ -47,6 +48,13 @@ def test_verify_rat_board_private(bot_fx):
     assert bot_fx._rat_board is None
 
 
+def test_verify_galaxy_private(bot_fx):
+    """
+    Asserts the private reference for the Galaxy is none, after instantiation.
+    """
+    assert bot_fx._galaxy is None
```
theunkn0wn1: [error] this test is unnecessary, as it tests python's behavior rather than our code specifically.

theunkn0wn1 (changes requested):
one small testing related issue, otherwise LGTM

tests/unit/test_mechaclient.py
Eearslya: May need to eventually look at the other 4 tests in there as well. I was simply following their example.

theunkn0wn1: Might be wise. guess we need to schedule a review at some point 😕

theunkn0wn1 (approved):

PR #149, opened 2019-07-12
[SPARK-2] APIv3 Integration

Eearslya:
Working on adding MechaSqueak's APIv3 integration.

**This pull request is a work in progress, and is NOT ready to be merged!** It is here to receive early feedback from team members in order to weed out potential problems before they grow too large.

### DIFF
--- a/src/packages/api/api_manager.py
+++ b/src/packages/api/api_manager.py
@@ -106,12 +106,15 @@ class APIManager:
         Returns:
             A Dict with the requested change made, and all other elements left alone.
         """
+        if not path:
+            raise ValueError("No path given")
         paths = path.split(".")
         full_object = base_object
+        final_index = len(paths) - 1
         data = full_object
         for i, element in enumerate(paths, 0):
             # If this is the end of the path, update the object with the new value.
-            if i == (len(paths) - 1):
+            if i == final_index:
                 data[element] = value
             else:
                 # For nested dicts, initialize any dicts that don't already exist.
--- a/src/packages/api/api_manager.py
+++ b/src/packages/api/api_manager.py
@@ -305,7 +305,7 @@ class APIManager:
 
         try:
             data = await self._call(f"rescues/{uuid}", {'include': 'rats'})
-            if data['data']:
+            if data and 'data' in data:
                 return self._deserialize_rescue(data['data'], data['included'])
         except aiohttp.ClientError:
             return None
@@ -323,7 +323,7 @@ class APIManager:
             del data['relationships']['firstLimpet']
         jsonapi_data = {'data': data}
         new_rescue = await self._call('rescues', method='POST', body=jsonapi_data)
-        return APIManager._deserialize_rescue(new_rescue)
+        return APIManager._deserialize_rescue(new_rescue['data'])
 
     async def update_rescue(self, rescue: Rescue):
         """
@@ -332,6 +332,7 @@ class APIManager:
         Args:
             rescue (Rescue): The Rescue object to update in the API.
         """
+        # @TODO: Pass Set of changed attributes once they're available.
         data = APIManager._serialize_rescue(rescue)
         jsonapi_data = {'data': data}
         await self._call(f"rescues/{rescue.api_id}", method='PATCH', body=jsonapi_data)
@@ -418,9 +419,9 @@ class APIManager:
                 return json.loads(await self._perform_http(url, method, body))
             except aiohttp.ClientError:
                 # Even if it errors, do not retry POST or PATCH requests multiple times.
-                if retry == (self.MAX_RETRIES - 1) or method != 'GET':
+                if retry == self.MAX_RETRIES or method != 'GET':
                     raise
-                await asyncio.sleep(retry ** 2)
+                await asyncio.sleep((retry + 1) ** 2)
 
     async def _perform_http(self,
                             uri: str,
@@ -446,8 +447,7 @@ class APIManager:
             'Content-Type': 'application/vnd.api+json',
             'User-Agent': 'MechaSqueak/3.0.0a'  # @TODO Use actual MechaClient version
         }
-        async with aiohttp.ClientSession(raise_for_status=True,
-                                         timeout=self.TIMEOUT,
+        async with aiohttp.ClientSession(timeout=self.TIMEOUT,
                                          headers=http_headers) as session:
             if method == 'GET':
                 async with session.get(uri) as response:
--- a/src/packages/api/api_manager.py
+++ b/src/packages/api/api_manager.py
@@ -141,7 +141,8 @@ class APIManager:
                                     {'marked': rescue.marked_for_deletion.marked,
                                      'reason': rescue.marked_for_deletion.reason,
                                      'reporter': rescue.marked_for_deletion.reporter
-                                    }),
+                                     }
+                                    ),
             'rats': ('relationships.rats.data',
                      [{'type': 'rats', 'id': f"{rat.uuid}"} for rat in rescue.rats]),
             'first_limpet': ('relationships.firstLimpet.data',
@@ -248,7 +249,7 @@ class APIManager:
             created_at=APIManager.parse_api_datetime(attr['createdAt']),
             updated_at=APIManager.parse_api_datetime(attr['updatedAt'])
         )
-        #rescue.outcome = attr['outcome']  #@TODO Needs a setter or something...
+        # rescue.outcome = attr['outcome']  # @TODO Needs a setter or something...
         return rescue
 
     def __init__(self, config: typing.Optional[typing.Dict]):
@@ -372,7 +373,7 @@ class APIManager:
                     params: typing.Optional[typing.Dict[str, str]] = None,
                     method: typing.Optional[str] = "GET",
                     body: typing.Optional[typing.Dict] = None
-                   ) -> typing.Optional[typing.Union[dict, list]]:
+                    ) -> typing.Optional[typing.Union[dict, list]]:
         """
         Perform an API call on the Fuel Rats API.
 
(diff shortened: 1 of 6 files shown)


src/packages/api/api_manager.py
```diff
@@ -0,0 +1,442 @@
...
+                path = ''
+                value = None
+                # If the value is just a string, grab the value from the Rescue object.
+                if isinstance(mapping, str):
+                    path = mapping
+                    value = getattr(rescue, attrib)
+                # Otherwise, it's a tuple, and the value is the second element.
+                else:
+                    path = mapping[0]
+                    value = mapping[1]
```
theunkn0wn1:
```suggestion
                    path, value = mapping
```

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,442 @@
...
+                mapping = maps[attrib]
+                path = ''
+                value = None
+                # If the value is just a string, grab the value from the Rescue object.
+                if isinstance(mapping, str):
+                    path = mapping
+                    value = getattr(rescue, attrib)
+                # Otherwise, it's a tuple, and the value is the second element.
+                else:
+                    path = mapping[0]
```
theunkn0wn1:
```suggestion

```

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+    @classmethod
+    @CONFIG_MARKER
+    def validate_config(cls, data: typing.Dict):
+        """
+        Validate the API's configuration values.
+
+        Args:
+            data (Dict): Configuration options
+        """
+        if 'api' not in data.keys():
```
theunkn0wn1: `.keys()` is redundant here, `dict.__contains__` checks the keys anyways.
```suggestion
        if 'api' not in data:
```

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+
+        api_data = data['api']
+
+        if not isinstance(api_data['url'], str):
+            raise ValueError("[API] API 'url' value must be a string.")
+
+        if not isinstance(api_data['online_mode'], bool):
+            raise ValueError("[API] API 'online_mode' must be true or false.")
+
+        if not api_data['url'].startswith('http'):
```
theunkn0wn1: shouldn't we be enforcing `HTTPS://` ?

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+            raise ValueError("[API] API 'url' must be a valid URL.")
```
theunkn0wn1: i think we should do more validation that we actually have a valid url, throwing a http/s prefix infront of the string does not ensure validity.

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+                                     'reason': rescue.marked_for_deletion.reason,
+                                     'reporter': rescue.marked_for_deletion.reporter
+                                     }
+                                    ),
+            'rats': ('relationships.rats.data',
+                     [{'type': 'rats', 'id': f"{rat.uuid}"} for rat in rescue.rats]),
+            'first_limpet': ('relationships.firstLimpet.data',
+                             {'type': 'rats',
+                              'id': f"{rescue.first_limpet}"} if rescue.first_limpet else None)
+        }
```
theunkn0wn1, line 192: I believe the objective of the map  /partial object serialization was to avoid initializing an entire ApiRescue data structure whenever we wanted to make an update. Hence i suggested using factory methods, so we could differ evaluation til the point we actually needed to initialize a partial data structure.

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+
+    async def create_rescue(self, rescue: Rescue):
+        """
+        Create the case in the API.
+
+        Args:
+            rescue (Rescue): The Rescue object to create in the API.
+        """
+        data = APIManager._serialize_rescue(rescue)
+        del data['id']
```
theunkn0wn1, line 328: [question] why are we deleting this field?

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+        if data['relationships']['firstLimpet']['data'] is None:
+            del data['relationships']['firstLimpet']
+        jsonapi_data = {'data': data}
+        new_rescue = await self._call('rescues', method='POST', body=jsonapi_data)
+        breakpoint()
```
theunkn0wn1: we probably don't need to include `breakpoint()`s in the source code.

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+
+        Args:
+            rescue (Rescue): The Rescue object to update in the API.
+        """
+        data = APIManager._serialize_rescue(rescue)
+        jsonapi_data = {'data': data}
+        await self._call(f"rescues/{rescue.api_id}", method='PATCH', body=jsonapi_data)
+
+    # @TODO: @functools.lru_cache()
+    async def get_rat(self, uuid: str):
```
theunkn0wn1: [error] missing return type annotation
```suggestion
    async def get_rat(self, uuid: str) -> typing.Optional[Rat]:
```

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+        """
+        Use the API to get a rat by UUID.
+
+        Args:
+            uuid (str): The Rat's UUID.
+
+        Returns:
+            A RatData object describing the Rat found, or None if none was found.
```
theunkn0wn1: Out of date docstring?

theunkn0wn1: missing Raises: block

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+
+        Returns:
+            A properly HTML-escaped URI query string with leading question mark.
+        """
+        param_string = ""
+        if params:
+            param_string = '&'.join(
+                [f"{key}={escape(str(value))}" for key, value in params.items()]
+            )
+        return f"?{param_string}"
```
theunkn0wn1: is there anything in our dependencies we can leverage (such as `requests`) to allow us not to need implement a method like this ourselves?

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+            body (Dict): The body of the request. (For HTTP POST/PATCH/PUT)
+
+        Returns:
+            A dict or list object representing the parsed JSONAPI data returned.
+            Returns None if in offline mode or the request fails.
+        """
+
+        if not self._online_mode:
+            LOG.debug(f"API call to {endpoint} was requested, but we're in offline mode.")
+            return None
```
theunkn0wn1: I think this should be an error, rather than a silent None.

src/packages/api/api_manager.py
```diff
@@ -0,0 +1,468 @@
...
+
+        base_url = self._base_url
+        param_string = self._build_uri_params(params)
+        url = f"{base_url}{endpoint}{param_string}"
+        for retry in range(self.MAX_RETRIES):
```
theunkn0wn1, line 426: a routine like this may be useful in other modules, perhaps we can extract this as an async generator?

theunkn0wn1 (commented):
Haven't looked at the test suite yet.

This is looking great though i do have some comments....

src/packages/api/api_manager.py
Eearslya: As of right now, APIv3 will throw a hissy-fit if you try to create a Rescue with a pre-made UUID.
That very well might change when we move away from the mockup, but that's how it works now.

src/packages/api/api_manager.py
Eearslya: Dangit.

src/packages/api/api_manager.py
theunkn0wn1: Yeah, thats gunna produce a problem given UUIDs are immutable in our data model under the understanding mecha is the one generating the UUIDs 😕

src/packages/api/api_manager.py
Eearslya: Not sure where you're going with this... I need that variable.

theunkn0wn1: see previous suggestion... GH suggestions are a bit silly in that i can only effect one line.

basically im replacing your two liner manual unpack with a 1 liner tuple unpack

src/packages/api/api_manager.py
Eearslya: I'm inclined to disagree. If this method raises when we're offline, that will then force any method that uses the API to catch this error. I feel like modules outside of the API itself shouldn't care if we're offline.

src/packages/api/api_manager.py
Eearslya: I honestly think that's not for Mecha to decide. Using plaintext HTTP could be completely valid if the API was hosted in a local network.

UncleClapton: FWIW The website internally communicates over HTTP with the API since they're both on the same host. Our API proxy system communicates with the client over HTTPS, which relays the request to the API over HTTP. All of our servers are running under a vlan now IIRC, so you could connect directly to the API via local ips.

theunkn0wn1: I still don't think its a good idea to use http where https can be used. of course, since all this is probably going to end up running on the same set of internal servers i suppose it makes little difference.
🤷‍♂

kenneaal: The point that we're trying to make here is that the client should not enforce the scheme. Yes, it makes perfect sense to only use HTTPS. We're only going to use HTTPS. But the software should not lock it down to that through hardcode. At most, it should fire warnings or require a special config directive to allow unsecure HTTP.

src/packages/api/api_manager.py
theunkn0wn1: well, thats the whole point of the Offline Awareness mechanism i created previously: is so other modules can be aware when the system is in offline mode and act accordingly.

src/packages/api/api_manager.py
theunkn0wn1: (for clarity, im refering to the retry loop)

src/packages/api/api_manager.py
Eearslya: Potential solution in c1dd9b4 for review.

src/packages/api/api_manager.py
Eearslya: 87a1a15

Eearslya:
See #158

PR #154, opened 2019-07-17
[SPARK-127] Switch .gitignore to TOML config files

Eearslya:
Something we neglected to do when we switched over to TOML.


theunkn0wn1 (approved):
yeah we should have checked this file. but who doesn't ignore files that are specifically named with `ignore`! :stuck_out_tongue_winking_eye:

PR #156, opened 2019-07-17
[SPARK-150] Add a config option for offline-awareness

Eearslya:
Use the config value `['api']['online_mode']` for offline awareness. Use the rehash handler to read the value and switch online status.

### DIFF
--- a/src/packages/offline_awareness/_offline_aware_abc.py
+++ b/src/packages/offline_awareness/_offline_aware_abc.py
@@ -9,10 +9,13 @@ Licensed under the BSD 3-Clause License.
 See LICENSE.md
 """
 
+import asyncio
 import logging
 import weakref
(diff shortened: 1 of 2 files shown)


src/packages/offline_awareness/_offline_aware_abc.py
```diff
@@ -44,6 +47,38 @@ def __init__(self):
...
+        Apply new configuration data
+
+        Args:
+            data (Dict): New configuration data to apply.
+        """
+        online = data['api']['online_mode']
+        if online:
+            method = cls.go_online
+        else:
+            method = cls.go_offline
```
theunkn0wn1, line 63: [suggestion] I think we can get away with a ternary operation here
```py
            method = cls.go_online if online else cls.go_offline
```

PR #158, opened 2019-12-09
WIP: Fuelrats API integration, phase 1 - Mock API

theunkn0wn1:
this Pull Request is a work in progress, feedback is welcome.

# New Interfaces
## `src.packages.fuelrats_api`
High level package container for API implementations

## `src.packages.fuelrats_api.FuelratsApiABC`
ABC / Protocol base class for API implementations.
Defines the general structure and mandatory methods all API implementations should have.
 - This is implemented using a `typing.Protocol`, which means API implementations are *structurally typed* - they don't need to explicitly subclass the ABC, only need to match the protocol's structure. 

## `src.packages.fuelrats_api.ApiConverter`
Protocol definition for a Converter class. These classes are responsible for converting objects from the API's internal representation to that of Mecha's.

Optionally they can also function in the other direction, converting Mecha's representation to that of the API.

The objective of having converters is to make the API system more modular, isolating changes between API versions where only the data model changed as opposed to those where the entire API changed. 

These protocols are "generic" types, to facilitate typing and static analysis.
Each converter class is responsible for exactly one type.
```py
class ApiConverter(typing.Protocol[_DTYPE]):
   ...
``` 
For instance, a converter for `Rescue` objects would have the annotation
```py
api_converter: ApiConverter[Rescue]
```
and the signature
```py
class SomeRescueConverter(ApiConverter[Rescue])
```
Note: explicit inheritance from `ApiConverter` is not required, only that the protocol is adhered to (simplest if inherited from).

# API Implementations
## `src.packages.fuelrats_api.v3.mockup.MockupAPI`
 implements the major components from the mockup API.

- [ ] Rescue GET
- [x] Rescue SET
- [x] Rescue CREATE
- [x] Rat (name) GET

# TODO:
- [ ] Deploy Mock API in CI integration tests
- [ ] Any added `# TODO` s and `# FIXME`s in added lines
- [ ] Converters for Epic, Quotes


src/packages/fuelrats_api/rat.py
```diff
@@ -0,0 +1,14 @@
+import aiohttp
```
theunkn0wn1: [question] Is this module even used anywhere or a unused artifact of a old implementation?

src/packages/fuelrats_api/v3/_converters.py
```diff
@@ -0,0 +1,77 @@
...
+from loguru import logger
+
+from src.packages.rat import Rat
+from src.packages.rescue import Rescue
+from src.packages.rescue.data import Data as RescueData
+from src.packages.utils import Platforms
+from .._converter import ApiConverter
+
+
+class RatConverter(ApiConverter[Rat]):
```
theunkn0wn1, line 40: Missing module docstring.

src/packages/mark_for_deletion/mark_for_deletion.py
```diff
@@ -10,137 +10,32 @@
...
-        return self._marked
-
-    @marked.setter
-    def marked(self, value: bool) -> None:
-        """
-        Set the MD status
 
-        Args:
-            value (bool): MD status
+def not_empty(inst, attr, value):
```
theunkn0wn1, line 18: [suggestion] move this to the utils package for cross-module use? seems handy.

src/packages/rescue/data.py
```diff
@@ -0,0 +1,18 @@
+import typing
+
+import attr
+
+from ..mark_for_deletion import MarkForDeletion
+
+
+@attr.s
+class Data:
```
theunkn0wn1: Suggestions on a better name for this are welcome, i really don't like `Data` as a class name as i tend to use it as a throw-away variable...

src/packages/board/board.py
```diff
@@ -74,10 +85,13 @@ class RatBoard(abc.Mapping):
...
                  "_storage_by_index",
-                 "_index_counter"]
+                 "_index_counter",
+                 "_offline",
+                 "__weakref__"
+                 ]
 
-    def __init__(self, api_handler=None):
-        self._handler = api_handler
+    def __init__(self, api_handler=None, offline=True):
```
Eearslya: [Style] Missing typehinting

config/testing.toml
```diff
@@ -62,3 +62,4 @@ level = 4
 
 [board]
 cycle_at = 15
+api_url = "https://api.thehellisthis.com"
```
shatteredbeam, line 65: I'm okay with this being added here.

src/packages/board/board.py
```diff
@@ -51,6 +59,9 @@ def validate_config(data: typing.Dict):  # pylint: disable=unused-argument
     if data['board']['cycle_at'] <= 0:
         raise ValueError("constraint cycle_at must be non-zero and positive")
```
shatteredbeam, line 60: I'm assuming we are not explicitly adding a logging call here because this is a fatal error?

src/packages/board/board.py
```diff
@@ -51,6 +59,9 @@ def validate_config(data: typing.Dict):  # pylint: disable=unused-argument
...
 
+    if data['board']['api_url'] == "":
+        raise ValueError("constraint api_url must not be empty.")
```
shatteredbeam, line 63: Same here, fatal error?

src/packages/board/board.py
```diff
@@ -97,6 +111,19 @@ def __init__(self, api_handler=None):
         """
         Internal counter for tracking used indexes
         """
+        self._offline = offline
+
+        super(RatBoard, self).__init__()
+
+    async def on_online(self):
+        logger.info("Board moving to online mode...")
```
shatteredbeam: Please note tense is important.  'Board now online' or 'Board moved to online mode'

src/packages/board/board.py
```diff
@@ -97,6 +111,19 @@ def __init__(self, api_handler=None):
...
+        self._offline = False
+        # TODO get API version from remote and log it
+        # TODO emit canned offline events to API
+
+    async def on_offline(self):
+        logger.warning("Board moving to offline mode...")
```
shatteredbeam: Logging comment tense.

src/packages/board/board.py
```diff
@@ -172,68 +220,79 @@ def __delitem__(self, key: _KEY_TYPE):
...
-            Rescue: rescue to modify based on its `key`
+            created rescue object
         """
+        index = self.free_case_number
+        rescue = Rescue(*args, board_index=index, **kwargs)
 
-        target = self[key]
+        try:
+            if not self.online:
+                logger.warning("creating case in offline mode...")
```
shatteredbeam, line 282: Log output tense.  "Created case in offline mode"

src/packages/board/board.py
```diff
@@ -172,68 +220,79 @@ def __delitem__(self, key: _KEY_TYPE):
...
+            else:
+                logger.trace("creating rescue on API...")
```
shatteredbeam, line 284: Log tense.  "Created rescue posted to API"

src/packages/board/board.py
```diff
@@ -172,68 +220,79 @@ def __delitem__(self, key: _KEY_TYPE):
...
+                rescue = await self._handler.create_rescue(rescue)
 
-        # most tracked attributes may be modified in here, so we pop the rescue
-        # from tracking and append it after
+        except ApiError:
+            logger.exception("unable to create rescue on API!")
+            raise
```
Eearslya, line 290: [Error] As discussed, raising an exception here would ensure that any Rescue that fails to make it to the API is then also not yielded to the caller of `create_rescue`.

src/packages/fuelrats_api/v3/mockup.py
```diff
@@ -0,0 +1,87 @@
...
+
+    async def get_rescue(self, uuid: UUID) -> Rescue:
+        query = f"{self.RESCUE_ENDPOINT}/{uuid}"
+        json = await self._query(method="GET", query=query)
+        return self.rescue_converter.from_api(json)
+
+    async def update_rescue(self, rescue: Rescue) -> None:
+        query = f"{self.RESCUE_ENDPOINT}/{rescue.api_id}"
+        data = self.rescue_converter.to_api(rescue)
+        logger.debug("update data := {}", data)
```
shatteredbeam, line 74: I am the walrus.

src/packages/fuelrats_api/__init__.py
```diff
@@ -0,0 +1,6 @@
+from ._base import FuelratsApiABC
+
+API_ROOT = "{target}/api"
```
Eearslya: [Suggestion] Add a comment describing this string. At first glance, it could be interpreted as an error in implementation.

tests/integration/test_board.py
```diff
@@ -0,0 +1,20 @@
...
+
+async def test_create_rescue(board_online_fx, rescue_plain_fx):
+    """ verifies a case can be created via the online board"""
+    # return Rescue(uuid4(), "UNIT_TEST", "ki", "UNIT_TEST", board_index=42, status=Status.OPEN,
+    #               platform=Platforms.PC)
+
+    async with board_online_fx.create_rescue(
+            "UNIT_TEST", "ki", "UNIT_TEST", status=Status.OPEN, platform=Platforms.PC
+    ) as rescue:
+        ...
```
shatteredbeam: Missing functions?

shatteredbeam (changes requested):

src/packages/fuelrats_api/rat.py
```diff
@@ -0,0 +1,14 @@
...
+from . import RAT_ROOT
+
+TARGET = "http://localhost:6543"
+
+
+async def do_api_action(query: str, mode="GET"):
```
Eearslya: [Style] Missing typehint on `mode`

config/testing.toml
theunkn0wn1: TODO: replace with localhost / CI hosted target

src/packages/board/board.py
theunkn0wn1: AFAIK validators arn't  responsible for emitting the log call, because exceptions here are fatal and get logged anyways upstream.

src/packages/fuelrats_api/v3/mockup.py
theunkn0wn1: Im gunna have to change all the logging calls now that `:=` has special meaning, arn't i? 😕

tests/integration/test_board.py
theunkn0wn1: nah, Though will change in next commit since `create_rescue` is now now a context manager (see `#mechadev` )

src/packages/fuelrats_api/v3/_converters.py
```diff
@@ -0,0 +1,77 @@
...
+
+class RescueConverter(ApiConverter[Rescue]):
+    @classmethod
+    def from_api(cls, data):
+        renames = [("codeRed", "codeRed"), ("unidentifiedRats", "unidentified_rats")]
+        removes = ["data", "createdAt", "updatedAt", "notes", "outcome"]
+        content = data["data"]
+        uuid = UUID(content["id"])
+        attributes = content["attributes"]
+        logger.debug("original attributes:= {}", attributes)
```
Eearslya, line 63: [Suggestion] Demotion to trace

src/packages/fuelrats_api/v3/_converters.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # some fields have different names in the API
+        for source, destination in renames:
+            attributes[destination] = attributes[source]
+            del attributes[source]
+        # some fields we don't / cannot want to send to the API at all.
+        for key in removes:
+            del attributes[key]
+
+        logger.debug("translated attributes := {}", attributes)
```
Eearslya, line 79: [Suggestion] Demotion to trace

src/packages/rescue/data.py
Eearslya: [Suggestion] Well, the package is `rescue.data`, how about `RescueData`?

src/packages/rescue/rat_rescue.py
```diff
@@ -783,27 +785,22 @@ def mark_delete(self, reporter: str, reason: str) -> None:
         Raises:
             TypeError: invalid params
         """
-        # type enforcement
-        if not isinstance(reporter, str) or not isinstance(reason, str):
-            raise TypeError(f"reporter and/or reason of invalid type. got {type(reporter)},"
-                            f"{type(reason)}")
+
+        self.marked_for_deletion = MarkForDeletion(reporter=reporter, reason=reason, marked=True)
+        if (reporter or reason) and (not reporter or not reason):
```
Eearslya: [Question] Why are we validating the inputs AFTER we have already used them to create the MFD?

tests/conftest.py
```diff
@@ -361,3 +361,17 @@ def configuration_fx() -> Dict:
     provides the session configuration dictionary, as loaded at test session start.
     """
     return ConfigReceiver.data
+
+
+@pytest.fixture
+def mock_fuelrats_api_fx():
+    # TODO pull from configuration system
+    return MockupAPI(url=r'http://api.thehellisthis.com:6543/api')
```
Eearslya, line 369: [Suggestion] Would it not be better to rely on `aiohttp`'s built-in mocking system? This makes our tests depend upon an outside source, or we would have to develop and spin up a webserver during testing. See `Galaxy`'s test suite for mocking examples.

tests/unit/test_rescue.py
```diff
@@ -102,7 +103,7 @@ def test_updated_at_raises_typeerror(rescue_sop_fx):
     Verify Rescue.updated_at raises TypeError if given incorrect value,
     or is set to a date in the past.
     """
-    rescue_sop_fx._createdAt = datetime(1991, 1, 1, 1, 1, 1,)
+    rescue_sop_fx._createdAt = datetime(1991, 1, 1, 1, 1, 1, )
```
Eearslya, line 106: [Style] Extraneous space

Eearslya (dismissed):
I'll go through this with a finer-tooth comb when it's ready, but here are some of my initial thoughts and nitpicks.

src/packages/rescue/data.py
theunkn0wn1: That is now i imported it in the API stuff... still seems a tad bad.

src/packages/rescue/rat_rescue.py
theunkn0wn1: MFD only validates the object is strictly invalid, I reasoned it was possible that valid MFD objects may exist that fail this extra check, but mecha should refuse to create new instances.

tests/conftest.py
theunkn0wn1: Great suggestion, TBH I totally forgot such features existed. I may consult you and the Galaxy code to help get this up to speed 😇

src/packages/fuelrats_api/v3/mockup.py
```diff
@@ -0,0 +1,101 @@
...
+            **kwargs:  kwargs passed to underlying Aiohttp query
+
+        Returns:
+            API response
+
+        Raises:
+            ApiError: non-2xx response code received
+        """
+        logger.debug("[{method}] {query}", method=method, query=query)
+        async with aiohttp.ClientSession() as session:
```
Eearslya, line 48: [Suggestion] You'll probably want to use the `raise_for_status` and `timeout` parameters here as well. It will save you from having to check the codes yourself.
By default, I believe `aiohttp` will simply return an empty body for a failed request, rather than raise an exception.
Default timeout is 5 minutes, which is probably far too long for our uses.

src/packages/rescue/rat_rescue.py
Eearslya: I'm not sure I understand.
I can see that `MarkForDeletion` has a validator to ensure that `reporter` and `reason` cannot be an empty string, but they're also optional, meaning they can be `None`.
This check does absolutely nothing to prevent creating a new `MarkForDeletion` instance with `reason` and `reporter` both set to `None`, as the object is instantiated and assigned to the `marked_for_deletion` member immediately. If this check fails and decides to raise, the only thing that ends up being skipped is the additional log call below.

theunkn0wn1: the MFD fields can be null, this is intended behavor.

the validators on MFD only validate that - if its not null - the attributes are non-empty strings.

This check ensures that , if MFD is filled in, its filled in entirely. you can't MFD without a reason.

src/packages/fuelrats_api/v3/_converters.py
theunkn0wn1: To my interpretation, "trace" is for "i was here" kind of messages.
"debug" is for "what was I thinking while i was here" kind of messages.

src/packages/rescue/rat_rescue.py
Eearslya: > This check ensures that , if MFD is filled in, its filled in entirely. you can't MFD without a reason.

Mmm, but that's not true either.

So, imagine for instance we call `rescue.mark_delete("Eearslya", None)`.
On line 789, the MFD object is created, instantiated, and assigned to the rescue with `reporter = "Eearslya"` and `reason = None`.

At this point in time, **the rescue is marked for deletion without a reason**.

On line 790, we check to make sure that if one field is filled out, they both are. Because `reason` is `None`, this check will raise an error.
But, the problem is...the MFD object is already attached to the rescue. It's still marked for deletion, there's no code that removes it if it's found to be invalid.

src/packages/board/board.py
theunkn0wn1: No, this log event occurs prior to the rescue being created on the API, thus present tense is correct.
Especially since this action can result in a failure, past-tense is inappropriate.

src/packages/board/board.py
theunkn0wn1: https://github.com/FuelRats/pipsqueak3/pull/158#discussion_r355844831

src/packages/board/board.py
theunkn0wn1: As discussed, this method is now not a context manager. Im gunna close this one as out-of -date.
If the issue still exists, which to my understanding has been resolved, please do tell.

src/packages/fuelrats_api/v3/_converters.py
theunkn0wn1: https://github.com/FuelRats/pipsqueak3/pull/158#discussion_r355830145

src/packages/rescue/rat_rescue.py
theunkn0wn1: ... i could have sworn i implemented this so the assignment occurs after 🤦‍♂ 
Further, the tests for this routine didn't cover this case properly. fix incoming.

theunkn0wn1: 9ba7b9ef

src/packages/board/board.py
shatteredbeam: Yep, thats fine. Just making sure this is intended.

src/packages/fuelrats_api/v3/mockup.py
shatteredbeam: Eh, I'm not ready to make stand on this one way or another, I would just ask for consistency.  Use it where appropriate, otherwise don't ;)

theunkn0wn1:
Base changed to a feature branch so other downstreams can use it.

Please note this is now **not** targeting develop, and is still a WIP. 
I would like to merge it as to un-block @shatteredbeam :innocent:

shatteredbeam (approved):
Leaving the nitpicky stuff out, this is fine.  Good job.

PR #160, opened 2019-12-11
[Spark 158] Galaxy Feature Update

Eearslya:
With the recent changes to the Systems API, Galaxy has had to be updated to match functionality. All of these changes are internal to Galaxy, and no change has occurred to the format of data returned to any other module using Galaxy.

---

The changes to Galaxy include the following:

**API Change:** `/api/systems` no longer has an `UPPERCASE` index on system names.
**Galaxy Solution:** When asked for a system by name, Galaxy will now search for the system on the `/search` endpoint, followed by a direct ID query to `/systems/api/:id`.

**API Change:** Added `/landmark`.
**Galaxy Solution:** Remove all internal implementation for finding landmark systems. `find_nearest_landmark` will now query the API and return the results it is given.

**API Change:** Added `/mecha`.
**Galaxy Solution:** Change internal implementation of `search_systems_by_name` to use `/mecha` instead of `/search`.

---

> ***NOTE:*** *The current implementation of this code is tested against mock data that includes an `id` attribute for the `/search` endpoint upon finding a perfect match. This does not match actual API functionality at the time of writing this PR, but is expected to be resolved soon.*


theunkn0wn1 (dismissed):
LGTM

shatteredbeam (dismissed):
Nice job.  No comments.  LGTM.

Eearslya:
Thanks for the approvals, but I'm going to put this on hold for a bit longer.

After speaking with Abs and Clap, it seems like we're likely dropping support for commands like `!plot`, as Spansh does a better job than we ever will, and there's no sense reinventing a very complex wheel.
Therefore, this means that I need a lot less data from the API and can now go through and optimize and/or remove some extraneous calls to keep things speedy.

PR #164, opened 2019-12-20
Redid Rescue.(add/remove)_rat to be more sane

theunkn0wn1:
- Rescue's (add/remove)_rat now simply accept a Rat object
 - no fancy cache lookups
 - removed outdated tests that verified the old behavior
 - added property to rat to tell if its an identified object

## bonus requested features
 - `Rescue.rats` is now a dictionary, `Dict[str, Rat]` by casefolded rat name.
 - `Rescue.unidentified_rats`  is now a dictionary, `Dict[str, Rat]` by casefolded rat name.

It is to my opinion the Rescue object should be naive to the source of a Rat object, only that its getting rat objects.

### DIFF
--- a/src/packages/rat/rat.py
+++ b/src/packages/rat/rat.py
@@ -13,11 +13,12 @@ See LICENSE.md
 This module is built on top of the Pydle system.
 """
 
-from loguru import logger
 from functools import reduce
 from operator import xor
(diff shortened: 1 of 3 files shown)


shatteredbeam (approved):
LGTM

PR #166, opened 2019-12-26
Infratructure/recently used rescues

theunkn0wn1:
the board object now holds two additional data members:

`_recently_closed` is a deque that holds the last N resscues closed out by this board object`
 - exposed on the public interface via `recently_closed` property

`_offline_rescue_storage` is a queue that holds all rescues closed out by this board, when the board is acting in offline mode. 
This queue should empty once the board receives its on_online callback.

### DIFF
--- a/src/packages/board/board.py
+++ b/src/packages/board/board.py
@@ -11,6 +11,7 @@ See LICENSE.md
 """
 from __future__ import annotations
 
+import collections
 import itertools
 import typing
 from asyncio import Lock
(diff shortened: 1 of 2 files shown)


src/packages/board/board.py
```diff
@@ -124,13 +130,24 @@ def __init__(
...
+        return self._recently_closed
+
     async def on_online(self):
         logger.info("Rescue board online.")
         self._offline = False
-        # TODO get API version from remote and log it
-        # TODO emit canned offline events to API
+        logger.info("emitting {} cached rescues to the API...", len(self._offline_rescue_storage))
+        for rescue in self._offline_rescue_storage:
+            await self._handler.update_rescue(rescue)
```
Eearslya: [Error] Shouldn't this be using `self._offline_rescue_storage.popleft()`?
As it stands, nothing is emptying the deque, meaning if we went off/online again, rescues would be sent again.

Eearslya (changes requested):
Looks good, except one flaw I noticed.

src/packages/board/board.py
theunkn0wn1: absolutely correct, fixed.

PR #167, opened 2020-01-04
[feature] starter system detection

mbravorus:
Added a function to test system name for belonging to a list of permit-locked starter systems; hopefully to be used alongside autocorrect for helpful announcement for rats about to call jumps.

Unit test supplied.

Minor fixes in utils __init__.py suggested

### DIFF
--- a/src/packages/utils/__init__.py
+++ b/src/packages/utils/__init__.py
@@ -10,12 +10,15 @@ See LICENSE.md
 """
 
 from .autocorrect import correct_system_name
+
+from .starter_systems import isStarterSystem
 from .ratlib import sanitize, Vector, Colors, color, bold, underline, italic, reverse, Platforms, \
     Singleton, Status, Formatting
 
 __all__ = [
-    "autocorrect",
-    "ratlib",
+    "correct_system_name",
+    "isStarterSystem",
+    "sanitize",
     "Vector",
     "Colors",
     "color",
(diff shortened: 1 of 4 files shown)


src/packages/utils/__init__.py
```diff
@@ -10,21 +10,24 @@
...
+
+from .starter_systems import isStarterSystem
 from .ratlib import sanitize, Vector, Colors, color, bold, underline, italic, reverse, Platforms, \
     Singleton, Status, Formatting
 
 __all__ = [
-    "autocorrect",
-    "ratlib",
+    "correct_system_name",
+    "isStarterSystem",
```
theunkn0wn1, line 20: [style] PEP8 [requires `snake_case` naming for functions and methods](https://www.python.org/dev/peps/pep-0008/#id45).

theunkn0wn1 (changes requested):
Greetings, thank you for your proposed changes to mecha3, however I wished you consulted with us prior to beginning work on this feature 😕 .

Mecha3 is not the source of truth for system data, it pulls the information it needs from the Systems API (`SAPI`). Likewise, hardcoding is not a good practice to hard-code data that doesn't need to be. 

I advise you look into our `src.packages.galaxy` implementation and adapt that to consume the starter system  + permit lock information from the SAPI, instead of specifying that information in this project's sources. Hardcoding here simply is not maintainable.

Once again, thank you for the time taken to prepare this feature, and I look forward to what you can come up with 🙂 .

o7, Commander.

theunkn0wn1:
Methinks this has been left open without progress long enough.
Feel free to reach out to me on IRC if you intend to develop this further.

Closing as stale.

PR #170, opened 2020-02-07
track Famous Last words

theunkn0wn1:
This PR tracks the last thing users say in any bot-visible channel. this is primarily to the benefit of `!grab`.

this adds one public method to the `MechaClient`, 
```py 
   def get_last_message(self, channel: str, user: str) -> typing.Optional[str]:...
```

### DIFF
--- a/src/features/message_history.py
+++ b/src/features/message_history.py
@@ -21,6 +21,9 @@ class MessageHistoryClient(RFC1459Support):
 
         return await super().on_message(target, by, message)
 
-    def get_last_message(self, channel: str, user: str):
+    def get_last_message(self, channel: str, user: str) -> typing.Optional[str]:
+        """
+        Fetches the last thing a specified user said in a specified channel the bot could see.
+        """
         users = self.__channel_history.get(channel, {})
         return users.get(user)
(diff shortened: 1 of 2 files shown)


src/mechaclient.py
```diff
@@ -78,6 +79,7 @@ def __init__(self, *args, mecha_config=None, **kwargs):
         :param message: message body
         :return:
         """
+        await super(MechaClient, self).on_message(channel, user, message)
```
theunkn0wn1:
```suggestion
        await super().on_message(channel, user, message)
```

src/features/message_history.py
```diff
@@ -0,0 +1,29 @@
...
+    __slots__ = [
+        "__channel_history"
+    ]
+
+    def __init__(self, nickname, fallback_nicknames=[], username=None, realname=None, eventloop=None,
+                 **kwargs):
+        super().__init__(nickname, fallback_nicknames, username, realname, eventloop, **kwargs)
+        self.__channel_history: typing.Dict[str, typing.Dict[str, str]] = {}
+
+    async def on_message(self, target, by, message):
```
Eearslya: [Style] Typehinting

src/mechaclient.py
Eearslya: [Comment, NonBlocking] Inheritance is scary.

Eearslya (dismissed):
One missing typehinting, but in general, LGTM

Eearslya (approved):

PR #172, opened 2020-03-01
Galaxy Search Update

Eearslya:
This update reworks Galaxy to search for systems on the `/api/systems` endpoint using `ILIKE`, rather than the `/search` endpoint.

Additionally, as `/landmark` requires a valid system name, Galaxy will now validate systems given to `find_nearest_landmark` before attempting to search.

### DIFF
--- a/src/packages/galaxy/galaxy.py
+++ b/src/packages/galaxy/galaxy.py
@@ -13,7 +13,7 @@ See LICENSE.md
 import asyncio
 import json
 import typing
-from html import escape
+from urllib.parse import urlencode
 
 import aiohttp
 
@@ -72,7 +72,7 @@ class Galaxy:
             "sort": "name",
             "limit": 1
         })
-        print(data)
+
         if 'data' in data and data['data']:
             if full_details:
                 return await self.find_system_by_id(data['data'][0]['id'])
@@ -186,8 +186,8 @@ class Galaxy:
         base_url = self.url
         param_string = ""
         if params:
-            param_string = '&'.join(
-                [f"{key}={escape(str(value))}" for key, value in params.items()]
+            param_string = urlencode(
+                [(key, value) for key, value in params.items()]
             )
         url = f"{base_url}{endpoint}?{param_string}"
         for retry in range(self.MAX_RETRIES):
--- a/src/packages/galaxy/galaxy.py
+++ b/src/packages/galaxy/galaxy.py
@@ -67,7 +67,11 @@ class Galaxy:
             A ``StarSystem`` object representing the found system, or ``None`` if none was found.
         """
 
-        data = await self._call("api/systems", {"filter[name:ilike]": name, "sort": "name", "limit": 1})
+        data = await self._call("api/systems", {
+            "filter[name:ilike]": name,
+            "sort": "name",
+            "limit": 1
+        })
         print(data)
         if 'data' in data and data['data']:
             if full_details:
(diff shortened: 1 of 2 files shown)


theunkn0wn1 (dismissed):
LGTM

tests/fixtures/galaxy_fx.py
```diff
@@ -69,32 +65,32 @@ def mock_system_api_server_fx():
...
         )
         # - Chua Eohn CT-F d12-2
-        httpserver.expect_request("/search", query_string=b"name=Chua+Eohn+CT-F+d12-2&limit=1").respond_with_data(
-            """{"meta": {"name": "Chua Eohn CT-F d12-2", "type": "Perfect match"}, "data": [{"name": "Chua Eohn CT-F d12-2", "similarity": "Perfect match", "id": 19626238}]}"""
+        httpserver.expect_request("/api/systems", query_string=b"filter%5Bname:ilike%5D=Chua+Eohn+CT-F+d12-2&sort=name&limit=1").respond_with_data(
```
kenneaal, line 88: It is probably a good idea to properly encode URLs with quote_plus(), you should not expect endpoints to handle incorrectly encoded URL characters.

theunkn0wn1: Nice catch, I believe we [found this out the hard way in mecha2](https://github.com/FuelRats/pipsqueak/pull/317).

theunkn0wn1 (changes requested):
As per above conversations, re [#317](https://github.com/FuelRats/pipsqueak/pull/317), we should urlencode sysapi calls

src/packages/galaxy/galaxy.py
```diff
@@ -67,12 +67,13 @@ def __init__(self, url: str = None):
             A ``StarSystem`` object representing the found system, or ``None`` if none was found.
         """
 
-        data = await self._call("search", {"name": name, "limit": 1})
-        if 'data' in data and data['data'] and data['data'][0]['name'].casefold() == name.casefold():
+        data = await self._call("api/systems", {"filter[name:ilike]": name, "sort": "name", "limit": 1})
```
kenneaal: Is the intention to use a different function for !search and other wilder-matching commands? Using the JSONAPI endpoint limits you to the single search strategy (ILIKE, without wildcards), while the /mecha endpoint is designed for mecha's !search functionality, providing similarity based searches.

Eearslya: The intention here is to fetch a system by name, case insensitively. Any system given to this method is expected to at least have the correct spelling, the call to the API is simply used to verify the correct capitalization. We have a separate `search_systems_by_name` method that uses `/mecha` for its results.

tests/fixtures/galaxy_fx.py
Eearslya: I've updated the `_call` method to use `urllib.urlencode` rather than `html.escape`. I think, however, I'm going to leave the test data as-is. The point of the mock data is to represent a known constant value, and calculating URIs at runtime would be unnecessary for that purpose.

src/packages/galaxy/galaxy.py
```diff
@@ -67,12 +67,17 @@ def __init__(self, url: str = None):
...
         """
 
-        data = await self._call("search", {"name": name, "limit": 1})
-        if 'data' in data and data['data'] and data['data'][0]['name'].casefold() == name.casefold():
+        data = await self._call("api/systems", {
+            "filter[name:ilike]": name,
+            "sort": "name",
+            "limit": 1
+        })
+        print(data)
```
theunkn0wn1: [error] logging.debug ?
```suggestion
        logger.debug("system API response := {}", data)
```

theunkn0wn1 (commented):
One minor missed `print` statement, otherwise ready to merge methinks.

theunkn0wn1 (approved):
LGTM

PR #165, opened 2019-12-21
Core Commands, First pass.

shatteredbeam:
First draft for core case commands.

There WILL be missing items, and possibly incorrect methods.  Thus far methods/interaction have been very liquid throughout the writing process.

See case_management.py.

There are currently no tests.

PR #175, opened 2020-04-06
Drillsqueak Beta 3.0.0b3

theunkn0wn1:
this is a continuation of #165 , fixing several bugs, added lots of tests, other backend magic etc.

there is honestly quite a bit to this pr... :sweat_smile: 

Note this is what is currently deployed to drillsqueak.


src/commands/_list_flags.py
```diff
@@ -0,0 +1,43 @@
...
+
+        show_inactive = "i" in argument
+        """
+        Show inactive rescues
+        """
+        show_assigned_rats = "r" in argument
+        """
+        Show assigned rats per rescue
+        """
+        filter_unassigned_rescues = "u" in argument
```
Eearslya, line 36: [Style, Suggestion] Newlines between docstrings to make it more obvious which string is referencing what.

src/commands/case_management.py
```diff
@@ -0,0 +1,700 @@
...
+from ..packages.commands import command
+from ..packages.context.context import Context
+from ..packages.epic import Epic
+from ..packages.permissions.permissions import require_permission, RAT, OVERSEER, require_channel
+from ..packages.quotation.rat_quotation import Quotation
+from ..packages.rat import Rat
+from ..packages.rescue import Rescue
+from ..packages.utils import Platforms, Status
+
+_TIME_RE = re.compile(r'(\d+)[: ](\d+)')
```
Eearslya: [Style] Might deserve a docstring describing what this is for.

src/commands/case_management.py
```diff
@@ -0,0 +1,700 @@
...
+
+
+# User input validation helper
+def _validate(ctx: Context, validate: str) -> typing.Optional[Rescue]:
```
Eearslya, line 46: [Style] Missing docstring.

src/commands/case_management.py
```diff
@@ -0,0 +1,700 @@
...
+async def cmd_case_management_clear(ctx: Context):
+    if len(ctx.words) < 2:
+        await ctx.reply("Usage: !clear <Client Name|Board Index> [First Limpet Sender]")
+        return
+
+    # Pass case to validator, return a case if found or None
+    rescue = _validate(ctx, ctx.words[1])
+
+    # Only set First Limpet if it was specified.
+    if len(ctx.words) == 3:
```
Eearslya, line 143: [Question] Maybe we should throw a usage error if the command has more than 3 words, too? Just to be safe. e.g. `!close 3 unknown [pc]` would close the case with no FL.

src/commands/case_management.py
```diff
@@ -0,0 +1,699 @@
...
+
+    if not rescue:
+        case = await ctx.bot.board.create_rescue(client=ctx.words[1])
+        async with ctx.bot.board.modify_rescue(case) as case:
+            case.add_quote(last_message)
+
+        return await ctx.reply(f"{case.client}'s case opened with {last_message!r}")
+
+    if ctx.words[1].casefold() in ctx.bot.last_user_message:
+        last_message = ctx.bot.last_user_message[ctx.words[1].casefold()]
```
Eearslya, line 302: [Error?] I'm trying to look at this as best as I can, but this seems to be redundant because of L274?

src/commands/case_management.py
```diff
@@ -0,0 +1,699 @@
...
+    elif int(ctx.words[1]) in ctx.bot.board:
+        if rescue.client.casefold() in ctx.bot.last_user_message:
+            last_message = last_message
```
Eearslya: [Error] What exactly does this do...?

src/commands/case_management.py
```diff
@@ -0,0 +1,699 @@
...
+    if not rescue:
+        logger.debug("creating rescue for {!r}", ctx.words[1])
+        rescue = await ctx.bot.board.create_rescue(client=ctx.words[1])
+        async with ctx.bot.board.modify_rescue(rescue) as case:
+            case.add_quote(ctx.words_eol[2], ctx.user.nickname)
+
+            for keyword in ctx.words_eol[2].split():
+                if keyword.upper() in {item.value for item in Platforms}:
+                    case.platform = Platforms[keyword.upper()]
+                if keyword.casefold() == "cr" or "code red" in keyword.casefold() \
```
Eearslya: [Error] If the `words_eol` is split on whitespace, `"code red"` is not a possible result.

src/commands/case_management.py
```diff
@@ -0,0 +1,699 @@
...
+        return
+
+    new_name = ctx.words[2]
+
+    async with ctx.bot.board.modify_rescue(rescue) as case:
+        case.irc_nickname = new_name
+        await ctx.reply(f"Set IRC Nickname to {case.irc_nickname!r}")
+
+    if new_name.casefold() != rescue.client.casefold():
+        await ctx.reply("Caution: IRC Nickname does not match CMDR Name.")
```
Eearslya: [Question] This warning seems redundant. Isn't the whole point of this command to set IRC/CMDR names to different values?

src/commands/deletion_management.py
```diff
@@ -0,0 +1,85 @@
...
+from ..packages.commands import command
+from ..packages.context.context import Context
+from ..packages.mark_for_deletion import MarkForDeletion
+from ..packages.permissions.permissions import require_permission, RAT, OVERSEER, require_channel
+from ..packages.rescue import Rescue
+from ..packages.utils import Status
+
+
+# User input validation helper
+def _validate(ctx: Context, validate: str) -> Optional[Rescue]:
```
Eearslya, line 30: [Error] Duplicate from `case_management.py`, consider moving it?

src/mechaclient.py
```diff
@@ -70,6 +75,9 @@ def __init__(self, *args, mecha_config=None, **kwargs):
     #
     # def on_join(self, channel, user):
     #     super().on_join(channel, user)
+    async def on_invite(self, channel, by):
```
Eearslya, line 86: [Suggestion] Consider limiting this by inviter VHost, if possible?

src/packages/board/board.py
```diff
@@ -155,85 +193,148 @@ def free_case_number(self) -> int:
...
+    async def append(self, rescue: Rescue, overwrite: bool = False) -> None:
+        """
+        Append a rescue to ourselves
+
+        If the rescue doesn't have a board index, it will be assigned one.
+
+        Args:
+            rescue (Rescue): object to append
+            overwrite(bool): overwrite existing cases
+    """
```
Eearslya, line 205: [Style] Docstring indentation

tests/integration/api/test_mock_v3.py
```diff
@@ -0,0 +1,29 @@
...
+    result = await mock_fuelrats_api_fx.get_rescue(target)
+
+    assert isinstance(result, Rescue)
+
+
+async def test_update_rescue(mock_fuelrats_api_fx, rescue_sop_fx):
+    target = UUID("c1ff8456-5c51-47b1-8d0e-be6104840820")
+    rescue_sop_fx._api_id = target
+
+    await mock_fuelrats_api_fx.update_rescue(rescue_sop_fx)
```
Eearslya, line 21: [Error] No assertions? If this test is meant to just assert that no exceptions are raised, it should probably say so explicitly.

tests/integration/test_board.py
```diff
@@ -0,0 +1,21 @@
...
+
+async def test_create_rescue(board_online_fx, rescue_plain_fx):
+    """ verifies a case can be created via the online board"""
+    # return Rescue(uuid4(), "UNIT_TEST", "ki", "UNIT_TEST", board_index=42, status=Status.OPEN,
+    #               platform=Platforms.PC)
+
+    rescue = await  board_online_fx.create_rescue(
+        "UNIT_TEST", "ki", "UNIT_TEST", status=Status.OPEN, platform=Platforms.PC
+    )
+    ...
```
Eearslya: [Style] Unnecessary pass

Eearslya (commented):
Some changes I'd like to see and questions answered, but nothing explicitly preventing merge imo.

src/commands/case_management.py
theunkn0wn1: I agree. @shatteredbeam what *does* this do?

src/commands/deletion_management.py
shatteredbeam: Interesting, in my version from awhile back this move was already done.

src/mechaclient.py
```diff
@@ -126,7 +135,7 @@ def fact_manager(self) -> FactManager:
         This is initialized in a lazy way to increase overall startup speed.
         """
         if not self._fact_manager:
-            self._fact_manager = FactManager()  # Instantiate Global Fact Manager
+            self._fact_manager = FactManager(fact_table="converted_facts")  # Instantiate Global Fact Manager
```
shatteredbeam: [Error] this as discussed is kind of hacky.  The fact manager wasn't designed to be given a table for security reasons, but rather than hard code the name it needs to be moved to a config option.

src/packages/commands/rat_command.py
```diff
@@ -76,10 +81,46 @@ class NameCollisionException(CommandException):
...
+        # don't do anything if the fact doesn't exist
+        if not await context.bot.fact_manager.exists(fact, lang):
+            logger.debug("no such fact name={!r} lang={!r}", fact, lang)
+            return False
+
+        logger.debug("fact exists, retrieving and returning!")
+        fact = await context.bot.fact_manager.find(fact, lang)
+        await context.reply(fact.message)
+        return True
+    except psycopg2.Error:
```
shatteredbeam, line 121: Can we use a more specific error handler here? I feel this is too broad.

src/packages/database/database_manager.py
```diff
@@ -174,6 +174,7 @@ def __init__(self,
             connection.set_client_encoding('utf-8')
             # Create cursor, and execute the query.
             with connection.cursor() as cursor:
+                logger.debug("executing query {}", query)
```
shatteredbeam: this has security concerns, if you're exposing the entire query.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -13,17 +13,20 @@
...
+from typing import Optional, Dict, TypedDict
 from src.config import CONFIG_MARKER
 from ..context import Context
 from ..rescue import Rescue
 from ..rules import rule
 from ..utils import Platforms
 
 
-_config: Dict = {}
+class _RatmammaConfig(TypedDict):
```
Eearslya: [Error?] Are we going with `Ratmama` or `Ratmamma`? The file-name seems to indicate the former.

theunkn0wn1: ... that wasn't even supposed to be comitted yet darnit.
will fix

shatteredbeam (changes requested):

src/mechaclient.py
shatteredbeam: I also agreed to take care of this, so this should not be considered blocking.

src/mechaclient.py
theunkn0wn1: yeah... this should probably be behind permission checks.

src/commands/case_management.py
```diff
@@ -0,0 +1,725 @@
...
+    RAT,
+    OVERSEER,
+    require_channel,
+)
+from ..packages.quotation.rat_quotation import Quotation
+from ..packages.rat import Rat
+from ..packages.rescue import Rescue
+from ..packages.utils import Platforms, Status
+
+_TIME_RE = re.compile(r"(\d+)[: ](\d+)")
```
Eearslya, line 38:
```suggestion
_TIME_RE = re.compile(r"(\d+)[: ](\d+)")
"""
Regex matcher used to find a time within a string. Used to determine
if a newly-submitted case is code red or not.
"""
```

shatteredbeam: I was just working on answering this, but yes.  this is just a simple way to see if time was included in an inject, and then to auto-mark the case CR if included.

src/commands/case_management.py
theunkn0wn1: good question. looking at the delta i introduced that i can't make sense of why i changed it either. reverting..

src/commands/case_management.py
theunkn0wn1: no. generally the IRC nickname matches the Commander name, and it is thus a warning in mecha2 if these two values diverge.

src/packages/database/database_manager.py
theunkn0wn1: placed behind a `__debug__` check.

src/packages/commands/rat_command.py
theunkn0wn1: Nope! not without having some way of asking the FM if it has a connection or not.

src/commands/case_management.py
Eearslya: That seems like a fair warning when the case is first announced, but the entire point of this command is to change the IRC Nickname away from the CMDR name, is it not? Since the client usually joins with their CMDR Name as their IRC Name to begin with (barring any forbidden characters).

src/packages/commands/rat_command.py
shatteredbeam: You can.  if the pool is None or in an error state there's no active connection.  I believe there's also a 'connected' property I added.

shatteredbeam: We discussed this, and decided there was a better way - by adding a method to check if the current connection exists.  This should also be considered non-blocking and out of scope.

src/commands/case_management.py
theunkn0wn1: yeah, this one would be better to throw a usage error; you can't close a case to more than one FL. After all, its not called a _first_ limpet for nothing.

tests/integration/api/test_mock_v3.py
theunkn0wn1: i... honestly don't even remember. I also don't know how much of this implementation is even being preserved. Looking at the implementation, this was for the mockup v3 rather than the actual api v3. i may just gut these tests entirely

src/packages/commands/rat_command.py
theunkn0wn1: For the sake of keeping the drillsqueak beta alive, we will leave this change in for now. 
As per our discussion, @shatteredbeam will provide better methods for achieving this end without such an ugly exception clause in a later PR.

src/commands/case_management.py
```diff
@@ -0,0 +1,732 @@
...
+
+    new_name = ctx.words[2]
+
+    async with ctx.bot.board.modify_rescue(rescue) as case:
+        case.irc_nickname = new_name
+        await ctx.reply(f"Set IRC Nickname to {case.irc_nickname!r}")
+
+    if new_name.casefold() != rescue.client.casefold():
+        await ctx.reply(
+            f"Caution: IRC Nickname {new_name!r}does not match CMDR Name {rescue.client.casefold()!r}."
```
Eearslya: [Style] Missing space after `new_name`.

src/commands/case_management.py
```diff
@@ -0,0 +1,732 @@
...
+        if output:
+            await ctx.reply(output)
+
+
+def _list_rescue(rescue_collection, format_specifiers):
+    buffer = io.StringIO()
+    buffer.write(f"{len(rescue_collection)} active cases. ")
+    for rescue in rescue_collection:
+        buffer.write(format(rescue, format_specifiers))
+        buffer.write("\n")
```
Eearslya, line 705: [Error?] Looking at Mecha2's current `list` functionality, isn't there only meant to be a newline between cases if additional flags are given? (e.g. show assigned rats)

src/commands/starsystems.py
```diff
@@ -0,0 +1,35 @@
...
+
+    system_name = ctx.words_eol[1]
+    logger.trace("searching for system {}", system_name)
+    found = await ctx.bot.galaxy.find_system_by_name(system_name)
+    if not found:
+        return await ctx.reply(f"{system_name} was not found in The Fuel Rats System Database.")
+
+    logger.debug("found system {}, acquiring nearest landmark...", found)
+    nearest_landmark = await ctx.bot.galaxy.find_nearest_landmark(found)
+    return await ctx.reply(f"{nearest_landmark}")
```
Eearslya, line 28: [Suggestion] `find_nearest_landmark` also returns the distance as a second return value, suggest we pass this info along.

src/packages/rescue/rat_rescue.py
```diff
@@ -837,3 +784,76 @@ def change(self):
...
+                buffer.write(base)
+
+        if self.platform:
+            base = self.platform.name
+
+            if coloured:
+                if self.platform is Platforms.XB:
+                    buffer.write(color(base, Colors.GREEN))
+                elif self.platform is Platforms.PS:
+                    buffer.write(color(base, Colors.LIGHT_BLUE))
```
Eearslya, line 823: [Error] Missing case for PC.

tests/strategies.py
```diff
@@ -0,0 +1,85 @@
...
+
+def rescues(min_size: int, max_size: int):
+    """ builds a list of rescues, shrinks towards smaller lists and smaller rescues """
+    return strategies.lists(rescue, min_size=min_size, max_size=max_size,
+                            unique_by=(
+                                lambda case: case.irc_nickname, lambda case: case.board_index,
+                                lambda case: case.client))
+
+
+""" A list of rescues """
```
Eearslya: [Style] What is this docstring attached to? If it's the above `rescues` function, it seems odd to have two newlines between them.

src/packages/rescue/rat_rescue.py
theunkn0wn1: you know i looked at this a good dozen times, and only now am i seeing that. the Else statement on l824 got be confused...

tests/strategies.py
theunkn0wn1: its orphaned. removing.

src/packages/commands/rat_command.py
theunkn0wn1: SPARK-164

src/mechaclient.py
theunkn0wn1: SPARK-167

src/commands/starsystems.py
theunkn0wn1: [error] actually implement this routine properly.
whats currently here is test code

src/mechaclient.py
theunkn0wn1: marking as  SPARK-163, will fix

src/mechaclient.py
```diff
@@ -102,6 +102,10 @@ def __init__(self, *args, mecha_config=None, **kwargs):
         try:
             self._last_user_message[user.casefold()] = sanitized_message  # Store sanitized message
             ctx = await Context.from_message(self, channel, user, sanitized_message)
+            if not ctx.words:
+                logger.trace("ignoring empty message")
```
Eearslya, line 122: [Suggestion] As this came up as a bug, should we add a regression test?

src/commands/case_management.py
theunkn0wn1: that isnt right *writes up counter example* oh, so it is.

Fixed.

src/mechaclient.py
theunkn0wn1: agreed.

shatteredbeam (approved):
👍

Eearslya (approved):
Good enough to be getting on with. Polishing can be done later.

PR #178, opened 2020-05-03
Spark 177 - Actually implement landmark near

theunkn0wn1:
This PR resolves SPARK-177 by actually implementing `!landmark near` instead of leaving it as dead code.

While not strictly necessary, I decided to use the `pyparsing` library for this command; and intend to use it in other commands as well to simplify things. 

Originally, I used pyparsing to manage subcommands, but it appears there is exactly ONE subcommand for the `!landmark` group - `near`.
Using pyparsing ive layed the foundation if we ever want to expand this in the future.

### DIFF
--- a/src/commands/starsystems.py
+++ b/src/commands/starsystems.py
@@ -14,11 +14,7 @@ async def cmd_search(ctx: Context):
     return await ctx.reply(f"{results!r}")
 
 
-async def cmd_landmark_near(ctx: Context):
-    if len(ctx.words) < 2:
-        return await ctx.reply("Usage: landmark near <system>")
-
-    system_name = ctx.words_eol[1]
+async def cmd_landmark_near(ctx: Context, system_name: str):
     logger.trace("searching for system {}", system_name)
     found = await ctx.bot.galaxy.find_system_by_name(system_name)
     if not found:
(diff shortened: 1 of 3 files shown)


theunkn0wn1:
pyparsing was already in our dependency tree FWIW, i just made it a direct dependency.

src/commands/starsystems.py
```diff
@@ -13,23 +14,37 @@ async def cmd_search(ctx: Context):
...
-    system_name = ctx.words_eol[1]
+async def cmd_landmark_near(ctx: Context, system_name: str):
     logger.trace("searching for system {}", system_name)
     found = await ctx.bot.galaxy.find_system_by_name(system_name)
     if not found:
-        return await ctx.reply(f"{system_name} was not found in The Fuel Rats System Database.")
+        return await ctx.reply(
+            f"{system_name} was not found in The Fuel Rats System Database.")
 
     logger.debug("found system {}, acquiring nearest landmark...", found)
```
Eearslya, line 24: [Question] Above log call is trace, this one is debug. Should we match them up?

Eearslya (approved):

src/commands/starsystems.py
theunkn0wn1: hmm,  id argue this is one of symantics. 

I made this one trace because its not really diagnostic information, its tracing an action.
The second log call isn't tracing an action, rather logging a result.

So my argument is `trace` is for tracing an action, `debug` is for the result thereof.
`debug` information may be more informative, as they apply to more situations.
`trace` is only really useful when its unclear the path the code takes (e.g. concurrent)

PR #181, opened 2020-05-09
Preliminary Prometheus integration

theunkn0wn1:
This PR implements preliminary telemetry for Mecha3,
Instrumentation of interest: 

## Histograms

 - time in `on_message`
 - time in `trigger`
 - time in `handle_fact`

## Gauges
 - size of `last_user_message`
## Counters
 - messages discarded
 - Trigger misses (triggers that end up not being anything actionable)
 - Errors (anything that invokes graceful errors via the `on_message` handler)

Demo of what these stats look like in Grafina.
image
Note this image is not necessarily reflective as the same size is quite small. we will need to sit a bot equipped with these features in a drill channel to get more actionable data.

### DIFF
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -27,6 +27,15 @@ from src.packages.commands import command
 from src.packages.context import Context
 from src.packages.permissions import require_permission, RAT
 
+import prometheus_client
+
+TIME_IN_PING = prometheus_client.Summary(
+    name="ping",
+    documentation="Time spent in the ping command",
+    namespace="command",
+    unit="seconds"
+)
+
 
 @require_permission(RAT)
 @command("ping")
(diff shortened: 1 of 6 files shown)


src/__main__.py
```diff
@@ -75,6 +85,7 @@ async def start():
 
 # entry point
 if __name__ == "__main__":
+    prometheus_client.start_http_server(8080, "localhost")
```
shatteredbeam: Are we going to send this out of app for analysis or will this remain?

shatteredbeam (commented):
I was expecting to see an actual telemetry package instead of locally placed hooks...?

src/__main__.py
theunkn0wn1: This should probably be at least a cli flag.
and yes, actual analysis occurs externally in prometheus / grafina.
there is very little mecha can do with these objects.

theunkn0wn1:
@shatteredbeam 
This was written as per Prometheus's documentation, which specifically states the telemetry objects to be placed in the same module it instruments.
```
Instrumentation should be an integral part of your code. Instantiate the metric classes in the same file you use them. This makes going from alert to console to code easy when you are chasing an error.
```
https://prometheus.io/docs/practices/instrumentation/#how-to-instrument

Eearslya (dismissed):

Eearslya (approved):

PR #184, opened 2020-06-17
[SPARK-125] Handle and test for two edge cases for confused clients

Padawan00123:


tests/unit/test_ratmama.py
```diff
@@ -178,6 +206,24 @@ async def test_manual_signal(bot_fx,
...
+        async_callable_fx,
+        monkeypatch):
+
+    context = await Context.from_message(
+        bot_fx,
+        "#unit_test",
+        "unit_test",
+        "Ratsignal some_service")
+
+    monkeypatch.setattr(context, "reply", async_callable_fx)
```
theunkn0wn1: [question] why was the reply attribute patched here? I don't see any usage of `async_callable_fx` used in an assertion as a result of this patch?

src/packages/ratmama/ratmama_parser.py
```diff
@@ -108,6 +109,16 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
     lang_code: str = result.group("language_code")
     nickname: Optional[str] = result.group("nick")
 
+    client = await User.from_pydle(ctx.bot, client_name)
+
+    if client is not None:
+        client_hostname = client.hostname
```
theunkn0wn1: [Error] Redundant variable assignment

You needlessly assign an instance attribute of a dataclass to a local name (read only, cheap retrieval) , it is better NOT to extract that one attribute from the object, as it may lead to broken expectations.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -108,6 +109,16 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
...
+
+        if client_hostname in (
+                "services.fuelrats.com",
+                "bot.fuelrats.com"):
+            return
```
theunkn0wn1:
```suggestion
    if client and client.hostname in ("services.fuelrats.com", "bot.fuelrats.com"):
        logger.warning("attempted to create rescue for a service!")
        # emit a warning to the dispatch, so they can handle this themselves.
        return await ctx.reply("signal attempted to create rescue for a service. Dispatch: please manually inject!")
```

theunkn0wn1 (changes requested):
Looks ok, but we need to warn the dispatch when we fail in this mode so they can handle it themselve.

theunkn0wn1 (dismissed):
LGTM :+1:

theunkn0wn1 (approved):
LGTM :+1:

PR #185, opened 2020-07-05
[SPARK-180|170|182] command parsing overhaul

theunkn0wn1:
This PR is a big one, it overhauls the way most commands are parsed.

Overarching idea is to parse the incoming commands using `pyparsing`, instead of raw string operations or regex.
By using `pyparsing` over regex, its far more readable.
By using `pyparsing` over direct string methods, its much more flexible *especially* for the more complicated commands.

This PR also has bug fixes for SPARK-170 and SPARK-182 as they fell into scope


src/commands/case_management.py
```diff
@@ -175,32 +215,37 @@ async def cmd_case_management_clear(ctx: Context):
...
         await ctx.reply("No case with that name or number.")
         return
 
     async with ctx.bot.board.modify_rescue(rescue) as case:
-        case.client = ctx.words[2]
+        case.client = tokens.new_cmdr.strip()
         await ctx.reply(f"Client for {case.board_index} is now CMDR {case.client}")
 
 
+CODE_RED_PATTERN = suppress_first_word + rescue_identifier.setResultsName("subject")
```
Eearslya: [Style] Should this not join its friends at the top of the file?

src/packages/ratmama/ratmama_parser.py
```diff
@@ -126,49 +145,47 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
...
+        client=client_name,
+        system=system_name,
+        irc_nickname=nickname,
+        code_red=not o2_status,
+        lang_id=lang_code,
+        platform=platform,
+    )
+    platform_signal = f"({rescue.platform.value.upper()}_SIGNAL)" if rescue.platform else ""
+    await ctx.reply(
+        f"DRILLSIGNAL - CMDR {rescue.client} - "
```
Eearslya: [Error] Hardcoded Drillsignal?

src/packages/ratmama/ratmama_parser.py
```diff
@@ -232,21 +243,34 @@ async def handle_ratsignal(ctx: Context) -> None:
             platform = Platforms["XB"]
 
         elif "o2" in part.casefold():
-            code_red = "o2 ok" not in part.casefold()
+            code_red = "o2 ok" not in message.casefold()
 
         else:
             system = part
 
+    if system and "signal" in system.lower():
```
Eearslya, line 250: [Question] What case is this meant to catch?

tests/unit/test_ratmama.py
```diff
@@ -152,8 +152,8 @@ async def test_announcer_reconnect_with_changes(bot_fx, async_callable_fx, monke
     await ratmama.handle_ratmama_announcement(context)
     await ratmama.handle_ratmama_announcement(context2)
 
-    assert async_callable_fx.was_called_with("System changed! Platform changed!"
-                                             " O2 Status changed, it is now CODE RED!")
+    assert async_callable_fx.was_called_with('Case #0 system, platform, O2 Status changed, '
+                                             'it is now CODE RED! O2 Status changed, it is now CODE RED!')
```
Eearslya: [Question] Is it intentional that O2 status is announced twice?

src/packages/ratmama/ratmama_parser.py
theunkn0wn1: I honestly don't recall what case this catches,  but it was a buggy edge case.

tests/unit/test_ratmama.py
theunkn0wn1: no, thats a bug :P

src/packages/ratmama/ratmama_parser.py
theunkn0wn1: :tired_face:

src/commands/case_management.py
theunkn0wn1: probably. will fix

tests/unit/test_ratmama.py
theunkn0wn1: Resolved by fixing the erroneous underlying logic.

src/packages/ratmama/ratmama_parser.py
theunkn0wn1: Resolved hardcoded signal.

src/commands/case_management.py
theunkn0wn1: Moved.

theunkn0wn1:
I removed two of the hypothesis tests that frequently failed health checks.
It was a good idea to have them, but more effort needs to be made in stabilizing them.

Eearslya (approved):

PR #187, opened 2020-09-10
Fix `inject`s parser, add platform detection support

theunkn0wn1:
In review it appears that `inject` never got platform detection support.
In the process of adding this, it became evident that `inject`s pattern was incorrect.

This PR corrects that and corrects the tests that missed this regression.

Caused by #185

### DIFF
--- a/src/commands/case_management.py
+++ b/src/commands/case_management.py
@@ -105,7 +105,10 @@ INJECT_PATTERN = (
     + rescue_identifier.setResultsName("subject")
     # The following group captures in any order (&).
     + (
-        pyparsing.Optional(pyparsing.CaselessKeyword("cr")).setResultsName("code_red")
+        pyparsing.Optional(
+            pyparsing.CaselessKeyword("cr")
+            ^ pyparsing.CaselessKeyword("code red")
+        ).setResultsName("code_red")
         & pyparsing.Optional(timer("timer"))
         & pyparsing.Optional(platform).setResultsName("platform")
     )
--- a/src/commands/case_management.py
+++ b/src/commands/case_management.py
@@ -26,7 +26,7 @@ from ..packages.commands import command
 from ..packages.context.context import Context
 from ..packages.epic import Epic
 from ..packages.parsing_rules import rescue_identifier, irc_name, suppress_first_word, timer, \
-    rest_of_line
+    rest_of_line, platform
 from ..packages.permissions.permissions import (
     require_permission,
     RAT,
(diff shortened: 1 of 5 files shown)


src/commands/case_management.py
```diff
@@ -387,10 +395,8 @@ async def cmd_case_management_inject(ctx: Context):
             for keyword in ctx.words_eol[2].split():
```
Eearslya: [Question] Why are we still using this loop at all when we have `platform` in the pyparsing object now?

tests/unit/commands/test_board.py
```diff
@@ -215,7 +215,7 @@ def _test_quote_header(cr_state, message, rescue):
 @pytest.mark.parametrize("platform", (Platforms.PC, Platforms.XB, Platforms.PS))
 async def test_quote_inject_interop(bot_fx, cr_state: bool, platform: Platforms):
     inject_ctx = await Context.from_message(bot_fx, "#ratchat", "some_ov",
-                                            f"!inject subject PC sol {'cr' if cr_state else ''}")
+                                            f"!inject subject PC {'cr' if cr_state else ''} sol")
```
Eearslya, line 218: [Suggestion] Add a test for injects with timers rather than "CR".

src/commands/case_management.py
```diff
@@ -94,14 +100,18 @@
...
-) + pyparsing.Optional(
-    timer("timer")
-) + rest_of_line.setResultsName(
-    "remainder"
+INJECT_PATTERN = (
+    suppress_first_word
+    + rescue_identifier.setResultsName("subject")
+    # The following group captures in any order (&).
+    + (
+        pyparsing.Optional(pyparsing.CaselessKeyword("cr")).setResultsName("code_red")
```
Eearslya: [Question] Might it be a good idea to expand this to include "code red" as an option?

Eearslya (changes requested):

src/commands/case_management.py
theunkn0wn1: as that takes two tokens that is somewhat involved...
My attempts to add `"code red"` as a recognized term only produced more problems...

theunkn0wn1: Nvm, figured it out. Will include!

theunkn0wn1: image

src/commands/case_management.py
theunkn0wn1: hold over from when i didn't have platform detection in the pyparsing object. will correct.

tests/unit/commands/test_board.py
theunkn0wn1: Covered by the hypothesis suite.

Eearslya (approved):

PR #199, opened 2020-09-23
[SPARK-277|SPARK-46] Return case handling improvements

theunkn0wn1:
Fixes to return case handling, including SPARK-277 in which crashes occured (regression test added)
Also made the output actually /work/ (SPARK-46)
﻿- [SPARK-277|SPARK-46] Fixed return case handling
- [SPARK-277|SPARK-46] Black -l103

### DIFF
--- a/src/packages/ratmama/ratmama_parser.py
+++ b/src/packages/ratmama/ratmama_parser.py
@@ -20,7 +20,7 @@ from ..context import Context
 from ..rescue import Rescue
 from ..rules import rule
 from ..user import User
-from ..utils import Platforms
+from ..utils import Platforms, color, Colors, bold
 
 import attr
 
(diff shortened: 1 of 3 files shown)


src/packages/ratmama/ratmama_parser.py
```diff
@@ -146,19 +153,38 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
...
+                ", O2 Status!"
+                if o2_status
+                else f", O2 Status changed, rescue is now {color('CODE RED', Colors.RED)}!"
+            )
+
+        if changed:
+            # SPARK-46: Warn when a client reconnects with different settings, but differ to dispatch
+            # to overwrite existing data instead of doing it ourselves.
+            await ctx.reply(f"{message}{', '.join(changed)}{cr_message}")
+            return
```
theunkn0wn1: [error] this return is supposed to be unconditional.

PR #200, opened 2020-09-26
An attempt at logging segfaults

theunkn0wn1:
Due to recent incidents in drillrats, its evident something is peeing in the bed.
Dunno how much use these handlers will be, but hopefully it will be of operational use.
This PR adds a handler to catch SIGSEGV and spit out a stacktrace prior to bailing out.

Since these signals are likely fatal, its probably NOT a good idea to use the logging infrastructure for it or attempt to make any further use of the async runtime.

### DIFF
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -32,8 +32,7 @@ from src.packages.permissions import require_permission, RAT
 
 import prometheus_client
 
-import signal
-
+import faulthandler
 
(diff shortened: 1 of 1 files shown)


theunkn0wn1:
This can actually be achieved without a code change, and on further consideration I don't wish to hardcode this anyways.

for future reference, running the interpreter with the `-X faulthandler` flag achieves the same result, without requiring a code change.

PR #202, opened 2020-10-13
[SPARK-173] Fixes towards dockerfiles.

theunkn0wn1:


theunkn0wn1:
At this point the dockerfiles arn't used at all in the deployment context.
Further as of present they arn't used for development either.
As such, the decision is to purge these files entirely from the codebase, such that individual developers can create these files if they need them as to facilitate their local development needs.

theunkn0wn1:
Superceeded by #204

PR #204, opened 2020-10-14
Doc/simplify infrastructure deployment

diraven:


docker-compose.template.yml
```diff
@@ -0,0 +1,24 @@
+version: '3.2'
+services:
+  irc:
+    init: true
+    image: theunkn0wn1/unrealircd
+    ports:
+      - "6667:6667"
+      - "6697:6697"
```
theunkn0wn1, line 8: [suggestion] please also expose the non-tls port.

config/config.template.toml
```diff
@@ -1,9 +1,9 @@
 [irc]
-nickname = "Mechasqueak3-tests[BOT]"
+nickname = "Mechasqueak3[BOT|SPARK]"
```
theunkn0wn1: Please revert this specific change. I would perfer the default for this configuration object NOT be the production name.

config/config.template.toml
```diff
@@ -14,31 +14,35 @@ log_file = "logs/unit_tests.log"
 
 # GELF logging output
 [logging.gelf]
-enabled = false
+enabled = true
```
theunkn0wn1: This should be default False, since unless you actually spin up a graylog cluster and configure it, this will produce needless network traffic and overhead.

config/config.template.toml
```diff
@@ -14,31 +14,35 @@ log_file = "logs/unit_tests.log"
...
-fact_log = "fact_transaction"
+fact_log = "fact_log"
 
 [commands]
 prefix = "!"
 
 [api]
-online_mode = false
-url = "http://localhost/"
+online_mode = true
```
theunkn0wn1: i think this should be off by default.

theunkn0wn1 (changes requested):
bunch of minor things, otherwise LGTM

docker-compose.template.yml
theunkn0wn1: (i can read!)

theunkn0wn1 (approved):
LGTM. thanks for this!

PR #205, opened 2020-10-14
[SPARK-165] Feature/ratmama announce process distance

diraven:


tests/conftest.py
```diff
@@ -136,8 +136,10 @@ def rat_board_fx() -> RatBoard:
 
 
 @pytest.fixture()
-def bot_fx(configuration_fx):
-    return MockBot(nickname="mock_mecha3[BOT]", mecha_config=configuration_fx)
+def bot_fx(configuration_fx, galaxy_fx):
+    mock_bot = MockBot(nickname="mock_mecha3[BOT]", mecha_config=configuration_fx)
+    mock_bot._galaxy = galaxy_fx
```
theunkn0wn1: `mock_bot.galaxy` has a setter. use it.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -203,9 +203,21 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
         platform=platform,
     )
     platform_signal = f"({rescue.platform.value.upper()}_SIGNAL)" if rescue.platform else ""
+
+    distance_str = "not found in the galaxy DB"
+    system = await asyncio.wait_for(ctx.bot.galaxy.find_system_by_name(system_name), timeout=2)
+    if system:
+        landmark_info = await asyncio.wait_for(ctx.bot.galaxy.find_nearest_landmark(system), timeout=2)
+        if landmark_info:
```
theunkn0wn1: [suggestion] Not sure in what context a system would, in fact, exist but no landmark data for it.
This case appears to be unhandled though, and the `find_nearest_landmark` *does* return an optional here soo....
we may want to explicitly handle that case

src/packages/ratmama/ratmama_parser.py
```diff
@@ -203,9 +203,21 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
         platform=platform,
     )
     platform_signal = f"({rescue.platform.value.upper()}_SIGNAL)" if rescue.platform else ""
+
+    distance_str = "not found in the galaxy DB"
+    system = await asyncio.wait_for(ctx.bot.galaxy.find_system_by_name(system_name), timeout=2)
```
theunkn0wn1: these can throw timeout exceptions, which concern me as they are unhandled.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -203,9 +203,26 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
...
+            landmark_info = await asyncio.wait_for(
+                ctx.bot.galaxy.find_nearest_landmark(system), timeout=2,
+            )
+            if landmark_info:
+                landmark, distance = landmark_info
+                if system.name != landmark.name:
+                    distance_str = f"{distance}ly from {landmark.name}"
+                else:
+                    distance_str = "landmark"
+    except asyncio.TimeoutError:
```
theunkn0wn1: i think aiohttp has a similar TimeoutError class, we should probably also catch that.

src/packages/ratmama/ratmama_parser.py
```diff
@@ -203,9 +203,26 @@ async def handle_ratmama_announcement(ctx: Context) -> None:
...
+        distance_str = "galaxy DB request timeout"
```
theunkn0wn1:
```suggestion
        distance_str = "<timeout requesting system data>"
```

tests/conftest.py
diraven: There is a warning that made me thinking using setter is discouraged. =]

theunkn0wn1 (commented):
Just some minor improvements, overall looks good :)

tests/conftest.py
theunkn0wn1: not entirely sure why there is a warning message there, but i think it had the rational that it shouldn't just be set without good reason (and thus should be logged).

That said, its bad practice to bypass a setter and access private attributes of an object from outside said object.
Yes this is a test, but i don't think this specific case warrants bypassing the setter.
*objectively this should probably be included in the MechaClient constructor but thats out of scope*

diraven: Noted.

Eearslya (dismissed):
Your Docker PR appears to have leaked into this PR...

diraven:
@Eearslya , it's done on purpose. I needed that infrastructure to run bot locally. Can rebase the PR if necessary. Might be easier of that one is accepted first. No need to rebase then and this one will be free of docker-related changes. They are just chained on top of each other.

theunkn0wn1 (approved):
LGTM

PR #209, opened 2020-11-01
Fix/landmark cmd perms usage

diraven:


theunkn0wn1 (approved):
How did THAT happen?!

thanks for that....

we need extra validation on `require_permission` to prevent invalid usages like these.
thats probably OOS for this PR so im marking this as approved.

theunkn0wn1 (changes requested):
Actually no, on further review the permission isn't even applied as its below the command decorator.

theunkn0wn1 (approved):
i think this is fine. even if its not, going to differ that to SPARK-154 and i am going to see to that ticket.

PR #220, opened 2020-12-31
[SPARK-216] Added !quiet command + tests

Master-Guy:


tests/unit/test_rat_command.py
```diff
@@ -149,3 +159,33 @@ async def the_command(context: Context):
...
+        # Pre-case creation
+        retn = await Commands.trigger(ctx)
+        print(f"sent pre: {bot_fx.sent_messages} - lastcase: {rat_board_fx._datetime_last_case}")
+        assert "Got no information yet" in bot_fx.sent_messages[0]['message'] # Works
+
+        # During case active
+        rescue = await rat_board_fx.create_rescue(client=random_string_fx)
+        retn = await Commands.trigger(ctx)
+        print(f"sent active: {bot_fx.sent_messages} - lastcase: {rat_board_fx._datetime_last_case}")
+        assert "There is corrently an active rescue" in bot_fx.sent_messages[1]['message'] # Does not work, still returns "Got no information yet"
```
theunkn0wn1: unresolved FIXME?

tests/unit/test_rat_command.py
```diff
@@ -149,3 +159,33 @@ async def the_command(context: Context):
...
+
+        # Post-case active
+        await rat_board_fx.remove_rescue(rescue._board_index)
+        retn = await Commands.trigger(ctx)
+        print(f"sent post: {bot_fx.sent_messages} - lastcase: {rat_board_fx._datetime_last_case}")
+        assert "The last case was created 0 minutes ago." in bot_fx.sent_messages[2]['message'] # Does not work, still returns "Got no information yet"
```
theunkn0wn1: unresolved FIXME?

src/commands/case_management.py
```diff
@@ -448,6 +448,41 @@ async def cmd_case_management_system(ctx: Context):
...
 
 
+@command("quiet", require_channel=True, require_permission=RAT)
+async def cmd_case_management_quiet(ctx: Context):
+    # Check if there is an active rescue
+    """
+    if ctx.bot is not None:
+        if ctx.bot.board is not None:
+    """
+    for rescue in ctx.bot.board.values():  # type: Rescue
```
theunkn0wn1: [minor] unnecessary type hint

src/commands/case_management.py
```diff
@@ -448,6 +448,41 @@ async def cmd_case_management_system(ctx: Context):
...
+
+    if ctx.bot.board._datetime_last_case == datetime.datetime.utcfromtimestamp(0):
+        await ctx.reply("Got no information yet")
+        return
+
+    timediff = divmod(
+        divmod(
+            int(
+                (
+                    datetime.datetime.now(tz=timezone.utc) - ctx.bot.board._datetime_last_case
```
theunkn0wn1: [error] access to protected data member of a different class

src/commands/case_management.py
```diff
@@ -448,6 +448,41 @@ async def cmd_case_management_system(ctx: Context):
...
+            60,
+        )[0],
+        60,
+    )
+    if timediff[0] > 0:
+        await ctx.reply(
+            f"The last case was created {timediff[0]} hours and {timediff[1]} minutes ago."
+        )
+    else:
+        await ctx.reply(f"The last case was created {timediff[1]} minutes ago.")
```
theunkn0wn1: I am quite confident there is a easier way to achieve this.

src/config/datamodel/board.py
```diff
@@ -11,8 +11,14 @@
...
 
 import attr
+import datetime
+from datetime import timezone
 
 
 @attr.dataclass
 class BoardConfigRoot:
     cycle_at: int = attr.ib(validator=attr.validators.instance_of(int))
+    datetime_last_case: datetime.datetime = attr.ib(
```
theunkn0wn1: [question] Please justify why this would be beneficial to be configurable?

tests/unit/test_rat_board.py
```diff
@@ -132,3 +135,12 @@ async def test_modify_rescue_explosion(rat_board_fx, random_string_fx):
             raise RuntimeError  # intentionally raise an exception
 
     assert random_string_fx in rat_board_fx, "the board dropped the rescue!"
+
+@pytest.mark.asyncio
+async def test_modify_rescue_datetimelastcase(rat_board_fx, monkeypatch):
```
theunkn0wn1: [style] snake_case between words in a function definition please.
```suggestion
async def test_modify_rescue_datetime_last_case(rat_board_fx, monkeypatch):
```

tests/unit/test_rat_command.py
Master-Guy: Removed the comments

tests/unit/test_rat_board.py
```diff
@@ -132,3 +135,12 @@ async def test_modify_rescue_explosion(rat_board_fx, random_string_fx):
...
+    """
+    Verifies that _datetime_last_case is being modified when creating a new case
+    """
+    pre_datetime = datetime.datetime.now(tz=timezone.utc)
```
theunkn0wn1: [error] This shouldn't pull the current datetime, but rather the current value of `rat_board_fx._datetime_last_case`

To ensure predictability, one might have to insert a small arbitrary delay to make the following assertion work.

tests/unit/test_rat_command.py
Master-Guy: Removed the comments

tests/unit/test_rat_command.py
```diff
@@ -149,3 +149,31 @@ async def the_command(context: Context):
...
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize("alias", ['quiet'])
+    async def test_call_command_quiet(self, alias, bot_fx, rat_board_fx, configuration_fx, random_string_fx, monkeypatch):
+        """
+        Verifiy that the !quiet command returns useful information
+        """
+        trigger_alias = f"{configuration_fx.commands.prefix}{alias}"
+        print(f"Triggering alias: {trigger_alias}")
+        bot_fx.board = rat_board_fx
+        ctx = await Context.from_message(bot_fx, "#unittest", "some_rat", trigger_alias)
```
theunkn0wn1, line 163: [error] please use provided logging facilities instead of `print`ing to STDOUT.

```py
from loguru import logger

logger.debug(...)
```

tests/unit/test_rat_board.py
```diff
@@ -8,6 +8,9 @@
 
 from src.packages.board.board import cycle_at
 
+import datetime
+from datetime import timezone
```
theunkn0wn1: [suggestion] this can be achieved in a single import, and would remove a level of ambiguity.

something to the effect of 
```py
from datetime import datetime, timezone
```

src/commands/case_management.py
```diff
@@ -448,6 +448,41 @@ async def cmd_case_management_system(ctx: Context):
...
+    """
+    if ctx.bot is not None:
+        if ctx.bot.board is not None:
+    """
+    for rescue in ctx.bot.board.values():  # type: Rescue
+        if rescue.active:
+            await ctx.reply("There is corrently an active rescue")
+            return
+
+    if ctx.bot.board._datetime_last_case == datetime.datetime.utcfromtimestamp(0):
```
theunkn0wn1: [error] access to private field of another class.

theunkn0wn1 (changes requested):
Bunch of relatively minor things that need to be resolved.

src/commands/case_management.py
Master-Guy: Removed

src/config/datamodel/board.py
Master-Guy: It isn't, I was drawn here because I was told that the datamodel had to be expanded, and this seemed logical to do so.

tests/unit/test_rat_board.py
Master-Guy: Added the sleep, but it shouldn't (only) pull the current version of rat_board_fx._datetime_last_case, because that value will be 1900-01-01 by default, and thus downgrade the test. I've added an additional step to the test to also check the initial value.

tests/unit/test_rat_command.py
Master-Guy: Replaced all (my) prints with loguru

tests/unit/test_rat_board.py
Master-Guy: Updated, also had to update case_management.py in more commands to make sure it all kept working (datetime.datetime -> datetime)

src/commands/case_management.py
Master-Guy: Didn't consider it to be private, due to only having a single underscore. Default conventions ask for (at least) two underscores prefixing a private variable. Added an @property to retrieve the value, as there are no external setters.

src/commands/case_management.py
Master-Guy: Didn't consider it to be private, due to only having a single underscore. Default conventions ask for (at least) two underscores prefixing a private variable. Added an @Property to retrieve the value, as there are no external setters.

src/commands/case_management.py
Master-Guy: Simplified the code, but it basically still does the same actions. Haven't found any workable datetime parsers that have the ability to conditionally format the output.

src/commands/case_management.py
```diff
@@ -448,6 +448,30 @@ async def cmd_case_management_system(ctx: Context):
...
+            await ctx.reply("There is corrently an active rescue")
+            return
+
+    if not ctx.bot.board.last_case_datetime:
+        await ctx.reply("Got no information yet")
+        return
+
+    timediff = relativedelta(datetime.now(tz=timezone.utc), ctx.bot.board.last_case_datetime)
+
+    hourpart = timediff.hours > 0 and f"{timediff.hours} hours and " or ""
```
theunkn0wn1:
```suggestion
    hourpart = ""
    if timediff.hours:
        hourpart = f"hour{'s' if timediff.hours>1}"
```

src/packages/board/board.py
```diff
@@ -356,3 +365,8 @@ async def remove_rescue(self, target: BoardKey):
             # TODO: add to internal deck in offline mode so we can push to the API when we eventually
             del self[target]
         logger.trace("Released modification lock.")
+
+    @property
+    def last_case_datetime(self):
```
theunkn0wn1:
```suggestion
    def last_case_datetime(self) -> Optional[datetime]:
```

src/commands/case_management.py
```diff
@@ -448,6 +448,30 @@ async def cmd_case_management_system(ctx: Context):
...
+    for rescue in ctx.bot.board.values():
+        if rescue.active:
+            await ctx.reply("There is corrently an active rescue")
+            return
+
+    if not ctx.bot.board.last_case_datetime:
+        await ctx.reply("Got no information yet")
+        return
+
+    timediff = relativedelta(datetime.now(tz=timezone.utc), ctx.bot.board.last_case_datetime)
```
theunkn0wn1: [style] PEP8 naming convention.
Also may i suggest using `delta` here?

src/commands/case_management.py
theunkn0wn1: [PEP8] Naming conventions. `snake_case` is prefered.
E.G `hour_part`

theunkn0wn1 (commented):
Minor nits at this point to improve readability.

theunkn0wn1 (dismissed):
LGTM :+1:

theunkn0wn1 (approved):
LGTM:+1:﻿

PR #221, opened 2020-12-31
[SPARK-157] Added optional inject message + tests

Master-Guy:
- Allowing an optional injection message with the !active command and aliases
- Added tests to check the bot output
- Added tests to see if !active toggles the active state of a case correctly
- Tests are executed with and without optional injection message


Master-Guy:
Replaced with PR223

PR #222, opened 2020-12-31
Spark 199 

Master-Guy:
!grab on a client that has no open case now tells the person that executed the command to use !inject instead to create a new case.


Master-Guy:
Replaced by PR 224

PR #223, opened 2020-12-31
SPARK-157 optional !active inject messages + tests

Master-Guy:
Let's try this again, and let's cancel PR221 for this one

### DIFF
--- a/src/commands/case_management.py
+++ b/src/commands/case_management.py
@@ -53,13 +53,14 @@ ASSIGN_PATTERN = (
     + rescue_identifier.setResultsName("subject")
     + pyparsing.OneOrMore(irc_name).setResultsName("rats")
 )
-ACTIVE_PATTERN = (
 
+ACTIVE_PATTERN = (
     suppress_first_word
     + rescue_identifier.setResultsName("subject")
     # This comes positionally LAST or it catches the wrong things.
     + rest_of_line.setResultsName("remainder")
 )
+
 CLEAR_PATTERN = (
     suppress_first_word
     + rescue_identifier.setResultsName("subject")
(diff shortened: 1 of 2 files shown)


src/commands/case_management.py
```diff
@@ -143,8 +148,16 @@ async def cmd_case_management_active(ctx: Context):
         return
 
     # We either have a valid case or we've left the method at this point.
-    async with ctx.bot.board.modify_rescue(rescue) as case:
+    async with ctx.bot.board.modify_rescue(rescue, impersonation=ctx.user.account) as case:
+        logger.debug(f"Switching case to active = {not case.active}")
         case.active = not case.active
+        logger.debug(f"Length: {len(ctx.words_eol)}")
+        if len(ctx.words_eol) > 2:
```
theunkn0wn1: [suggestion] in `ACTIVE_PATTERN` you will have `tokens.remainder`, i suggest checking that instead

theunkn0wn1 (commented):
One minor nit, i believe it would be more logically correct to check for a non-empty remainder (remember to lstrip whitespace)

src/commands/case_management.py
```diff
@@ -143,8 +149,16 @@ async def cmd_case_management_active(ctx: Context):
...
-    async with ctx.bot.board.modify_rescue(rescue) as case:
+    async with ctx.bot.board.modify_rescue(rescue, impersonation=ctx.user.account) as case:
+        logger.debug(f"Switching case to active = {not case.active}")
         case.active = not case.active
+        logger.debug(f"Inject message: {tokens.remainder}")
+        if tokens.remainder:
+            # Inject message before toggling active/inactive if it is passed
+            case.add_quote(ctx.words_eol[0], ctx.user.nickname)
+            await ctx.reply(
+                f"{case.client}'s case updated with: {ctx.words_eol[2]!r} (Case {case.board_index})"
```
theunkn0wn1: [error] You inject `ctx.words_eol[0]` but indicate you update with `ctx.words_eol[2]`.
```suggestion
                f"{case.client}'s case updated with: {ctx.words_eol[0]!r} (Case {case.board_index})"
```

theunkn0wn1 (changes requested):
Missed a logical error. This causes regression SPARK-242, see fix #215

theunkn0wn1 (approved):

PR #224, opened 2020-12-31
SPARK-199 !grab no longer creates new cases

Master-Guy:
Instead, it refers the dispatch to !inject a case first.


theunkn0wn1 (approved):
LGTM :+1:

PR #229, opened 2021-01-02
SPARK-143 Removed assert from prod code

Master-Guy:
SPARK-143 Removed assert from prod code, and replaced it by an if-statement with a raise

### DIFF
--- a/src/packages/database/database_manager.py
+++ b/src/packages/database/database_manager.py
@@ -111,21 +111,26 @@ class DatabaseManager:
             # Utilize function arguments if they are provided,
             # otherwise retrieve from config file and use those values.
             self._dbhost = dbhost if dbhost is not None else self._config.database.host
-            assert self._dbhost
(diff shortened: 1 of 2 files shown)


src/packages/database/database_manager.py
```diff
@@ -111,21 +111,26 @@ def __init__(self, dbhost=None, dbport=None, dbname=None, dbuser=None, dbpasswor
             # Utilize function arguments if they are provided,
             # otherwise retrieve from config file and use those values.
             self._dbhost = dbhost if dbhost is not None else self._config.database.host
-            assert self._dbhost
+            if not self._dbhost:
+                raise UnboundLocalError("_dbhost is not set")
```
theunkn0wn1: I don't believe these should be unbound local errors, but assertion errors.

The fields exist (Thus are bound), they just fail an assertion

Master-Guy:
Replacing with new PR

PR #230, opened 2021-01-03
SPARK-143 Removed assert from prod code

Master-Guy:
SPARK-143 Removed assert from prod code, and replaced with raised errors


theunkn0wn1 (approved):
LGTM!

PR #232, opened 2021-01-03
SPARK-188 Allow !md in privmsg and tests added

Master-Guy:
With the rescue revisions, marking deletion does not need to be send in a channel anymore. Deletion also creates a rescue revision.


theunkn0wn1 (approved):
LGTM :+1:

